PR #1, opened 2019-07-23
Improve maybe util types

jwm0:
Adds overload types for better static return types


dominik-zeglen:
Already merged to `mirumee/saleor`

PR #147, opened 2019-09-02
Password reset flow

### LINKED ISSUE #139: Add views to set/reset password
We need to add views to set/reset password. Currently, in emails we have hardcoded links to Storefront 1.0 views, but the API to support other redirect links is already built. 

image

dominik-zeglen:
I want to merge this change because it adds password resetting.
Resolves #139 

### Screenshots
Screenshot 2019-09-03 at 15 45 35
Screenshot 2019-09-03 at 15 45 44
Screenshot 2019-09-03 at 15 46 03

### Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,3 +16,4 @@ All notable, unreleased changes to this project will be documented in this file.
 - Fix table cell padding - #143 by @dominik-zeglen
 - Add fallback locale - #153 by @dominik-zeglen
 - Replace checkbox with switch component in "product type has variants" - #152 by @dominik-zeglen
+- Add password reset flow - #147 by @dominik-zeglen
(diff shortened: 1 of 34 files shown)


maarcingebala:
Few issues:

- There is no error handling in the "reset-password" view. Some errors that I've encountered so far:
    - in the `email` field - if the given email doesn't exist
    - in the `redirectUrl` field when the server isn't configured to accept given URL we'll get an error like "localhost:9000 this is not valid storefront address.". While this error doesn't make sense for the user, we should catch this error and display some message anyway e.g. "Failed to send the email, please contact the administrator or try again later."

- In the "new-password" view the "New password" input should get the focus first, currently the "Confirm password" inputs gets it.
- Joining URL is broken - `http://localhost:9000//new-password/` - do we use any lib for joining URL or we do it manually?

PR #157, opened 2019-09-05
Reset apollo client store cache on logout

jwm0:


maarcingebala:
@dominik-zeglen What do we do with this one?

maarcingebala:
@dominik-zeglen Bump

PR #158, opened 2019-09-05
Implement the Credential Management API

### LINKED ISSUE #156: Implement the Credential Management API
### What I'm trying to achieve
Log returning users in automatically.

### Describe a proposed solution
When an anonymous user visits the dashboard, check if they have any stored credentials and offer them to sign in.

When a user logs in using a password, attempt to store/update the credentials.

https://developers.google.com/web/fundamentals/security/credential-management/

patrys:
Fixes #156

Very rough implementation, Credential Management API calls should be extracted and live outside of the component, there should be a single "is the API supported" helper etc.

### Pull Request Checklist

### DIFF
--- a/src/auth/AuthProvider.tsx
+++ b/src/auth/AuthProvider.tsx
@@ -105,9 +105,7 @@ class AuthProvider extends React.Component<
     if (!!token && !user) {
       this.verifyToken(token);
     } else {
-      if (isCredentialsManagementAPISupported) {
-        loginWithCredentialsManagementAPI(this.login);
-      }
+      loginWithCredentialsManagementAPI(this.login);
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/auth/AuthProvider.tsx
```diff
@@ -99,15 +104,24 @@ class AuthProvider extends React.Component<
     const token = getAuthToken();
     if (!!token && !user) {
       this.verifyToken(token);
+    } else {
+      if (isCredentialsManagementAPISupported) {
```
patrys: Wouldn't it make more sense to move this `if` inside the `loginWithCredentialsManagementAPI` function?

dominik-zeglen: I think it makes more sense to check it here since we're about to call an API which we don't know if it's supported by user's browser. But it's purely semantic difference and I'll change it if you want 🤷‍♂

PR #213, opened 2019-10-16
Add language switch

dominik-zeglen:
I want to merge this change because it adds ability to switch dashboard's language.

### Screenshots
Screenshot 2019-10-16 at 17 32 30

### Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,3 +40,4 @@ All notable, unreleased changes to this project will be documented in this file.
 - Update sale details design - #207 by @dominik-zeglen
 - Improve autocomplete UX - #212 by @dominik-zeglen
 - Fix empty attribute values - #218 by @dominik-zeglen
+- Add language switch - #213 by @dominik-zeglen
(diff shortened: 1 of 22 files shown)


maarcingebala:
image

- Shouldn't this dropdown has its own scroll and be a bit shorter? Now when opening it, entire page gets the height of the dropdown. I think we should show around 10 choices and implement a scrollbar inside.
- Dropdown choices should be capitalized.
- How do we determine the available choices in the dropdown? Wouldn't it make sense to only show those languages where we have some meaningful coverage?

dominik-zeglen:
1. #212 is merged, so it's cropped and scrollable now.
2. Apparently no, at least that's what @patrys and `django.utils.translation.get_language_info` say 🤷‍♂️
3. Contrary to the input label, it changes not only language but also locale - meaning date, money and weight formats. If it doesn't suit you, you can change it any time, so I don't think it's a problem here.

maarcingebala:
JS doesn't support such sophisticated features as capitalizing the first letter? It looks bad, I'm sure we can do something @dominik-zeglen.

About languages - I'm not sure how helpful it is to switch to some locale and have half of the dashboard in English anyway. I'd rather limit the list of choices and provide only those where we really have some meaningful coverage. Offering a language choice that actually doesn't do much looks more like a bug on our side.

Issue #171, opened 2019-09-12
UI Improvements #5


pwgryglak:
- [x] modal have overflow. They shouldn't
![Screen Shot 2019-09-12 at 13 01 27](https://user-images.githubusercontent.com/40759237/64779877-5a978680-d55f-11e9-93bc-9c1d61542466.png)

- [x] those texts shouldn't be green. 
<img width="59" alt="Screen Shot 2019-09-12 at 12 53 19" src="https://user-images.githubusercontent.com/40759237/64779896-66834880-d55f-11e9-9370-6234d8be9657.png">

- [x] Cancel buttons shouldn't be green. Should be gray.
![Screen Shot 2019-09-12 at 12 56 45](https://user-images.githubusercontent.com/40759237/64779908-73a03780-d55f-11e9-92c9-4bc80cb5fcf9.png)

- [x] This input after adding a note shouldn't keep the text
![Screen Shot 2019-09-12 at 12 57 57](https://user-images.githubusercontent.com/40759237/64779924-7dc23600-d55f-11e9-93ae-8ae2320a0cd0.png)

- [x] need to lose focus on dropdowns to change value.
![Screen Shot 2019-09-12 at 12 54 26](https://user-images.githubusercontent.com/40759237/64780001-a813f380-d55f-11e9-9e2f-351a2420f7af.png)

- [ ] autocomplete doesn't match the ui. They are not fully filled with color.

- [x] Improve padding of this element. Text should be further.
<img width="620" alt="Screen Shot 2019-09-12 at 13 19 39" src="https://user-images.githubusercontent.com/40759237/64780190-05a84000-d560-11e9-9ec0-9e6d0f82dc04.png">

- [x] Save bar is static. Should be always on the bottom of the browser.

- [x] Dropdown lost arrows.

- [x] wrong styling of the note element. Should be this
<img width="276" alt="Screen Shot 2019-09-12 at 13 25 52" src="https://user-images.githubusercontent.com/40759237/64780519-dd6d1100-d560-11e9-85ed-53506b4c2fb2.png">

- [x] buttons should be green on hover
- [x] faceted search should be capitalized
<img width="283" alt="Screen Shot 2019-09-12 at 13 30 39" src="https://user-images.githubusercontent.com/40759237/64780801-9cc1c780-d561-11e9-9a21-f212217f462f.png">

- [x] paddings should be smaller
[...]

PR #226, opened 2019-10-22
Fix ui improvements

benekex2:
I want to merge this change because its fix UI issues

Ref #171 

### Pull Request Checklist

1. [ ] Translated strings are extracted to `.pot` file.

### DIFF
--- a/src/theme.ts
+++ b/src/theme.ts
@@ -60,7 +60,7 @@ export default (colors: IThemeColors): Theme =>
         },
         flat: {
           "& span": {
-            color: colors.font.gray
+            color: colors.font.default
           }
         },
         flatPrimary: {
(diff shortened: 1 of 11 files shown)


src/theme.ts
```diff
@@ -60,7 +60,7 @@ export default (colors: IThemeColors): Theme =>
         },
         flat: {
           "& span": {
-            color: colors.font.gray
+            color: colors.font.default
```
dominik-zeglen: Do we need to set this at all?

src/theme.ts
```diff
@@ -100,6 +100,9 @@ export default (colors: IThemeColors): Theme =>
       },
       MuiDialogContent: {
         root: {
+          "& label": {
```
dominik-zeglen: I think `overflow-x: hidden` would be a lot better

src/theme.ts
benekex2: Yes, we have to.

PR #267, opened 2019-11-22
Add navigator 

dominik-zeglen:
I want to merge this because it adds the navigator. It allows to quickly move around dashboard as a power user. Also features catalog and customer search.

### Screenshots
Screenshot 2019-11-25 at 17 56 00
Screenshot 2019-11-25 at 17 56 17
Screenshot 2019-11-25 at 17 56 33
Screenshot 2019-11-25 at 17 56 52
Screenshot 2019-11-25 at 17 57 01
Screenshot 2019-11-25 at 17 57 26

### Pull Request Checklist


src/components/Navigator/Navigator.tsx
```diff
@@ -0,0 +1,159 @@
...
+    hotkeys(navigatorHotkey, event => {
+      event.preventDefault();
+      setVisible(!visible);
+    });
+
+    if (cmp(APP_VERSION, "2.1.0") !== 1 && !notifiedAboutNavigator) {
+      notify({
+        text: intl.formatMessage({
+          defaultMessage:
+            "Our new feature to help you with your daily task. Run Navigator using Ctrl+K shortcut. (Cmd+K for Mac users)",
```
maarcingebala: Little grammar fix:

```suggestion
            "Our new feature to help you with your daily tasks. Run Navigator using Ctrl+K shortcut (Cmd+K for Mac users).",
```

I would check if we could actually use "⌘" symbol instead of "Cmd".  Also not a big deal but keyboard shortcuts looks nice formatted as code: `⌘+K`, maybe we could do something like that in this message?

mirekm:
Super cool feature. One issue I found testing is that searching on Clients doesn't seem to be working correctly: 
image

src/components/Navigator/modes/messages.ts
```diff
@@ -0,0 +1,57 @@
+import { defineMessages } from "react-intl";
+
+const messages = defineMessages({
+  addCategory: {
+    defaultMessage: "Add Category",
```
maarcingebala: Buttons in the UI use "Create <object>" convention and I think these actions should use the same convention. BTW once you're inside the "Create" view the page header usually says "Add <object>" - is this on purpose?

mateuszkula:
Comment on first impression.

Maybe it would be better if position of input was fixed. When writing down something the input is changing its position many times. 

Below is presentation what I mean:
Kapture 2019-11-26 at 11 08 17

maarcingebala:
@dominik-zeglen When typing an "operator" e.g. `#` I need to put additional space after it to start searching. It would be more natural to just type `#51` to go to order of that ID.

Second thought- it would be nice to somehow present the user available operators as they are not obvious. E.g. in VSCode you can type `?` to see the list of available operators. But this could be added later as well.

maarcingebala:
@mirekm you need to use `@` operator:
image

src/components/Navigator/modes/messages.ts
maarcingebala: OK it's a bug in the Customers section - it should say "Create Customer".

dominik-zeglen:
I've added help, which should improve discoverability of modes. Pic rel
Screenshot 2019-11-26 at 15 17 51

@mateuszkula thanks for feedback, I've pinned navigator so it won't change its place when typing.

PR #286, opened 2019-12-03
Fix error cleanup when switching product variants

### LINKED ISSUE #239: Input errors are not cleared when switching between variants
1. Go to variant details page. 
2. Edit the variant to cause an error, e.g. by using the same SKU or attribute values combination that already exists.
3. Switch to a different variant.
4. The input that failed is still rendered as error.

### What I expected to happen
Switching to a new variant should clear errors.

### Screenshots

The issue was caused for "500ml" option, here I'm viewing "2l":

image

tomaszszymanski129:
I want to merge this change because it fixes #239 

### DIFF
--- a/src/products/views/ProductVariant.tsx
+++ b/src/products/views/ProductVariant.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import { useIntl } from "react-intl";
 
 import placeholderImg from "@assets/images/placeholder255x255.png";
@@ -43,6 +43,7 @@ export const ProductVariant: React.FC<ProductUpdateProps> = ({
       require={["productVariant"]}
     >
       {({ data, loading }) => {
+        const [errors, setErrors] = useState([]);
         const variant = data ? data.productVariant : undefined;
         const handleBack = () => navigate(productUrl(productId));
         const handleDelete = () => {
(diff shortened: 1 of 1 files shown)


src/products/views/ProductVariant.tsx
```diff
@@ -146,6 +144,7 @@ export const ProductVariant: React.FC<ProductUpdateProps> = ({
                     }}
                     onVariantClick={variantId => {
                       navigate(productVariantEditUrl(productId, variantId));
+                      setErrors([]);
```
dominik-zeglen: I'd rather set it in `React.useEffect` - now errors persist if you go back using browser history.

src/products/views/ProductVariant.tsx
```diff
@@ -35,6 +35,10 @@ export const ProductVariant: React.FC<ProductUpdateProps> = ({
   const navigate = useNavigator();
   const notify = useNotifier();
   const intl = useIntl();
+  const [errors, setErrors] = useState([]);
+  useEffect(() => {
+    setErrors([]);
+  }, [variantId]);
```
dominik-zeglen: How about `useEffect(() => setErrors([]), [variantId]);`?
#285 will autofix this but it's not merged yet 🤷‍♂️

PR #326, opened 2019-12-16
bugfix on products page

FrBoss:
fixed bug when publishing or unpublishing a single product wont open the modal because a single id in the url parameters wont be recognized as an array

 
Didnt find a relevant issue

### Screenshots

No changes to the UI were made

### Pull Request Checklist

1. [ ] Translated strings are extracted to `.pot` file.


dominik-zeglen:
Thank you for your contribution. I investigated why it's happening and i found out that `src/products/urls.ts` uses `stringify` function with no indices, which means `ids` can be either `string` or `string[]`.
To be honest I don't know why it's done this way, but my guess is that the case I resolved with this fix does not exist anymore, because I tested it locally and it didn't break anything. 
So, summary: delete `src/utils/urls.ts` and import `stringifyQs` in `src/products/urls.ts` using `import { stringify as stringifyQs } from "qs";` :)

PR #411, opened 2020-02-25
Snapshot strict null mode error count

dominik-zeglen:
I want to merge this change because it helps us to keep track on how close we are to enable `strictNullCheck` flag in `tsconfig.json` for good.

### Pull Request Checklist


scripts/count-strict-null-check-errors.js
```diff
@@ -0,0 +1,29 @@
...
+  if (shouldCount.test(line)) {
+    errors++;
+  }
+});
+
+rl.on("close", () => {
+  if (updateSnapshot) {
+    fs.writeFileSync(snapshotPath, errors);
+  } else {
+    console.log(errors);
```
maarcingebala: We could use `console.error` here I guess.

PR #412, opened 2020-02-25
Check if number of errors in strict null check mode increased

dominik-zeglen:
I want to merge this change because it fails CI job if number of errors in strict null checking mode increased.

### Pull Request Checklist

### DIFF
--- a/.travis/check-strict-null-errors.sh
+++ b/.travis/check-strict-null-errors.sh
@@ -1,19 +1,54 @@
 #!/bin/sh
 
-snapshot=`cat .travis/check-strict-null-errors.snapshot` 
+snapshotPath=".travis/check-strict-null-errors.snapshot"
+
+snapshot=`cat $snapshotPath` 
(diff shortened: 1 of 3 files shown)


.travis/check-strict-null-errors.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+snapshot=`cat .travis/check-strict-null-errors.snapshot` 
+calculated=`npx tsc --noEmit --strictNullChecks | node scripts/count-strict-null-check-errors.js`
```
krzysztofwolski: @dominik-zeglen Can we save --noEmit and --strictNullChecks into a config file to make sure that linters will always have same config, no matter how configured?

dominik-zeglen: No, because webpack won't emit anything.
Also every build will fail because of `strictNullChecks`.

PR #452, opened 2020-03-18
Use translated error messages based on error codes

dominik-zeglen:
I want to merge this change because it allows us to display translated error messages in the application, instead of relying on backend generated ones. 

### Pull Request Checklist


src/utils/errors/order.ts
```diff
@@ -0,0 +1,106 @@
...
+    defaultMessage: "Shipping method is not valid for chosen shipping address",
+    description: "error message"
+  },
+  shippingRequired: {
+    defaultMessage: "Shipping method is required for this order",
+    description: "error message"
+  }
+});
+
+// cannot_delete
```
krzysztofwolski: ?

dominik-zeglen: Leftover comment 🤷‍♂️

dominik-zeglen:
```suggestion
```

PR #462, opened 2020-03-27
Fixing npm security

jtomaszon:


jtomaszon:
No need this dependency per #459

PR #471, opened 2020-04-07
Display validation errors on NewPassword page

### LINKED ISSUE #468: Password validation errors are not shown
### What I'm trying to achieve
To set password as a new staff user.

1. Create a new staff user through the dashboard.
2. Receive an email with the link to set up password and click it.
3. Provide a password that doesn't validate against password rules, e.g. too short.

### What I expected to happen
To see a validation error.

### Screenshots

image

gabmartinez:
I want to merge this change because it displays the validation errors on the New Password Page.

fix #468 

### Pull Request Checklist

### DIFF
--- a/src/auth/views/NewPassword.tsx
+++ b/src/auth/views/NewPassword.tsx
@@ -8,18 +8,12 @@ import NewPasswordPage, {
   NewPasswordPageFormData
 } from "../components/NewPasswordPage";
 import { SetPasswordMutation } from "../mutations";
-import {
-  SetPassword,
-  SetPassword_setPassword_errors
-} from "../types/SetPassword";
+import { SetPassword } from "../types/SetPassword";
 import { NewPasswordUrlQueryParams } from "../urls";
 
 const NewPassword: React.FC<RouteComponentProps> = ({ location }) => {
   const navigate = useNavigator();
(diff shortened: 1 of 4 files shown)


src/auth/views/NewPassword.tsx
```diff
@@ -8,19 +8,27 @@ import NewPasswordPage, {
...
+import {
+  SetPassword,
+  SetPassword_setPassword_errors
+} from "../types/SetPassword";
 import { NewPasswordUrlQueryParams } from "../urls";
 
 const NewPassword: React.FC<RouteComponentProps> = ({ location }) => {
   const navigate = useNavigator();
   const { loginByToken } = useUser();
+  const [errors, setErrors] = React.useState<
```
dominik-zeglen, line 20: The state here is not needed. `SetPasswordMutation` returns `setPasswordOpts`, which has `data` property, allowing us to access `data.setPassword.errors`, which we can directly convey to `NewPasswordPage` as prop. I'll look like this
```
<NewPasswordPage
  errors={setPasswordOpts.data.?.setPassword.errors || []}
  ...
/>
```

gabmartinez: You're right. done.

Thank you ;)

PR #480, opened 2020-04-15
The icons' colours are hardcoded in the SVG files

wael-mohamed:

The icons' colours were hardcoded in the SVG file, now they are set to the theme primary colour 

### Pull Request Checklist


now.json
```diff
@@ -0,0 +1,3 @@
+{
+    "routes": [{ "src": "/.*", "dest": "" }]
```
dominik-zeglen, line 2: I don't think this belongs here

wael-mohamed: Yes, you are it, I've fixed my error

PR #547, opened 2020-06-02
App details view

AlicjaSzu:

### Pull Request Checklist


src/apps/views/AppDetails/AppDetails.tsx
```diff
@@ -1,23 +1,94 @@
...
+  >(navigate, params => appUrl(id, params), params);
+
+  const [
+    openDeactivateModal,
+    closeDeactivateModal
+  ] = createDialogActionHandlers<AppDetailsUrlDialog, AppDetailsUrlQueryParams>(
+    navigate,
+    params => appUrl(id, params),
+    params
+  );
```
dominik-zeglen, line 47: You don't need to use `createDialogActionHandlers` for each dialog, the first callback has argument which is the name of the action, and like below, you can use it like this:
```
const [openModal, closeModal ] =  createDialogActionHandlers<AppDetailsUrlDialog, AppDetailsUrlQueryParams>(
    navigate,
    params => appUrl(id, params),
    params
  );

openModal("activate-app");
```

src/apps/views/AppDetails/AppDetails.tsx
```diff
@@ -1,23 +1,94 @@
...
+
+  useEffect(() => {
+    const { status, data } = activateAppResult;
+    if (status === "success" && data?.appActivate?.errors.length === 0) {
+      refetch();
+      closeActivateModal();
+    }
+  }, [activateAppResult.status]);
```
dominik-zeglen, line 55: Mutations accept object containing success and error handlers as an argument, so no need for `useEffect` here if we can do it more declarative way

PR #556, opened 2020-06-08
Add gh actions

krzysztofwolski:
...want to have same test stack on both repos.
...also want to LHCI to work properly.

Bonus script - messages extraction!
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build-lint-benchmark:
     docker:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build-lint-benchmark:
     docker:
-      - image: circleci/node:10.7.0-browsers
+      - image: circleci/node:14.4.0-browsers
```
krzysztofwolski, line 5: Had to bump node version to have lighthouse operational

PR #620, opened 2020-07-17
Add product export

dominik-zeglen:
I want to merge this change because it adds product export.

### Screenshots
Screenshot 2020-07-17 at 11 08 06
Screenshot 2020-07-17 at 11 08 25
Screenshot 2020-07-17 at 11 08 46

### Pull Request Checklist

### Test environment config


src/products/components/ProductExportDialog/ProductExportDialog.tsx
```diff
@@ -0,0 +1,175 @@
...
+                  selectedProducts={selectedProducts}
+                  onChange={change}
+                />
+              )}
+            </DialogContent>
+
+            {notFormErrors.length > 0 && (
+              <DialogContent>
+                {notFormErrors.map(err => (
+                  <Typography color="error">
```
krzysztofwolski, line 131: Add key

krzysztofwolski: (besides that LGTM)

IKarbowiak:
The issue fixed with https://github.com/mirumee/saleor/pull/5904#event-3575334917. Ready for re-testing.

maarcingebala:
Two things that I found:

- Does it make sense to have some way to select all checkboxes in "Information Exported" tab? Not necessarily to select all checkboxes in all different sections, but maybe one "Select all" per section? If I'm always exporting all information it may be annoying when I always have to select 10 or 15 checkboxes.

- Maybe it would be a good idea to show number of exported products in all three cases, not only for "Selected products":
image
I find it useful to see how many products will be exported.

What do you think @pwgryglak?

pwgryglak:
@maarcingebala @dominik-zeglen 
1) Yes each of these should have "select all" option as on design:
Screenshot 2020-07-24 at 14 02 59

2) Yes, we should add that. Will update designs.

michalina-graczyk:
Update 27.07.20 - Issue 3:

1. More than 4 attributes look badly:
image
2. Missed "add all attributes" for the checkbox for attributes 
image

3. Only attributes have "tags", rest of the exported information doesn't look like the attributes tags - @dominik-zeglen could you explain that differences?

dominik-zeglen:
2. Backend is unable to do this so we'll have to wait until we can do it
3. Added chips to all sections
4. Sections are now tagged

Export is not working on staging because of some bug on the backend side.

maarcingebala:
Ad 2 - can't we have a dummy checkbox that selects all attributes that are on the list?

dominik-zeglen:
It won't behave as you'd expect, so it's better to add it later 🤷‍♂️

PR #685, opened 2020-09-02
Fix/container size

eaglesemanation:
I want to merge this change because it reduces size of Docker images for both development and production. Also it removes some inconsistencies in build process between this project and saleor-storefront.

### Screenshots

container-size

### Pull Request Checklist

### Test environment config


krzysztofwolski (changes requested):

krzysztofwolski:
@NyanKiyoshi I need help from a docker/devops wizard. Could you take a look at the proposed changes? I'm not sure if your team use base docker img for dashboard deployment.

krzysztofwolski:
@eaglesemanation Thank you for PR! Size reduction looks impressive. We have to validate internally proposed changes and check if it plays well with saleor-platform repo.

Dockerfile
```diff
@@ -1,17 +1,38 @@
...
+ENV STATIC_URL ${STATIC_URL:-/}
+ENV PATH=/app/node_modules/.bin:$PATH
 WORKDIR /app
-COPY ./nginx/default.conf /etc/nginx/conf.d/default.conf
-COPY --from=builder /app/build/ /app/
+COPY --chown=node:node package*.json ./
+RUN npm install
+
+FROM base as dev
+CMD npm start
```
NyanKiyoshi, line 16: If you want it to be a specific layer that users can use, I think it should be documented somewhere

Dockerfile
```diff
@@ -1,17 +1,38 @@
...
+
+FROM base as source
+COPY --chown=node:node assets ./assets/
+COPY --chown=node:node locale ./locale/
+COPY --chown=node:node testUtils ./testUtils/
+COPY --chown=node:node src ./src/
+COPY --chown=node:node *.js ./
+COPY --chown=node:node *.json ./
+COPY --chown=node:node *.ts ./
```
NyanKiyoshi, line 25: Rather than having one layer per type of files, you could only create one:
```dockerfile
COPY --chown=uid:gid \
        assets \
        locale \
        testUtils \
        etc. \
        /app/
```

NyanKiyoshi (commented):
I'm not fully hyped on the idea of giving the user the choice between multiple layers they can use... It's not straightforward and not user-friendly.

I would prefer having a base image and multiple variants:
- saleor-dashboard:latest-dev (`npm start`)
- saleor-dashboard:latest-static (`npm run build`)
- saleor-dashboard:latest-nginx (latest-static + `nginx`)

Dockerfile
```diff
@@ -1,17 +1,38 @@
...
-COPY ./nginx/default.conf /etc/nginx/conf.d/default.conf
-COPY --from=builder /app/build/ /app/
+COPY --chown=node:node package*.json ./
+RUN npm install
+
+FROM base as dev
+CMD npm start
+
+FROM base as source
+COPY --chown=node:node assets ./assets/
```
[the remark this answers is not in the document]
eaglesemanation: I feel like `.dockerignore` is abused. It ignores files when copies context to docker daemon. I wanted to copy nginx files separately, so I could modify it without breaking cached `source` stage. But because I still need it I can't add it to `.dockerignore`. Maybe I'm wrong about that, I'm not sure.

Dockerfile
eaglesemanation: Didn't know that, fixed

Dockerfile
```diff
@@ -1,17 +1,38 @@
...
+COPY --chown=node:node locale ./locale/
+COPY --chown=node:node testUtils ./testUtils/
+COPY --chown=node:node src ./src/
+COPY --chown=node:node *.js ./
+COPY --chown=node:node *.json ./
+COPY --chown=node:node *.ts ./
+
+FROM source as test
+# AuthProvider.test.ts breaks testing during build,
```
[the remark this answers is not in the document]
eaglesemanation: Agree, I would use it as my own CI in k8s cluster, but it shouldn't be upstream. Removed.

eaglesemanation:
> I'm not fully hyped on the idea of giving the user the choice between multiple layers they can use... It's not straightforward and not user-friendly.
But it helps with caching same stages between different builds. I'm not sure why adding `--target dev` is not straightforward, especially if I would document it. Only problem is that I don't know where to add documentation.

NyanKiyoshi:
> > I'm not fully hyped on the idea of giving the user the choice between multiple layers they can use... It's not straightforward and not user-friendly.
> 
> But it helps with caching same stages between different builds. I'm not sure why adding `--target dev` is not straightforward, especially if I would document it. Only problem is that I don't know where to add documentation.

Then it's fine! I didn't know about the `--target` flag, I just checked the docs and it sounds good to me, I will check it out tomorrow.

eaglesemanation:
> > > I'm not fully hyped on the idea of giving the user the choice between multiple layers they can use... It's not straightforward and not user-friendly.
> > 
> > 
> > But it helps with caching same stages between different builds. I'm not sure why adding `--target dev` is not straightforward, especially if I would document it. Only problem is that I don't know where to add documentation.
> 
> Then it's fine! I didn't know about the `--target` flag, I just checked the docs and it sounds good to me, I will check it out tomorrow.

Docker compose also has targets in build, starting from version 2.3 (docker engine 17.06, from 2017), so it shouldn't break saleor-platform.

PR #696, opened 2020-09-07
fix product availability date

AlicjaSzu:

### Screenshots

Screenshot 2020-09-07 at 14 48 44

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/components/VisibilityCard/VisibilityCard.tsx
+++ b/src/components/VisibilityCard/VisibilityCard.tsx
@@ -239,11 +239,12 @@ export const VisibilityCard: React.FC<VisibilityCardProps> = props => {
               value={isAvailable}
               onChange={e => {
                 const { value } = e.target;
-                if (!value) {
-                  onChange({
-                    target: { name: "availableForPurchase", value: null }
-                  });
-                }
+                onChange({
+                  target: {
+                    name: "availableForPurchase",
+                    value: value ? localizeDate(new Date().toString()) : null
+                  }
+                });
                 return onChange(e);
               }}
             />
--- a/src/components/VisibilityCard/VisibilityCard.tsx
+++ b/src/components/VisibilityCard/VisibilityCard.tsx
@@ -239,12 +239,14 @@ export const VisibilityCard: React.FC<VisibilityCardProps> = props => {
               value={isAvailable}
               onChange={e => {
                 const { value } = e.target;
-                onChange({
-                  target: {
-                    name: "availableForPurchase",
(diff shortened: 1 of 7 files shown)


src/components/VisibilityCard/VisibilityCard.tsx
```diff
@@ -239,11 +239,12 @@ export const VisibilityCard: React.FC<VisibilityCardProps> = props => {
...
                 const { value } = e.target;
-                if (!value) {
-                  onChange({
-                    target: { name: "availableForPurchase", value: null }
-                  });
-                }
+                onChange({
+                  target: {
+                    name: "availableForPurchase",
+                    value: value ? localizeDate(new Date().toString()) : null
```
maarcingebala, line 245: `productSetAvailabilityForPurchase` mutation in the response returns the product with `availableForPurchase` set by the backend. The current logic assumes that we set here the current date if `null` was passed as `startDate` in the mutation. If we change this logic in the backend, `new Date()` here will be invalid, won't it? Can we use the date returned by API here?

AlicjaSzu: yes we can, I wanted to avoid refetching all product data only for such small thing

orzechdev (approved):

dominik-zeglen (changes requested):
If you go to the `src/products/mutations.ts` and modify `productSetAvailabilityForPurchase` to return 
```
product {
        id
        isAvailableForPurchase
      }
```
there wouldn't be any need for refetching whole product.

src/products/views/ProductUpdate/ProductUpdate.tsx
```diff
@@ -177,13 +178,19 @@ export const ProductUpdate: React.FC<ProductUpdateProps> = ({ id, params }) => {
...
+          availableForPurchase: updatedProduct.availableForPurchase
+        }));
         notify({
           status: "success",
           text: intl.formatMessage({
             defaultMessage: "Product availability updated",
             description: "snackbar text"
           })
         });
+        refetch();
```
dominik-zeglen, line 193: It doesn't update by itself?

dominik-zeglen (changes requested):
Mutation to update availability should not be fired if previous failed.

src/products/views/ProductUpdate/ProductUpdate.tsx
AlicjaSzu: I forgot to remove it.

krzysztofwolski (approved):

orzechdev (approved):

PR #842, opened 2020-11-19
[1516] Create product only when all required fields are valid

jwm0:
Since product creation is split into multiple mutations and some of the required fields are scattered throughout them, I propose to only create the product when validation of all of them is successful. Due to mutations needing to be fired in order we have to first create a product to fire up the rest of them, which always results in a product being created and therefore the form state was lost. To avoid a larger refactor we remove the previously created product if the subsequent mutations fail, this way we don't duplicate products and the UI requirements are matched with the inner logic.

https://app.clickup.com/t/2549495/SALEOR-1516

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/products/views/ProductCreate/ProductCreate.tsx
+++ b/src/products/views/ProductCreate/ProductCreate.tsx
@@ -7,7 +7,6 @@ import useChannels from "@saleor/hooks/useChannels";
 import useNavigator from "@saleor/hooks/useNavigator";
 import useNotifier from "@saleor/hooks/useNotifier";
 import useShop from "@saleor/hooks/useShop";
-import { getMutationStatus } from "@saleor/misc";
 import ProductCreatePage from "@saleor/products/components/ProductCreatePage";
 import {
   useProductChannelListingUpdate,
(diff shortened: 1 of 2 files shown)


patrys:
Task linked: [SALEOR-1516 [Create Product] Values are missing after validation error](https://app.clickup.com/t/2549495/SALEOR-1516)

dominik-zeglen (changes requested):
I believe that all this logic should be contained in product create handler rather than in component

src/products/views/ProductCreate/ProductCreate.tsx
```diff
@@ -133,17 +133,32 @@ export const ProductCreateView: React.FC = () => {
...
+    }
+  };
+
+  React.useEffect(() => {
+    const productId = productCreateOpts.data?.productCreate?.product?.id;
+
+    if (productCreateComplete && productId) {
+      handleSuccess(productId);
+    }
+  }, [productCreateComplete]);
```
dominik-zeglen, line 161: I think we can add this to handler too and skip `productCreateComplete` state from component 🤷‍♂️

dominik-zeglen (changes requested):

dominik-zeglen (approved):

mlamperski:
Tested. Issue doesn't occure on deployed environment.

orzechdev (approved):

PR #903, opened 2020-12-14
Stitch returns form

mmarkusik:
I want to merge this change because it stitches the order return form with api 

### Pull Request Checklist

### Test environment config

API_URI=https://feature-refund-products.api.saleor.rocks/graphql/


patrys:
Task linked: [SALEOR-1874 Stitch returns page](https://app.clickup.com/t/2549495/SALEOR-1874)

krzysztofwolski (approved):

src/orders/components/OrderRefundReturnAmount/OrderRefundReturnAmount.tsx
```diff
@@ -240,14 +240,13 @@ const OrderRefundAmount: React.FC<OrderRefundAmountProps> = props => {
...
         <Button
-          type="submit"
           color="primary"
           variant="contained"
           fullWidth
           size="large"
           onClick={onRefund}
           className={classes.refundButton}
-          disabled={disabled || disableRefundButton}
+          disabled={false}
```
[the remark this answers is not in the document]
mmarkusik: Actually yes. Noticed a bug here, when calculating the refund for items when the whole order was refunded - will be added in fixes pr along with some other stuff, so for now just allowed it for development

src/orders/views/OrderReturn/utils.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return itemsQuantities.reduce((result, { value: quantity, id }) => {
+    if (!quantity) {
+      return result;
+    }
+
+    const shouldReplace = !!itemsToBeReplaced.find(getById(id))?.value;
+
+    return [
+      ...result,
+      ({ [idKey]: id, quantity, replace: shouldReplace } as unknown) as T
```
orzechdev, line 63: Why is it treated as `unknown`? Returned objects have strictly defined type (has `[idKey]`, `quantity` and `replace` keys), while `unknown` allows to pass arbitrary generic T type what may cause that returned value is not consistent with T type. In other words typescript will not warn if you write `getParsedLineData<{ anything: any }>`, while it should, as `anything` doesn't exist in returned objects.

Maybe restricting allowed types to `getParsedLineData<T extends OrderReturnLineInput | OrderReturnFulfillmentLineInput>` may help?

mmarkusik: Good point! Unfortunately the `uknown` conversion was still needed by ts, but indeed now when initialing `getParsedLineData<{foo: boolean}>` typescript throws an error.

orzechdev (approved):

PR #906, opened 2020-12-14
Saleor 1855/implement new cards for return statuses

mmarkusik:
I want to merge this change because it adds new cards - Returned, replaced. Also lots of refactor to reduce boilerplate. 

Screenshot 2020-12-15 at 14 36 23

Screenshot 2020-12-15 at 14 23 42

[SEJLOR-1855](https://app.clickup.com/t/2549495/SALEOR-1855)

### Pull Request Checklist

### Test environment config

API_URI=https://feature-refund-products.api.saleor.rocks/graphql/


patrys:
Task linked: [SALEOR-1855 Implement Returned and Returned & Refunded Cards](https://app.clickup.com/t/2549495/SALEOR-1855)

src/orders/components/OrderFulfillment/TableLine.tsx
```diff
@@ -0,0 +1,124 @@
...
+        className={classes.colName}
+        thumbnail={maybe(() => line.orderLine.thumbnail.url)}
+      >
+        {maybe(() => line.orderLine.productName) || <Skeleton />}
+      </TableCellAvatar>
+      <TableCell className={classes.colSku}>
+        {line?.orderLine.productSku || <Skeleton />}
+      </TableCell>
+      <TableCell className={classes.colQuantity}>
+        {line?.quantity || <Skeleton />}
```
krzysztofwolski, line 99: quantity is string or number? Wonder if skeleton will be rendered if quantity is 0

src/orders/components/OrderFulfillment/TableLine.tsx
```diff
@@ -0,0 +1,124 @@
...
+  }),
+  { name: "TableLine" }
+);
+
+interface TableLineProps {
+  line: OrderDetails_order_fulfillments_lines | OrderDetails_order_lines;
+  isOrderLine?: boolean;
+}
+
+const TableLine: React.FC<TableLineProps> = ({
```
krzysztofwolski, line 70: Maybe FullfillmentTableLine would be more descriptive name? Helpful especially when using auto-import for code completion.

src/orders/components/OrderFulfillment/TableLine.tsx
mmarkusik: Tbh noticed this (and some other stuff like that in this component) but didn't check, because this is just copied pasted from the main component. I wanted to wait with more refactoring of this because seems there's a situation that the designs don't cover - when a fulfillment line in a fulfillment has no orderline, because the purchased item has been removed from the shop DB altogether. So waiting for the designs from @pwgryglak and then I think it'll make more sense to refactor this component.

src/orders/components/OrderFulfillment/TableLine.tsx
mmarkusik: Yeeees...and no. It's a table line for both fulfillment and order line, and since in most of the code naming something 'fulfillment' maps directly to actual fulfillment, would rather not call it that cause then it would be misleading. But will move the table elements to some suitable directory so the whole dir will be named correctly

orzechdev (approved):

src/orders/components/OrderFulfillment/TableLine.tsx
krzysztofwolski: Then OrderTableRow would be good, since it cover different states (fullfillment, etc). My gripe is that we have Customer tables, attribute tables, etc

krzysztofwolski (approved):

src/orders/components/OrderFulfillment/TableLine.tsx
mmarkusik: Changed to *even better ones*

PR #921, opened 2020-12-21
Show hidden radio buttons of shipping method zip codes

tomaszszymanski129:
I want to merge this change because it shows the radio buttons on shipping method details in zip codes sections.
Why? As a user I have no idea in current solution what these zip codes range does. And it supports only exclude for now which isn't displayed to the user anywhere.

Before:
image
After:
image

### Pull Request Checklist

### Test environment config


jwm0 (changes requested):
Looking at it I would suggest inversing the disabled style on the radio inputs. The "Include ZIP-codes" option should be the one disabled. The "exclude" one could be enabled, so we better communicate what's implemented now and what's coming later.

tomaszszymanski129:
@jwm0 good point, looks better now:
image

jwm0 (approved):

krzysztofwolski (approved):

PR #924, opened 2020-12-28
Saleor 1712 dashboard unable to handle empty channel list

MarekChoinski:
I want to merge this change because... it handles empty channels on:
- homepage
- `/dashboard/channels`
- `/dashboard/orders`
- `/dashboard/products`

NOTE: Tested by mocking up response to `[{"data": {"channels": []}}]` on `AppChannelContext.tsx`.

### Screenshots

image

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/categories/views/CategoryDetails.tsx
+++ b/src/categories/views/CategoryDetails.tsx
@@ -6,7 +6,6 @@ import useAppChannel from "@saleor/components/AppLayout/AppChannelContext";
 import NotFoundPage from "@saleor/components/NotFoundPage";
 import Skeleton from "@saleor/components/Skeleton";
 import { WindowTitle } from "@saleor/components/WindowTitle";
-import { channelDetailsFragment } from "@saleor/fragments/channels";
 import useBulkActions from "@saleor/hooks/useBulkActions";
 import useNavigator from "@saleor/hooks/useNavigator";
 import useNotifier from "@saleor/hooks/useNotifier";
--- a/src/categories/views/CategoryDetails.tsx
+++ b/src/categories/views/CategoryDetails.tsx
@@ -4,7 +4,9 @@ import DeleteIcon from "@material-ui/icons/Delete";
 import ActionDialog from "@saleor/components/ActionDialog";
 import useAppChannel from "@saleor/components/AppLayout/AppChannelContext";
 import NotFoundPage from "@saleor/components/NotFoundPage";
+import Skeleton from "@saleor/components/Skeleton";
 import { WindowTitle } from "@saleor/components/WindowTitle";
+import { channelDetailsFragment } from "@saleor/fragments/channels";
 import useBulkActions from "@saleor/hooks/useBulkActions";
 import useNavigator from "@saleor/hooks/useNavigator";
 import useNotifier from "@saleor/hooks/useNotifier";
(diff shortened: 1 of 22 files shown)


src/home/views/index.tsx
```diff
@@ -8,57 +9,64 @@ import { orderListUrl } from "../../orders/urls";
...
 import HomePage from "../components/HomePage";
-import { HomePageQuery } from "../queries";
+import { useHomePage } from "../queries";
 
 const HomeSection = () => {
   const navigate = useNavigator();
   const { user } = useUser();
   const { channel } = useAppChannel();
 
+  const noChannel = !channel && typeof channel !== "undefined";
```
mmarkusik, line 19: I'd rather see this as a positive const, such as `channelExists` or `hasChannel`, especially since most places we use it, it's negated. Besides that it seems easier to overall think of positives +- negation, rather than negatives +- negation (that more easily created negated negative), don't you think?

src/home/views/index.tsx
```diff
@@ -8,57 +9,64 @@ import { orderListUrl } from "../../orders/urls";
...
-          ordersToFulfill={maybe(() => data.ordersToFulfill.totalCount)}
-          productsOutOfStock={maybe(() => data.productsOutOfStock.totalCount)}
-          userName={getUserName(user, true)}
-          userPermissions={user?.userPermissions || []}
-        />
-      )}
-    </HomePageQuery>
+    <HomePage
+      activities={data?.activities.edges.map(edge => edge.node).reverse()}
+      orders={!noChannel ? data?.ordersToday.totalCount : 0}
```
mmarkusik, line 31: I think in this case HomePage component should have a default when orders is undefined. Then it also doesn't require to check if `!noChannel`

src/home/views/index.tsx
```diff
@@ -8,57 +9,64 @@ import { orderListUrl } from "../../orders/urls";
...
+      sales={maybe(() => data?.salesToday.gross)}
+      topProducts={
+        !noChannel ? data?.productTopToday.edges.map(edge => edge.node) : null
```
mmarkusik, line 34: Same as above

src/home/views/index.tsx
```diff
@@ -8,57 +9,64 @@ import { orderListUrl } from "../../orders/urls";
...
+        )
+      }
+      onProductsOutOfStockClick={() =>
+        navigate(
+          productListUrl({
+            stockStatus: StockAvailability.OUT_OF_STOCK
+          })
+        )
+      }
+      ordersToCapture={!noChannel ? data?.ordersToCapture.totalCount : 0}
```
mmarkusik, line 63: 63-65 same as above

src/home/views/index.tsx
```diff
@@ -8,57 +9,64 @@ import { orderListUrl } from "../../orders/urls";
...
+      ordersToFulfill={!noChannel ? data?.ordersToFulfill.totalCount : 0}
+      productsOutOfStock={!noChannel ? data?.productsOutOfStock.totalCount : 0}
+      userName={getUserName(user, true)}
+      userPermissions={user?.userPermissions || []}
```
mmarkusik, line 67: Also should be default in home page

src/index.tsx
```diff
@@ -145,7 +145,7 @@ const Routes: React.FC = () => {
   return (
     <>
       <WindowTitle title={intl.formatMessage(commonMessages.dashboard)} />
-      {channel &&
+      {typeof channel !== "undefined" &&
```
mmarkusik, line 148: Would happily see those 4 &&'s moved to a const with a nice name

src/orders/views/OrderList/OrderList.tsx
```diff
@@ -71,6 +71,8 @@ export const OrderList: React.FC<OrderListProps> = ({ params }) => {
 
   const { channel, availableChannels } = useAppChannel();
 
+  const noChannel = !channel && typeof channel !== "undefined";
```
mmarkusik, line 74: Same as before

src/home/components/HomePage/HomePage.tsx
```diff
@@ -99,7 +103,9 @@ const HomePage: React.FC<HomePageProps> = props => {
                   />
                 }
               >
-                {sales ? (
+                {noChannel ? (
```
mmarkusik, line 106: Maybe if there are three outcomes here, it'd be worth to just move it to a function? Chaining conditions makes it less readable. Also, would rather see this in positive, like channelExists + negation

src/home/views/index.tsx
```diff
@@ -8,57 +9,64 @@ import { orderListUrl } from "../../orders/urls";
...
 
 const HomeSection = () => {
   const navigate = useNavigator();
   const { user } = useUser();
   const { channel } = useAppChannel();
 
+  const noChannel = !channel && typeof channel !== "undefined";
+
+  const { data } = !noChannel
+    ? useHomePage({
```
orzechdev, line 22: I'm afraid conditionally using hook might cause React errors in the future - https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

mmarkusik (changes requested):

orzechdev (changes requested):

src/categories/views/CategoryDetails.tsx
```diff
@@ -210,166 +211,179 @@ export const CategoryDetails: React.FC<CategoryDetailsProps> = ({
...
-        onCategoryClick={id => () => navigate(categoryUrl(id))}
-        onDelete={() => openModal("delete")}
-        onImageDelete={() =>
-          updateCategory({
-            variables: {
-              id,
-              input: {
-                backgroundImage: null
-              }
+      {typeof channel !== "undefined" ? (
```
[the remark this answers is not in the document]
MarekChoinski: Nope, because `undefined` means that channels are not loaded yet and `null` means that there are no channels.

`!channels === (channels === null || channels === 'undefined')`, but `null !== 'undefined'`

src/home/views/index.tsx
MarekChoinski: I left `noChannel` after the refactor in which I added default props in `HomePage`. In this component it's mostly positive value. What do you think?

src/home/views/index.tsx
```diff
@@ -1,64 +1,69 @@
...
   const navigate = useNavigator();
   const { user } = useUser();
   const { channel } = useAppChannel();
 
+  const noChannel = !channel && typeof channel !== "undefined";
+
+  const { data } = useHomePage({
+    displayLoader: true,
+    variables: { channel: channel?.slug },
+    skip: false
```
orzechdev, line 24: I think query is not skipped by default, so passing `skip: false` seems to be redundant. If the query shouldn't be made, when the channel isn't available, can it be `skip: noChannel`?

MarekChoinski: Ah yes, sorry. I forgot to change it after testing

mmarkusik (approved):

orzechdev (approved):

jwm0 (changes requested):
I clicked through the dashboard and there's a lot of pages that do not work because they "cannot read property id of undefined". Please check where we use `channel.id` and replace it with `channel?.id`. Products/categories/orders/discounts should be accessible

jwm0 (approved):

PR #935, opened 2021-01-07
Fix bug with uncomplete publish date

MarekChoinski:
I want to merge this change because... it fixes error when trying to save uncomplete publish date.

**Explanation:**
Empty `publicationDate` should be `""` in `ChannelsAvailability` component, but need to be sent as `null` in query.

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/components/ChannelsAvailability/ChannelsAvailability.tsx
+++ b/src/components/ChannelsAvailability/ChannelsAvailability.tsx
@@ -230,7 +230,7 @@ const Channel: React.FC<ChannelProps> = ({
                           )
                         : ""
                     }
-                    value={publicationDate}
+                    value={publicationDate || ""}
                     onChange={e =>
                       onChange(id, {
                         ...formData,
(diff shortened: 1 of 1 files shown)


src/components/ChannelsAvailability/ChannelsAvailability.tsx
```diff
@@ -230,11 +230,11 @@ const Channel: React.FC<ChannelProps> = ({
                           )
                         : ""
                     }
-                    value={publicationDate ? publicationDate : ""}
+                    value={publicationDate}
```
jwm0, line 233: If `null` is a valid value we should fallback to empty string `""` like before, otherwise we might get the infamous React warning "A component is changing an uncontrolled input..."

```suggestion
                    value={publicationDate ? publicationDate : ""}
```

jwm0 (changes requested):

src/components/ChannelsAvailability/ChannelsAvailability.tsx
MarekChoinski: It looks like it was working fine without default `""`, but I can't find any fallback for that. I will change it anyway because otherwise it may have unexpected behavior in the future

jwm0 (approved):

PR #946, opened 2021-01-13
1863 - Support reference attribute reordering

orzechdev:
I want to merge this change because... it adds support for reference attribute reordering.

Feature branch: https://github.com/mirumee/saleor-dashboard/pull/917

**PR intended to be tested with API branch:**  feature/page-reference-attributes
https://github.com/mirumee/saleor/pull/6624

### Pull Request Checklist

### Test environment config


src/attributes/utils/handlers.ts
```diff
@@ -17,6 +26,99 @@ import {
 } from "../types/AttributeValueDelete";
 import { getFileValuesToUploadFromAttributes, isFileValueUnused } from "./data";
 
+export function createAttributeChangeHandler(
+  changeAttributeData: FormsetChange<string[]>,
+  triggerChange: () => void
+): FormsetChange<string> {
+  return (attributeId: string, value: string) => {
+    triggerChange();
+    changeAttributeData(attributeId, value === "" ? [] : [value]);
```
mmarkusik, line 35: maybe instead 
```suggestion
    changeAttributeData(attributeId, !!value ? [] : [value]);
```

mmarkusik (approved):

src/attributes/utils/handlers.ts
orzechdev: I just moved code from one place to another and it needs more refactor, as there are cases, which use array null or undefined as array value. I know it seems to be wrong, but I think I'll change it in other PR.

dominik-zeglen (approved):

PR #966, opened 2021-01-27
Saleor 1736 configuration channels

karola312:
I want to merge this change because I created tests for channels

### Pull Request Checklist

### Test environment config


cypress/elements/catalog/product-selectors.js
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable sort-keys */
 export const PRODUCTS_SELECTORS = {
+  productsList: "[data-test-id][data-test='id']",
```
krzysztofwolski, line 3:
```suggestion
  productsList: "[data-test='id']",
```

`[data-test-id]` is redundant here IMO

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
krzysztofwolski, line 1: FIle can be removed

cypress/integration/channels.js
```diff
@@ -0,0 +1,152 @@
...
+    cy.location()
+      .should(loc => {
+        const urlRegex = new RegExp(`${URL_LIST.orders}/.+`, "g");
+        expect(loc.pathname).to.match(urlRegex);
+      })
+      .get(DRAFT_ORDER_SELECTORS.addProducts)
+      .should("not.exist");
+  });
+
+  function createChannel(name, otherCurrency, slug = name) {
```
krzysztofwolski, line 132: Such helper functions should be in separate utils file. 
Performance: creating channel using GUI (cy.click etc) should be done once in test for channel creation view. If you want to create channel to perform other tests on it, its better to just run direct API query - it will be faster.

package.json
```diff
@@ -227,5 +229,7 @@
     "test": "jest src/",
     "transpile-messages": "node scripts/transpile-tx.js",
     "lint": "npx eslint \"src/**/*.@(tsx|ts|jsx|js)\" --fix ; npx prettier --check \"src/**/*.@(tsx|ts|jsx|js)\" --write"
-  }
+  },
+  "description": "![Saleor Dashboard](https://user-images.githubusercontent.com/249912/82305745-5c52fd00-99be-11ea-9ac6-cc04a6f28c91.png)",
```
krzysztofwolski, line 233: Was it added on purppose?

krzysztofwolski (changes requested):

cypress/integration/channels.js
karola312: This function is used in three tests, only in one of them channel will be created. Two other tests are using it to try to create channel and in result validation messages will show up, so I can't do it through api. This function will be used only in this file, so I didn't want to move it to separate file. Maybe I should rename function to tryToCreateChannel() ?

cypress/elements/catalog/product-selectors.js
karola312: [data-test-id] is necessary. Before all products are loaded there is an empty row on products table with [data-test='id']. Without [data-test-id] once in a while test will crush because it will try to click on that empty row.

package.json
karola312: No, I will remove it

cypress/integration/channels.js
mmarkusik: I agree with @krzysztofwolski as to moving it outside of this file (utils ?) - it's not a problem now, but once we add more tests and more utils functions it'll become more and more difficult to navigate through those files, since it's not a place for utils function to be in a test file.

cypress/url/url-list.js
```diff
@@ -0,0 +1,6 @@
+export const URL_LIST = {
+  dashbord: "/",
+  channels: "/channels/",
+  products: "/products/",
+  orders: "/orders"
```
mmarkusik, line 5: Looking at the urls above, shouldn't `/orders` also have a `/` at the end?

mmarkusik (changes requested):

cypress/integration/channels.js
```diff
@@ -0,0 +1,134 @@
...
+      .click()
+      .visit(URL_LIST.products)
+      .waitForGraph("InitialProductFilterData");
+    cy.get(PRODUCTS_SELECTORS.productsList)
+      .first()
+      .click()
+      .get(PRODUCTS_SELECTORS.availableManageButton)
+      .click()
+      .get(PRODUCTS_SELECTORS.channelsAvailabilityForm)
+      .contains(randomChannel);
```
krzysztofwolski, line 65: Few comments would help maintain test in the future

```suggestion
    // New channel should be visible in channels list
    cy.waitForGraph("Channel")
      .get(ADD_CHANNEL_FORM_SELECTOS.backToChannelsList)
      .click()
      .get(CHANNELS_SELECTORS.channelsTable)
      .contains(randomChannel)
      
    // new channel should be visible in channel selector
    cy.visit(URL_LIST.dashbord)
      .get(HEADER_SELECTORS.channelSelect)
      .click()
      .get(HEADER_SELECTORS.channelSelectList)
      .contains(randomChannel)
      .click()
      
    // new channel should be visible at product availability form
    cy.visit(URL_LIST.products)
      .waitForGraph("InitialProductFilterData");
    cy.get(PRODUCTS_SELECTORS.productsList)
      .first()
      .click()
      .get(PRODUCTS_SELECTORS.availableManageButton)
      .click()
      .get(PRODUCTS_SELECTORS.channelsAvailabilityForm)
      .contains(randomChannel);
```

krzysztofwolski (changes requested):

mmarkusik:
> view on LGTM.com

If this `new Set` actually always resolves to true then can we remove it?

mmarkusik (approved):
Besides cypress failing all is good

krzysztofwolski (approved):

jwm0 (changes requested):
Why are the brand new tests failing on the pipeline? We shouldn't merge those as they add no value if they're not passing in the original branch. Please fix or explain why they're failing and what needs to be done to get them working.

PR #977, opened 2021-02-11
Saleor 1738 tests for home page

karola312:
I want to merge this change because I created tests for dashboard

### Pull Request Checklist

1. [ ] The changes are tested in different browsers (Chrome, Firefox, Safari).
1. [ ] The changes are tested in light and dark mode.

### Test environment config


krzysztofwolski (changes requested):

cypress/integration/dashboard.js
```diff
@@ -0,0 +1,143 @@
...
+        });
+    });
+    cy.visit("/");
+    cy.get(HEADER_SELECTORS.channelSelect)
+      .click()
+      .get(HEADER_SELECTORS.channelSelectList)
+      .contains(randomName)
+      .click()
+      .waitForGraph("Home");
+    const salesRegexp = new RegExp(`\\D+${sales}[\,.]00\\D*`);
```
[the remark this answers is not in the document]
karola312: Sales is exact value. Regex is used because on my local i get "PLN 56.00", on staging, "56,00 zł". I want this test to pass regardless of whether if there will be comma, or dot, "zł" or "PLN".

cypress/integration/dashboard.js
```diff
@@ -0,0 +1,143 @@
...
+    cy.fixture("addresses").then(json => {
+      channels
+        .createChannel(true, randomName, randomName, json.plAddress.currency)
+        .then(channelsResp => {
+          const channelId = channelsResp.body.data.channelCreate.channel.id;
+          const channelSlug = channelsResp.body.data.channelCreate.channel.slug;
+          customer
+            .createCustomer(randomEmail, randomName, json.plAddress)
+            .then(resp => {
+              const customerId = resp.body.data.customerCreate.user.id;
```
mmarkusik: These could be done with 
```suggestion
              const { user: { id : customerId, email: customerEmail } } = resp.body.data.customerCreate
```

cypress/utils/ordersUtils.js
```diff
@@ -0,0 +1,46 @@
...
+  createWaitingForCaptureOrder(
+    channelSlug,
+    email,
+    variantsList,
+    shippingMethodId
+  ) {
+    const checkout = new Checkout();
+    return checkout
+      .createCheckout(channelSlug, email, 1, variantsList)
+      .then(createCheckoutResp => {
```
mmarkusik, line 14: Would also rather see these as asyns and awaits

cypress/utils/productsUtils.js
```diff
@@ -0,0 +1,124 @@
...
+    const category = new Category();
+    product.getProductTypes(100, startsWith).then(resp => {
+      const productTypes = resp.body.data.productTypes.edges;
+      productTypes.forEach(productType => {
+        if (productType.node.name.includes(startsWith)) {
+          product.deleteProductType(productType.node.id);
+        }
+      });
+    });
+    attribute.getAttributes(100, startsWith).then(resp => {
```
mmarkusik, line 98: These parts are copied pasted. I'd suggest making a class of it, or at least
```suggestion
    const ELEMENTS_AMOUNT = 100

deleteProperCategories = (startsWith) =>  {
    const res => category.getCategories(ELEMENTS_AMOUNT, startsWith)
    const { edges : categories } = res.body.data.categories

    const deleteElement = handleDeleteElement(category.deleteCategory)

    categories.forEach(element => deleteElement(categories, startsWith))
}

handleDeleteElement = (deleteFunction) => (element, startsWith) => {
    if(!isAmongDeletableElements(element, startsWith)) { return; }

    deleteFunction(element.node.id)
}

isAmongDeletableElements = (element, startsWith) => element.node.name.includes(startsWith)
```

cypress/utils/shippingUtils.js
```diff
@@ -0,0 +1,71 @@
...
+import Warehouse from "../apiRequests/Warehouse";
+class ShippingUtils {
+  shippingMethodId;
+  shippingZoneId;
+  warehouseId;
+
+  createShipping(channelId, name, address, price) {
+    const shippingMethod = new ShippingMethod();
+    const warehouse = new Warehouse();
+    return shippingMethod
```
mmarkusik, line 11: We generally try to avoid then chaining to avoid callback hell - feel free to check [this](https://blog.hellojs.org/asynchronous-javascript-from-callback-hell-to-async-and-await-9b9ceb63c8e8) out

mmarkusik (changes requested):

cypress/integration/homePage.js
```diff
@@ -0,0 +1,210 @@
...
+
+  const customer = new Customer();
+  const productsUtils = new ProductsUtils();
+  const shippingUtils = new ShippingUtils();
+  const ordersUtils = new OrdersUtils();
+  const channelsUtils = new ChannelsUtils();
+  const homePageUtils = new HomePageUtils();
+  const homePageSteps = new HomePageSteps();
+
+  let customerId;
```
mmarkusik, line 25: `let`s complicate understanding the code. Why not use awaits and retrieve values straight from there?

cypress/integration/homePage.js
```diff
@@ -0,0 +1,210 @@
...
+
+    ordersUtils.createReadyToFulfillOrder(
+      customerId,
+      shippingUtils.getShippingMethod().id,
+      defaultChannel.id,
+      productsUtils.getCreatedVariants()
+    );
+
+    cy.get("@salesAmount").then(salesAmount => {
+      let totalAmount = salesAmount + productPrice;
```
mmarkusik, line 178: I think instead of creating a let and changing it a couple of times, what would make this much clearer, would be to remove the let and make even 3 or 4 consts, but properly named so the names would indicate the stages of what's happening in the function. So totalAmount is not exactly totalAmount every time and I have to guess which one is what underneath, since it's an amount (indicating number) at the beginning, but then it's turned into array and then it's string. I think that `totalAmount` (btw `toFixed` returns a string already) -> `[totalAmountIntegral, totalAmountFraction] = totalAmount.toFixed(2).split(".")` -> `totalAmountRegexString` -> `totalAmountRegex`

I strongly recommend avoiding lets in general

cypress/utils/channelsUtils.js
```diff
@@ -0,0 +1,40 @@
+import Channels from "../apiRequests/Channels";
+
+class ChannelsUtils {
+  channels = new Channels();
+
+  deleteChannels(nameStartsWith) {
+    this.channels.getChannels().then(resp => {
+      const channelsArray = new Set(resp.body.data.channels);
+      if (channelsArray) {
```
mmarkusik, line 9: I think an early return could be nice here. It makes the code look a bit more simple in cases such as this, since there's less braces wrapping

```suggestion
      if (!channelsArray) { return; }
      ...rest
```

cypress/integration/homePage.js
```diff
@@ -0,0 +1,210 @@
...
+  before(() => {
+    cy.clearSessionData().loginUserViaRequest();
+    productsUtils.deleteProperProducts(startsWith);
+    customer.deleteCustomers(startsWith);
+    shippingUtils.deleteShipping(startsWith);
+    let addresses;
+
+    channelsUtils
+      .getDefaultChannel()
+      .then(channel => {
```
mmarkusik, line 41: Let's avoid nested `then`s. Async and await should do

cypress/integration/homePage.js
```diff
@@ -0,0 +1,210 @@
...
+      .as("ordersReadyToFulfill");
+
+    ordersUtils.createReadyToFulfillOrder(
+      customerId,
+      shippingUtils.getShippingMethod().id,
+      defaultChannel.id,
+      productsUtils.getCreatedVariants()
+    );
+    cy.get("@ordersReadyToFulfill").then(ordersReadyToFulfill => {
+      const regexp = new RegExp(`^${ordersReadyToFulfill + 1}\D*`);
```
mmarkusik, line 106: I'd like those `regexp`s to be named to indicate what they cover

mmarkusik (changes requested):

cypress/integration/homePage.js
karola312: Cypress doesn't recommended to use async/await commands. These commands don't behave as expected when running code on cypress.

cypress/integration/homePage.js
karola312: Cypress doesn't recommended to use async/await commands. These commands don't behave as expected when running code on cypress.

cypress/integration/homePage.js
mmarkusik: Ok, seen it - makes sense 👍

mmarkusik (approved):
✨

jwm0 (approved):

PR #983, opened 2021-02-18
Implement inclusion feature of postal codes

tomaszszymanski129:
I want to merge this change because it adds inclusion to zip code rules.

API changed - we dropped dedicated mutations for in-place adding / removing postal codes.
Now we use the shippingMethodUpdate which utilizes inclusionType, addPostalCodes and removePostalCodes.
Entire zip*[codeRules] were refactored to postal*[codeRules]
On inclusion change we drop all the rules (not deleted in place but on API call of shippingMethodUpdate)

Testing tips (already checked but worth to be confirmed all works well):
- check shippingRatePrice update page
- check shippingRateWeight update page
- check shippingRatePrice create page
- check shippingRateWeight create page

few crucial AC:
- inclusion change should clear the zip codes list
- inclusion change, adding postal code, removing postal code should unlock the save button on bottom of the update pages listed above
- if i change the inclusion type and add new rules all old rules should be dropped from DB after save

feature/add-inclusion-to-zip-code-rules

### Pull Request Checklist

1. [ ] The changes are tested in different browsers (Chrome, Firefox, Safari).
1. [ ] The changes are tested in light and dark mode.

### Test environment config

API_URI=https://feature-add-inclusion-to-zip-code-rules.api.saleor.rocks/graphql/


src/shipping/handlers.ts
```diff
@@ -45,12 +47,22 @@ export const createChannelsChangeHandler = (
 
 export function getCreateShippingPriceRateVariables(
   data: ShippingZoneRatesPageFormData,
-  id: string
+  id: string,
+  addPostalCodeRules: any,
```
mmarkusik, line 51: Why any?

src/shipping/handlers.ts
```diff
@@ -45,12 +47,22 @@ export const createChannelsChangeHandler = (
...
+  radioInclusionType: PostalCodeRuleInclusionTypeEnum
 ): CreateShippingRateVariables {
   const parsedMinDays = parseInt(data.minDays, 10);
   const parsedMaxDays = parseInt(data.maxDays, 10);
+  const postalCodeRules = addPostalCodeRules
+    .filter(code => !code.id)
```
mmarkusik, line 57: In which case would they not have ids?

src/shipping/handlers.ts
```diff
@@ -45,12 +47,22 @@ export const createChannelsChangeHandler = (
...
+    .map(code => ({
```
mmarkusik, line 58: Why map if it returns the same properties?

src/shipping/handlers.ts
```diff
@@ -45,12 +47,22 @@ export const createChannelsChangeHandler = (
...
 export function getCreateShippingPriceRateVariables(
   data: ShippingZoneRatesPageFormData,
-  id: string
+  id: string,
+  addPostalCodeRules: any,
+  radioInclusionType: PostalCodeRuleInclusionTypeEnum
 ): CreateShippingRateVariables {
   const parsedMinDays = parseInt(data.minDays, 10);
   const parsedMaxDays = parseInt(data.maxDays, 10);
+  const postalCodeRules = addPostalCodeRules
```
mmarkusik, line 56: This is being used multiple times, should be moved to utils function

src/shipping/views/PriceRatesCreate/PriceRatesCreate.tsx
```diff
@@ -123,29 +124,27 @@ export const PriceRatesCreate: React.FC<PriceRatesCreateProps> = ({
...
         openChannelsModal={handleChannelsModalOpen}
         onChannelsChange={setCurrentChannels}
-        onZipCodeAssign={() => openModal("add-range")}
-        onZipCodeUnassign={id =>
-          openModal("remove-range", {
-            id
-          })
-        }
+        onPostalCodeAssign={() => openModal("add-range")}
+        onPostalCodeUnassign={code => {
```
mmarkusik, line 131: Should be moved outside of render

src/shipping/views/PriceRatesUpdate/PriceRatesUpdate.tsx
```diff
@@ -248,6 +238,32 @@ export const PriceRatesUpdate: React.FC<PriceRatesUpdateProps> = ({
     reset();
   };
 
+  let originalCodes = [];
```
mmarkusik, line 241: Why is this not kept with useState?

src/shipping/views/PriceRatesUpdate/PriceRatesUpdate.tsx
```diff
@@ -344,45 +361,32 @@ export const PriceRatesUpdate: React.FC<PriceRatesUpdateProps> = ({
...
+        onPostalCodeAssign={() => openModal("add-range")}
+        onPostalCodeUnassign={code => {
+          if (code.id !== undefined) {
+            setCodesToDelete([...codesToDelete, code.id]);
+            rate.postalCodeRules = rate.postalCodeRules.filter(
+              rule => rule.id !== code.id
+            );
+          } else {
+            rate.postalCodeRules = rate.postalCodeRules.filter(
+              rule => rule.start !== code.start && rule.end !== code.end
```
mmarkusik, line 374: Seen this used a few times in some components, should be a utils function like 
```suggestion
     const getRulesUnmatchingPostalCode = (codeToCompare) => 
        (obj: { start, end }) => obj.start !== codeToCompare.start 
        && rule.end !== code.ToCompareEnd
```

And then do 
```suggestion
          filter(getRulesUnmatchingPostalCode(code))
```

src/shipping/views/PriceRatesUpdate/PriceRatesUpdate.tsx
```diff
@@ -344,45 +361,32 @@ export const PriceRatesUpdate: React.FC<PriceRatesUpdateProps> = ({
...
         }
-        onZipCodeAssign={() => openModal("add-range")}
-        onZipCodeUnassign={id =>
-          openModal("remove-range", {
-            id
-          })
-        }
+        onPostalCodeInclusionChange={onPostalCodeInclusionChange}
+        onPostalCodeAssign={() => openModal("add-range")}
+        onPostalCodeUnassign={code => {
```
mmarkusik, line 366: Please avoid doing more than one liners in return, it gets very crowded

src/shipping/views/PriceRatesUpdate/PriceRatesUpdate.tsx
```diff
@@ -208,11 +171,38 @@ export const PriceRatesUpdate: React.FC<PriceRatesUpdateProps> = ({
 
   const [updateMetadata] = useMetadataUpdate({});
   const [updatePrivateMetadata] = usePrivateMetadataUpdate({});
+  let radioInclusionType;
```
mmarkusik, line 174: Why use lets here? If this prop changes shouldn't it be useState?

src/shipping/views/WeightRatesCreate/WeightRatesCreate.tsx
```diff
@@ -126,29 +130,27 @@ export const WeightRatesCreate: React.FC<WeightRatesCreateProps> = ({
...
         openChannelsModal={handleChannelsModalOpen}
         onChannelsChange={setCurrentChannels}
-        onZipCodeAssign={() => openModal("add-range")}
-        onZipCodeUnassign={id =>
-          openModal("remove-range", {
-            id
-          })
-        }
+        onPostalCodeAssign={() => openModal("add-range")}
+        onPostalCodeUnassign={code => {
```
mmarkusik, line 137: Move outside of render

src/shipping/views/WeightRatesUpdate/WeightRatesUpdate.tsx
```diff
@@ -84,6 +84,50 @@ export const WeightRatesUpdate: React.FC<WeightRatesUpdateProps> = ({
...
+    setCodesToDelete(
+      rate.postalCodeRules
+        .filter(code => code.id !== undefined)
+        .map(code => code.id)
+    );
+    setHavePostalCodesChanged(true);
+    rate.postalCodeRules = [];
+  };
+
+  const onPostalCodeAssign = (rule: any) => {
```
mmarkusik, line 107: Same as above

src/shipping/views/WeightRatesUpdate/WeightRatesUpdate.tsx
```diff
@@ -346,45 +362,32 @@ export const WeightRatesUpdate: React.FC<WeightRatesUpdateProps> = ({
...
         }
-        onZipCodeAssign={() => openModal("add-range")}
-        onZipCodeUnassign={id =>
-          openModal("remove-range", {
-            id
-          })
-        }
+        onPostalCodeInclusionChange={onPostalCodeInclusionChange}
+        onPostalCodeAssign={() => openModal("add-range")}
+        onPostalCodeUnassign={code => {
```
mmarkusik, line 367: Move outside of return

mmarkusik (dismissed):

src/shipping/handlers.ts
tomaszszymanski129: when codes are newly added and not yet saved. Array which leads here contains new entries added by user and existing elements from DB (with ids). The mutation should utilize `addPostalCodeRules` only with ones with no id, if we pass it all it raises an exception.

src/shipping/handlers.ts
tomaszszymanski129: Rush. Changed everywhere to use proper typing.

src/shipping/components/ShippingZoneRatesPage/ShippingZoneRatesPage.tsx
```diff
@@ -102,6 +108,15 @@ export const ShippingZoneRatesPage: React.FC<ShippingZoneRatesPageProps> = ({
     type: rate?.type || null
   };
 
+  let postalCodesInclusionType = rate?.postalCodeRules[0]?.inclusionType;
+
+  const postalCodeInclusionChange = (
+    inclusion: PostalCodeRuleInclusionTypeEnum
+  ) => {
+    postalCodesInclusionType = inclusion;
```
jwm0, line 116: This reassignment has no effect as it will always get called after the initial render and this variable value will always equal to:
`rate?.postalCodeRules[0]?.inclusionType;`

src/shipping/handlers.ts
```diff
@@ -43,14 +46,32 @@ export const createChannelsChangeHandler = (
   triggerChange();
 };
 
+const getaddPostalCodeRules = (
```
jwm0, line 49: This function name isn't clear on what it does, perhaps `getAddedPostalCodeRules` makes more sense?

src/shipping/handlers.ts
```diff
@@ -43,14 +46,32 @@ export const createChannelsChangeHandler = (
...
+  rules: ShippingMethodFragment_postalCodeRules[]
+) =>
+  rules
+    .filter(code => !code.id)
+    .map(
+      code =>
+        ({
+          end: code.end,
+          start: code.start
+        } as ShippingPostalCodeRulesCreateInputRange)
```
jwm0, line 59: By looking at `ShippingMethodFragment_postalCodeRules` type the `code.start` is an optional field whereas `ShippingPostalCodeRulesCreateInputRange` has a required `start` type of `String`. This is either not typed properly in the graphql schema or you should handle this case appropriately by using some kind of a fallback in case `code.start` is a null.

src/shipping/views/PriceRatesUpdate/PriceRatesUpdate.tsx
```diff
@@ -209,10 +174,40 @@ export const PriceRatesUpdate: React.FC<PriceRatesUpdateProps> = ({
...
+  const onPostalCodeInclusionChange = (
+    inclusion: PostalCodeRuleInclusionTypeEnum
+  ) => {
+    setInclusionType(inclusion);
+    setCodesToDelete(
+      rate.postalCodeRules
+        .filter(code => code.id !== undefined)
+        .map(code => code.id)
+    );
+    setHavePostalCodesChanged(true);
```
jwm0, line 195: Lots of these state values seem to be codependent on each other, this is a perfect use case for the `React.useReducer` hook

src/shipping/views/PriceRatesUpdate/PriceRatesUpdate.tsx
```diff
@@ -209,10 +174,40 @@ export const PriceRatesUpdate: React.FC<PriceRatesUpdateProps> = ({
...
+    rate.postalCodeRules = [];
```
jwm0, line 196: Mutating a variable inside a render cycle has no effect, either remove this or move to a React state.

src/shipping/views/PriceRatesUpdate/PriceRatesUpdate.tsx
```diff
@@ -248,6 +243,42 @@ export const PriceRatesUpdate: React.FC<PriceRatesUpdateProps> = ({
...
+  };
+
+  const onPostalCodeUnassign = code => {
+    if (code.id !== undefined) {
+      setCodesToDelete([...codesToDelete, code.id]);
+      rate.postalCodeRules = rate.postalCodeRules.filter(
+        rule => rule.id !== code.id
+      );
+    } else {
+      rate.postalCodeRules = filterPostalCodes(rate.postalCodeRules, code);
```
jwm0, line 276: Same here, you're treating `rate` as a state but modifying it has no effect. The only reason you see an update in the UI is because you're calling setState functions below and triggering a rerender with a new value of `rate` defined in line 94

src/shipping/components/ShippingZonePostalCodes/ShippingZonePostalCodes.tsx
```diff
@@ -60,23 +57,44 @@ const useStyles = makeStyles(
...
+    initialInclusionType
+  );
   const intl = useIntl();
   const classes = useStyles({});
 
+  const onInclusionRadioChange = (event: React.ChangeEvent<any>) => {
+    const value = event.target.value;
+    setInclusionType(value);
+    onPostalCodeInclusionChange(value);
+  };
```
jwm0, line 84: It looks like you're keeping the exact same state twice. What happens if `initialInclusionType` prop changes and your state is out-of-sync? Parent's `initialInclusionType` state could be modified by another component/query outside this component. Consider if this is ever a case, perhaps getting rid off `inclusionType` internal state here would solve this issue. You could then change the prop name `initialInclusionType` => `inclusionType` and lift the state out of this component

jwm0 (changes requested):

jwm0 (approved):
Some issues raised in the review will be fixed in the following PR

maarcingebala (approved):

PR #1115, opened 2021-05-14
Use available channels from context

jwm0:
I want to merge this change because it reduces the number of api calls and instead uses available channels from global state. Additionally this fixes the issue where users without MANAGE_CHANNELS permission cannot fetch basic channels metadata. An example would be a user with MANAGE_PRODUCTS who cannot see to which channels a product is available in.

**QA instructions**
The following views have changed and should work the same as our current staging:
1. Collection Create
2. CollectionUpdate
3. Sales Create
4. Sales Update
5. Voucher Create
6. Voucher Update
7. Product Create
8. Product Update
9. Shipping Zone Details

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/channels/ChannelsWithVariantsAvailabilityCard/ChannelsWithVariantsAvailabilityCard.tsx
+++ b/src/channels/ChannelsWithVariantsAvailabilityCard/ChannelsWithVariantsAvailabilityCard.tsx
@@ -18,6 +18,7 @@ import {
   areAnyChannelVariantsSelected,
   getTotalSelectedChannelsCount
 } from "@saleor/products/views/ProductUpdate/utils";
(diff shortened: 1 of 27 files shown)


orzechdev (approved):

dominik-zeglen (approved):

dominik-zeglen:
Just a quick note, I think `availableChannels` field can be dropped from `AppChannelContext` since it's just memozing apollo output

jwm0:
> Just a quick note, I think `availableChannels` field can be dropped from `AppChannelContext` since it's just memozing apollo output

Having them defined in context does us no harm and I believe it makes sense to include them together with the rest

eCzerniak (approved):

PR #1122, opened 2021-05-18
Saleor 3391 tests for staff members

karola312:
I want to merge this change because I created tests for staff users

### Pull Request Checklist

### Test environment config


patrys:
Task linked: [SALEOR-3391 Create tests for staff members](https://app.clickup.com/t/2549495/SALEOR-3391)

eCzerniak (approved):

cypress/utils/users.js
```diff
@@ -0,0 +1,40 @@
...
+      firstName,
+      lastName,
+      email,
+      isActive,
+      permissionId: permissions[0].node.id
+    });
+  });
+}
+
+export function getMailActivationLinkForUser(email, i = 0) {
```
jwm0, line 21: This function is very cryptic, why is it recursive? If it's supposed to check few times before failing there should be some timeout, otherwise it's synchronously calling `cy.mhGetMailsByRecipient` few times immediately.

karola312: There is timeout. cy.mhGetMailsByRecipient(email).should - should by default has timeout

mmarkusik: I agree with @jwm0, but if this function needs to be recursive, then can we add a comment to explain this is purposeful?

karola312: Comment like this will be transparent? "Function mhGetMailsByRecipient first get all emails from mailhog with a timout, and after that it finds email from recipient. It cloud happened that invite email from saleor will not be recived yet, so in this case I want to take again all of the emails and check if there is an email to the correct recipient"

mmarkusik (approved):

jwm0 (approved):

PR #1155, opened 2021-06-11
SALEOR-3521 - replaced skeleton placeholder in warehouse list view

SektorDV:
I want to merge this change because the skeleton placeholder for missing shipping zones on warehouse list has been replaced.

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/warehouses/components/WarehouseList/WarehouseList.tsx
+++ b/src/warehouses/components/WarehouseList/WarehouseList.tsx
@@ -132,7 +132,7 @@ const WarehouseList: React.FC<WarehouseListProps> = props => {
               className={classes.tableRow}
               hover={!!warehouse}
               onClick={warehouse ? onRowClick(warehouse.id) : undefined}
-              key={warehouse ? warehouse.id : "skeleton"}
+              key={warehouse ? warehouse.id : "n/a"}
               data-test="warehouseEntry"
               data-testid={warehouse?.name.toLowerCase().replace(" ", "")}
             >
@@ -142,7 +142,7 @@ const WarehouseList: React.FC<WarehouseListProps> = props => {
               <TableCell className={classes.colZones} data-test="zones">
                 {mapEdgesToItems(warehouse?.shippingZones)
                   .map(({ name }) => name)
-                  .join(", ") || <Skeleton />}
+                  .join(", ") || "-"}
               </TableCell>
               <TableCell className={classes.colActions}>
                 <div className={classes.actions}>
--- a/src/warehouses/components/WarehouseList/WarehouseList.tsx
+++ b/src/warehouses/components/WarehouseList/WarehouseList.tsx
@@ -132,7 +132,7 @@ const WarehouseList: React.FC<WarehouseListProps> = props => {
               className={classes.tableRow}
               hover={!!warehouse}
               onClick={warehouse ? onRowClick(warehouse.id) : undefined}
-              key={warehouse ? warehouse.id : "n/a"}
+              key={warehouse ? warehouse.id : "skeleton"}
               data-test="warehouseEntry"
               data-testid={warehouse?.name.toLowerCase().replace(" ", "")}
             >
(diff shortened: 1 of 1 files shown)


patrys:
Task linked: [SALEOR-3521 Add something instead of skeleton to warehouse list when no shipping zones added to a warehouse](https://app.clickup.com/t/2549495/SALEOR-3521)

orzechdev (changes requested):
Please also update test snapshots with `npm run test -- -u`

src/warehouses/components/WarehouseList/WarehouseList.tsx
```diff
@@ -132,7 +132,7 @@ const WarehouseList: React.FC<WarehouseListProps> = props => {
               className={classes.tableRow}
               hover={!!warehouse}
               onClick={warehouse ? onRowClick(warehouse.id) : undefined}
-              key={warehouse ? warehouse.id : "skeleton"}
+              key={warehouse ? warehouse.id : "n/a"}
```
dominik-zeglen, line 135: Unnecessary

src/warehouses/components/WarehouseList/WarehouseList.tsx
```diff
@@ -142,7 +142,7 @@ const WarehouseList: React.FC<WarehouseListProps> = props => {
               <TableCell className={classes.colZones} data-test="zones">
                 {mapEdgesToItems(warehouse?.shippingZones)
                   .map(({ name }) => name)
-                  .join(", ") || <Skeleton />}
+                  .join(", ") || "-"}
```
dominik-zeglen, line 145: Skeleton should remain visible when page is loading and data is `undefined`, dash symbol should be displayed only if `mapEdtesToItems` returns `[]`.

dominik-zeglen (changes requested):

src/warehouses/components/WarehouseList/WarehouseList.tsx
SektorDV: I'm not sure if that's applicable though? When the data is still loading, the entire list and all its fields are unpopulated:
image

src/warehouses/components/WarehouseList/WarehouseList.tsx
```diff
@@ -140,9 +140,13 @@ const WarehouseList: React.FC<WarehouseListProps> = props => {
...
-                {mapEdgesToItems(warehouse?.shippingZones)
-                  .map(({ name }) => name)
-                  .join(", ") || <Skeleton />}
+                {disabled ? (
+                  <Skeleton />
+                ) : (
+                  mapEdgesToItems(warehouse?.shippingZones)
+                    ?.map(({ name }) => name)
+                    .join(", ") || "-"
+                )}
```
dominik-zeglen, line 149: Why using disabled here?

```suggestion
                {warehouse?.shippingZones === undefined ? (
                  <Skeleton />
                ) : (
                  mapEdgesToItems(warehouse?.shippingZones).map(({ name }) => name)
                    .join(", ") || "-"
                )}
```

It would be even nicer to extract `mapEdgesToItems(warehouse?.shippingZones)` to separate variable

SektorDV: i used `disabled` because it's assigned the value of `loading` a few levels up, and i need it to know whether the item has finished loading. i wanted to avoid introducing another prop which would end up having the same value as `disabled`. seeing as how there seems to be an ongoing discussion on the topic elsewhere, i might end up doing it anyway.

dominik-zeglen: I'm strongly against using `disabled` as a loading state only because it's dependent on `loading` a few levels higher. `warehouse?.shippingZones === undefined` is the only case when skeleton should be displayed.

dominik-zeglen (changes requested):

dominik-zeglen (approved):

orzechdev (approved):

PR #1172, opened 2021-06-17
Fix app configuration frame embedding

dominik-zeglen:
I want to merge this change because it fixes problems with iframe scripts embedding - it currently handles cases when configuration page has no `<script />` tags at all or when they don't have `src` property. Also fixes relative path handling.

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/apps/components/AppDetailsSettingsPage/useAppConfigLoader.ts
+++ b/src/apps/components/AppDetailsSettingsPage/useAppConfigLoader.ts
@@ -1,5 +1,5 @@
 import { AppFragment } from "@saleor/fragments/types/AppFragment";
-import { MutableRefObject, useEffect, useRef } from "react";
+import { useEffect, useRef } from "react";
 import urlJoin from "url-join";
 
 export type UseAppConfigLoaderCallbacks = Record<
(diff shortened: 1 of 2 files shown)


src/apps/components/AppDetailsSettingsPage/useAppConfigLoader.ts
```diff
@@ -0,0 +1,75 @@
...
+  return (node: HTMLScriptElement) => {
+    // Using node.getAttribute beacuse node.src returns absolute path
+    const src = node.getAttribute("src");
+    if (src?.startsWith("/")) {
+      node.src = urlJoin(origin, src);
+    }
+  };
+}
+
+async function _fetchAndSetContent(
```
jwm0, line 20: I don't think we need to prefix this with `_` as we're not exporting it anywhere

src/apps/components/AppDetailsSettingsPage/useAppConfigLoader.ts
```diff
@@ -0,0 +1,75 @@
...
+  frameContainer: MutableRefObject<HTMLDivElement>,
```
jwm0, line 21: Let's pass the object reference itself without any React logic
```suggestion
  frameContainer: HTMLDivElement,
```

src/apps/components/AppDetailsSettingsPage/useAppConfigLoader.ts
```diff
@@ -0,0 +1,75 @@
...
+
+function useAppConfigLoader(
+  data: AppFragment,
+  backendHost: string,
+  callbacks: UseAppConfigLoaderCallbacks
+) {
+  const frameContainer = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    _fetchAndSetContent(frameContainer, data, backendHost, callbacks);
```
jwm0, line 69:
```suggestion
    _fetchAndSetContent(frameContainer.current, data, backendHost, callbacks);
```

jwm0 (changes requested):

src/apps/components/AppDetailsSettingsPage/useAppConfigLoader.ts
dominik-zeglen: Im not so sure if it would work, this object has to be mutable. Ill test it tomorrow

mmarkusik (approved):

src/apps/components/AppDetailsSettingsPage/useAppConfigLoader.ts
dominik-zeglen: Suprisingly it works

jwm0 (approved):

SektorDV (approved):

kamilpastuszka (approved):

PR #1180, opened 2021-06-20
[SALEOR-3088] Date & DateTime attributes

piotrgrundas:
I want to merge this change because:
- it adds date & date-time attributes handling,
- handles boolean attributes in variant creator,

[Saleor PR](https://github.com/mirumee/saleor/pull/7500)
**PR intended to be tested with API branch: `SALEOR-3088/datetime_attributes`** 

### Screenshots
image
image
image
image
image

### Pull Request Checklist

### Test environment config

API_URI=https://saleor-3088-datetime-attributes.api.saleor.rocks/graphql/


patrys:
Task linked: [SALEOR-3088 Datetime & date attributes](https://app.clickup.com/t/2549495/SALEOR-3088)

piotrgrundas:
I'll update snapshots & translations later. I want to start gathering review comments ( ͡° ͜ʖ ͡°)

pwgryglak (approved):
For now it looks good, but during redesign we need to adjust to fit designs - chips for example

src/products/components/ProductListPage/filters.ts
```diff
@@ -76,15 +78,26 @@ const messages = defineMessages({
   }
 });
 
+const extractAttributes = (opts: ProductListFilterOpts) => (
```
dominik-zeglen, line 81: Most if not whole project uses maps and filters. What do you think that instead of currying we'd implement something like this?

```JSX
function filterByType(type: AttributeInputTypeEnum) {
  return (attribute: attributetype) => attribute.inputType === type
}

const attributes = opts.attributes;
const booleanAttributes = attributes.filter(filterByType(AttributeInputTypeEnum.BOOLEAN))
...
```

src/products/components/ProductVariantCreatorPage/ProductVariantCreatorValues.tsx
```diff
@@ -78,15 +114,30 @@ const ProductVariantCreatorValues: React.FC<ProductVariantCreatorValuesProps> =
...
+              boolean: attributeValue,
+              name: `${attributeName}: ${attributeValue ? "Yes" : "No"}`
+            }
+          }
+        : {
+            slug: attributeValue,
+            value:
+              dataAttribute?.values.find(getBySlug(attributeValue))?.value ||
+              attributeValues.find(getBySlug(attributeValue))
+          }
```
dominik-zeglen, line 139: How about making it a util function?

src/products/views/ProductList/filters.ts
```diff
@@ -207,25 +208,78 @@ export function getFilterOpts(
...
 function getFilteredAttributeValue(
   params: ProductListUrlFilters
-): Array<({ boolean: boolean } | { values: string[] }) & { slug: string }> {
+): Array<
+  (
+    | { boolean: boolean }
+    | { values: string[] }
+    | { dateTime: GteLte<string> }
+    | { date: GteLte<string> }
+  ) & { slug: string }
```
dominik-zeglen, line 251: Please extract it to named type

dominik-zeglen (changes requested):

SektorDV (commented):
There seems to be an overreliance across the entire PR on `any` types, but I'm not sure if those files are autogenerated or coded manually. In either case, it would perhaps be worthwhile to add specificity to those types?

`  date: any | null;
  dateTime: any | null;`

patrys:
I've posted it in another review: explicit `any` is an antipattern and should be forbidden. We should use `unknown` where we don't care about the return type of a callback or if we only pass the data along.

schema.graphql
```diff
@@ -697,7 +705,8 @@ type AttributeValueUpdate {
 
 input BulkAttributeValueInput {
   id: ID
-  values: [String!]!
+  values: [String]
```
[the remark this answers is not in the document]
piotrgrundas: Absolutely right.

piotrgrundas:
It's autogenerated via codegen.

src/products/components/ProductListPage/filters.ts
piotrgrundas: Ur playground so what ever is more convenient for u 👌

src/products/components/ProductVariantCreatorPage/utils.ts
```diff
@@ -44,3 +47,33 @@ export function getStockAttributeValues(
...
+  attributeValue: boolean
+): AttributeValue<Partial<AttributeValueFragment>> => ({
+  slug: attributeValue.toString(),
+  value: {
+    boolean: attributeValue,
+    name: `${attributeName}: ${attributeValue ? "Yes" : "No"}`
+  }
+});
+
+export const geBasicAttributeValue = (
```
dominik-zeglen, line 65:
```suggestion
export const getBasicAttributeValue = (
```

dominik-zeglen (approved):

SektorDV (approved):

eCzerniak (changes requested):
1. [ ] If an attr is not required then after saving new product these attributes disappear
2. [ ] It is not possible to save without an error new Page with attributes set as vale not required. 
3. [x] filtering by dateTime type attributes it is not possible. 
4. [ ] It is not possible to export products. 
5. [ ] Payload in webhooks is missing: boolean, date_time & unit. 
6. [ ] (Need to confirm) at webhook in attributes.values.name - there is a name of an attribute and the value of it.

pwgryglak (approved):
UI wise looks ok 
filtering doesn't work

eCzerniak (approved):

PR #1187, opened 2021-06-22
Adjust list views

jwm0:
I want to merge this change because it adjusts Products, Sales and Vouchers lists to use channel from filters. Full definition of done (DoD) can be found in the task below:
https://app.clickup.com/t/2549495/SALEOR-3416

Additionally, filter error handling has been refactored to be more generic. It introduces basic validation along with the ability to specify filter dependencies - a list of other filters the active filter depends on.

**PR intended to be tested with API branch:** https://github.com/mirumee/saleor/pull/7374

### Screenshots
image
image
image

### Pull Request Checklist

### Test environment config

API_URI=https://saleor-2979-use-root-level-channel-argument-for-filtering-and-s.api.saleor.rocks/graphql/


patrys:
Task linked: [SALEOR-3416 Adjust product list view](https://app.clickup.com/t/2549495/SALEOR-3416)

src/utils/maps.ts
```diff
@@ -18,40 +20,45 @@ export function mapEdgesToItems<T>(data?: EdgesType<T>): T[] {
...
   }
 
   return nodes.map(node => ({
     label: node.name,
-    value: node.id
+    value: getterFn ? getterFn(node) : node.id
   }));
 }
 
 export function mapSlugNodeToChoice(
```
mmarkusik, line 59: I think since this can be now well handled by mapNodeToChoice which is used in this pr exactly for slug item, in order to keep it consistent we should either: 
- also remove `mapSlugNodeToChoice` and use mapNodeToChoice instead
- make mapSlugNodeToChoice use mapNodeToChoice underneath `(nodes) => mapNodeToChoice(nodes, node => node.slug)` and use that instead of manually doing it each time

Imo since the slug nodes are second most commonly mapped to choices in the app (right after items that use `id`) it'd be nice to keep it separate, not to rewrite with each usage `mapNodeToChoice(availableChannels, channel => channel.slug)`. And I'd keep manually adding the getter function when items we're mapping are singularly used. Wdyt?

mmarkusik (changes requested):

src/utils/maps.ts
jwm0: Second options looks good, I wasn't aware there was a dedicated fn for that. Nice catch

SektorDV (approved):

eCzerniak (approved):

mmarkusik (approved):

pwgryglak (approved):
Looks good

PR #1188, opened 2021-06-22
fix multiselect dropdowns in variant creator

SektorDV:
I want to merge this change because it fixes the issues present when there's more than one multiselect on the variant creator page.

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/products/components/ProductVariantCreatorPage/ProductVariantCreatorValues.tsx
+++ b/src/products/components/ProductVariantCreatorPage/ProductVariantCreatorValues.tsx
@@ -10,6 +10,8 @@ import { getById } from "@saleor/orders/components/OrderReturnPage/utils";
 import { ProductDetails_product_productType_variantAttributes } from "@saleor/products/types/ProductDetails";
 import { SearchAttributeValues_attribute_choices_edges_node } from "@saleor/searches/types/SearchAttributeValues";
 import { FetchMoreProps } from "@saleor/types";
+// eslint-disable-next-line no-restricted-imports
+import _ from "lodash";
 import React from "react";
 import { defineMessages, FormattedMessage, useIntl } from "react-intl";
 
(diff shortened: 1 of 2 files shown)


patrys:
Task linked: [SALEOR-3617 Dropdowns bug](https://app.clickup.com/t/2549495/SALEOR-3617)

src/products/components/ProductVariantCreatorPage/ProductVariantCreatorValues.tsx
```diff
@@ -10,6 +10,8 @@ import { getById } from "@saleor/orders/components/OrderReturnPage/utils";
 import { ProductDetails_product_productType_variantAttributes } from "@saleor/products/types/ProductDetails";
 import { SearchAttributeValues_attribute_choices_edges_node } from "@saleor/searches/types/SearchAttributeValues";
 import { FetchMoreProps } from "@saleor/types";
+// eslint-disable-next-line no-restricted-imports
+import _ from "lodash";
```
dominik-zeglen, line 14:
```suggestion
import isEqual from "lodash/isEqual";
```

dominik-zeglen (changes requested):
I don't think it's the right approach - passing attribute values if the attribute id matches the one used by query and empty array if not makes more sense (and it's easier to understand what is actually happening in there) than using memoization with deep compare + onBlur handler

jwm0:
> I don't think it's the right approach - passing attribute values if the attribute id matches the one used by query and empty array if not makes more sense (and it's easier to understand what is actually happening in there) than using memoization with deep compare + onBlur handler

In general I believe the issue is not with the component itself but rather the way we're it. Single root level state for multiple selects doesn't work. Having it as a single state is fine but it should be persisted inside some key to maintain the cache and have a clear seperation. (using the select `name` as a key should be possible)

jwm0 (changes requested):
See my above comment

PR #1212, opened 2021-07-06
change default port to 9001

tungnt3-smart:
As MacOS user, my default port for php-fpm is 9000
I would like to change default port to 9001 for better development experience


chatchaiatscg:
Can you change php-fpm  to another port, if we change then every time we do mainstream sync then it will conflict.

tungnt3-smart:
> Can you change php-fpm to another port, if we change then every time we do mainstream sync then it will conflict.

I see, I will try to resolve that issue on my local. I will close thif PR as well

PR #1263, opened 2021-07-27
fix crash when deleting more than 21 vouchers

SektorDV:
I want to merge this change because it resolves the issue of the dashboard crashing when more than 21 vouchers are selected for deletion.

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/discounts/index.tsx
+++ b/src/discounts/index.tsx
@@ -63,7 +63,7 @@ const SaleCreateView: React.FC<RouteComponentProps> = ({ location }) => {
 };
 
 const VoucherListView: React.FC<RouteComponentProps<{}>> = ({ location }) => {
-  const qs = parseQs(location.search.substr(1));
+  const qs = parseQs(location.search.substr(1), { arrayLimit: Infinity });
   const params: VoucherListUrlQueryParams = asSortParams(
     qs,
     VoucherListUrlSortField,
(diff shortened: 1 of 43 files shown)


patrys:
Task linked: [SALEOR-3967 Dashboard error on removing multiple vouchers](https://app.clickup.com/t/2549495/SALEOR-3967)

orzechdev (approved):
Maybe it's also worth to see if there are similar places in dashboard where same situation may occur, besides vouchers?

src/discounts/index.tsx
```diff
@@ -63,7 +63,7 @@ const SaleCreateView: React.FC<RouteComponentProps> = ({ location }) => {
 };
 
 const VoucherListView: React.FC<RouteComponentProps<{}>> = ({ location }) => {
-  const qs = parseQs(location.search.substr(1));
+  const qs = parseQs(location.search.substr(1), { arrayLimit: Infinity });
```
jwm0, line 66: Document why you need to do it this way, it looks confusing. Doesn't it also touch other parts of the app?

jwm0 (changes requested):

SektorDV:
turns out the problem was related to the way the application creates URL query strings (a number of places in the app used stringifyQs imported directly from the qs library rather than the one in utils). as a result, i removed the previous fix and globally replaced the stringifyQs imports across the app, which resolves the issue with vouchers and hopefully prevents similar issues from arising anywhere else. i checked the app for possible issues arising from the changes, but a doublecheck by QA probably couldn't hurt.

jwm0 (approved):
@SektorDV Please mention all the views that this change touches so QA can check each one of them

PR #1271, opened 2021-08-03
Add gift card creation

mmarkusik:
I want to merge this change because it adds creating single gift card

### Pull Request Checklist

### Test environment config


patrys:
Task linked: [SALEOR-3910 Add single Gift Card creation](https://app.clickup.com/t/2549495/SALEOR-3910)

src/components/ContentOrSkeleton.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+interface ContentOrSkeletonProps {
+  condition: any;
+  children: React.ReactNode;
+}
+
+const ContentOrSkeleton: React.FC<ContentOrSkeletonProps> = ({
+  condition,
+  children
+}) => (!!condition ? <>{children}</> : <Skeleton />);
```
orzechdev, line 12: I don't like this component. It's good to hide conditional displaying of skeleton, but I would rather think that clearly the main purpose of this component should be to display content, while showing skeleton should be temporary state. Currently the name `ContentOrSkeleton` doesn't suggest it, I would rather see e.g. `WithSkeleton`. Also `condition` prop is not precise, as we don't know clearly what occurs when it is true or false, without seeing implementation details. An interface should clearly define what it expects, e.g. `showSkeleton`.

But going even further, what do you think about even refactoring current `Skeleton` component in `@saleor/components/Skeleton` to be able to accept `show` prop and children, and then use it like:
```tsx
<Skeleton show={!loading}>
  <SomeChildren />
</Skeleton>
```
or use it like it is now when `children` are absent and `show` defaults to true:
```tsx
<Skeleton />
```
?

orzechdev: Hiding conditional displaying of the skeleton is a very good idea 👍 .

orzechdev (changes requested):

src/components/ContentOrSkeleton.tsx
mmarkusik: Sounds good to me, @jwm0 does adding this ability to exsiting skeleton sounds ok to you?

src/giftCards/GiftCardsList/GiftCardsListTable/GiftCardsListTable.tsx
```diff
@@ -7,50 +7,48 @@ import {
...
-    // TEMP
-    // disabled
-  };
+  const {
+    toggle,
+    isSelected,
+    giftCards,
+    numberOfColumns,
+    loading
+  } = useContext(GiftCardsListContext);
```
jwm0, line 42: Didn't we want to export a hook: `useGiftCardBulkActions` that hooks into this context?

src/giftCards/GiftCardsList/GiftCardsListTable/GiftCardsListTableHeader.tsx
```diff
@@ -9,25 +9,25 @@ import Label, {
...
-  const { toggleAll, listElements, giftCards } = useContext(
-    GiftCardsListContext
-  );
+  const {
+    toggleAll,
+    listElements,
+    giftCards,
+    numberOfColumns,
+    loading
+  } = useContext(GiftCardsListContext);
```
jwm0, line 30: same as above

src/giftCards/components/GiftCardExpirySelect/GiftCardExpirySelect.tsx
```diff
@@ -0,0 +1,110 @@
...
+  {
+    label: messages.expirationDateLabel,
+    value: GiftCardExpiryTypeEnum.EXPIRY_DATE
+  }
+];
+
+interface GiftCardExpirySelectProps {
+  change: FormChange;
+  expiryPeriodType: TimePeriodTypeEnum;
+  expiryPeriodAmount: string;
```
jwm0, line 41: Should it be a string? Perhaps we should cast it later to a string but it makes more sense to have it as a number to me?

src/giftCards/GiftCardUpdatePage/messages.ts
```diff
@@ -1,12 +1,34 @@
...
   }
 });
+
+export function getGiftCardErrorMessage(
+  error: Omit<GiftCardUpdate_giftCardUpdate_errors, "__typename"> | undefined,
+  intl: IntlShape
+): string {
+  if (error) {
+    switch (error.code) {
+      case GiftCardErrorCode.GRAPHQL_ERROR:
```
jwm0, line 21: This doesn't seem like a good error for input level field 🤔

jwm0 (changes requested):

src/components/TextWithSelectField/TextWithSelectField.tsx
```diff
@@ -3,47 +3,100 @@ import SingleSelectField, {
...
+  const {
+    name: selectFieldName,
+    value: selectFieldValue,
+    className: selectFieldClassName
+  } = selectFieldProps;
+
+  // in case one of the fields in the form is empty
+  // we need to save the other part of the field as well
+  const handleChange = (type: "selectField" | "textField") => (
+    event: React.ChangeEvent<any>
```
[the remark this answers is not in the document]
mmarkusik: Actually I think our own `ChangeEvent` is even better

src/giftCards/GiftCardUpdatePage/messages.ts
mmarkusik: We use this logic in multiple places, maybe we should remove it from all of them then? 🤔

src/giftCards/components/GiftCardExpirySelect/GiftCardExpirySelect.tsx
jwm0: Is it also a String in schema? If not we should treat it as number everywhere and only cast it to String in the input itself

src/giftCards/GiftCardUpdatePage/messages.ts
jwm0: Create a ticket for this in 3.1/maintenance list and leave it as is right now

src/giftCards/components/GiftCardExpirySelect/GiftCardExpirySelect.tsx
mmarkusik: Not, then adding this as a number and parsing to the text with select field 👍

jwm0 (approved):

PR #1300, opened 2021-08-17
Add gift card settings view

orzechdev:
I want to merge this change because... it adds gift card settings view + updates settings header (gift cards, orders and site settings) with underline.

**PR intended to be tested with API branch:**  feature/gift-cards-mvp

### Screenshots

Zrzut ekranu 2021-08-17 o 18 42 26

### Pull Request Checklist

### Test environment config

API_URI=https://feature-gift-cards-mvp.api.saleor.rocks/graphql/


patrys:
Task linked: [SALEOR-3908 Add Gift Card settings view](https://app.clickup.com/t/2549495/SALEOR-3908)

src/giftCards/GiftCardSettings/GiftCardSettingsPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { useGiftCardSettingsQuery } from "./queries";
+import { GiftCardSettingsFormData } from "./types";
+import { getGiftCardSettingsInputData } from "./utils";
+
+const GiftCardSettingsPage: React.FC = () => {
+  const intl = useIntl();
+
+  const navigate = useNavigator();
+
+  const { data, loading } = useGiftCardSettingsQuery({});
```
jwm0, line 32: We should allow for empty function call if no variables/opts are required, possibly by making the default value `{}`, food for thought.

src/giftCards/GiftCardSettings/messages.ts
```diff
@@ -0,0 +1,33 @@
...
+      case GiftCardSettingsErrorCode.INVALID:
+        return intl.formatMessage(commonErrorMessages.invalid);
+
+      default:
+        return intl.formatMessage(commonErrorMessages.unknownError);
+    }
+  }
+
+  return undefined;
+}
```
jwm0, line 33: I feel like we're redefining this function many times with the exact same messages defined. I'd suggest to do a an error message extractor for all the generic ones but leave the specific functions in order to extend errors when needed:
```suggestion
export function getGiftCardSettingsErrorMessage(
  error: GiftCardSettingsErrorFragment | undefined,
  intl: IntlShape
): string {
  if (error.code) {
    // and here we could extend the messages to giftcard specific when/if needed
    return getGenericFormFieldErrorMessage(error.code);
  }
  
  return undefined;
}
```

jwm0 (changes requested):

src/giftCards/GiftCardSettings/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+import { GiftCardSettingsFormData } from "./types";
+
+export const getGiftCardSettingsInputData = ({
+  expiryPeriodActive,
+  expiryPeriodType,
+  expiryPeriodAmount
+}: Pick<
+  GiftCardSettingsFormData,
+  "expiryPeriodActive" | "expiryPeriodType" | "expiryPeriodAmount"
+>): GiftCardSettingsUpdateInput => ({
```
mmarkusik, line 15: I think this could look a bit cleaner
```suggestion
>): GiftCardSettingsUpdateInput => {
  if (expiryPeriodActive) {
    return {
      expiryType: GiftCardSettingsExpiryTypeEnum.EXPIRY_PERIOD,
      expiryPeriod: {
        type: expiryPeriodType,
        amount: expiryPeriodAmount
      }
    };
  }

  return {
    expiryType: GiftCardSettingsExpiryTypeEnum.NEVER_EXPIRE
  };
};
```

Especially since we always keep a default value for the select in the form initial data, right? So the case when we chose expiry period, and have no amount (like the check in line 20 here) will never happen, therefore I don't think we need that check

mmarkusik (changes requested):

src/giftCards/GiftCardSettings/utils.ts
orzechdev: No, there is a possibility when user may wipe out amount input and then using backend side validation it returns error.

src/giftCards/GiftCardSettings/messages.ts
orzechdev: The types are different, here `GiftCardSettingsErrorCode` but in other place `GiftCardErrorCode`, this is why function with switch case is redefined.

orzechdev: I've added `CommonErrorCode` enum to represent common schema enums members.
```typescript
enum CommonErrorCode {
  GRAPHQL_ERROR = "GRAPHQL_ERROR",
  INVALID = "INVALID",
  REQUIRED = "REQUIRED"
}
interface CommonError<ErrorCode> {
  code: ErrorCode | CommonErrorCode;
  field: string | null;
}

export function getCommonFormFieldErrorMessage<ErrorCode>(
  error: CommonError<ErrorCode> | undefined,
  intl: IntlShape
): string {
  if (error) {
    switch (error.code) {
      case CommonErrorCode.GRAPHQL_ERROR:
        return intl.formatMessage(commonErrorMessages.graphqlError);
      case CommonErrorCode.REQUIRED:
        return intl.formatMessage(commonMessages.requiredField);
      case CommonErrorCode.INVALID:
        return intl.formatMessage(commonErrorMessages.invalid);

      default:
        return intl.formatMessage(commonErrorMessages.unknownError);
    }
  }

  return undefined;
}
```

src/utils/errors/common.ts
```diff
@@ -12,4 +13,35 @@ const commonErrorMessages = defineMessages({
   }
 });
 
+enum CommonErrorCode {
+  GRAPHQL_ERROR = "GRAPHQL_ERROR",
+  INVALID = "INVALID",
+  REQUIRED = "REQUIRED"
+}
```
jwm0, line 20: This should just be a string union

jwm0 (changes requested):

src/giftCards/GiftCardSettings/messages.ts
jwm0: I like it but there's no need to have those keys as enums, use string unions:
```
type CommonErrorCode = "GRAPHQL_ERROR" | ....
getCommonFormFieldErrorMessage(error: CommonErrorCode, intl: IntlShape)
```

mmarkusik (approved):

eCzerniak (approved):

jwm0 (approved):

PR #1342, opened 2021-08-30
Fix products table content display issue when item is selected

kamilpastuszka:
I want to merge this to fix products' content display issue when item is selected.
 

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/products/components/ProductList/ProductList.tsx
+++ b/src/products/components/ProductList/ProductList.tsx
@@ -42,7 +42,7 @@ const useStyles = makeStyles(
   theme => ({
     [theme.breakpoints.up("md")]: {
       colName: {
-        minWidth: 200
+        minWidth: 250
       },
       colPrice: {
         width: 300
--- a/src/products/components/ProductList/ProductList.tsx
+++ b/src/products/components/ProductList/ProductList.tsx
@@ -40,9 +40,9 @@ import { messages } from "./messages";
 
 const useStyles = makeStyles(
   theme => ({
-    [theme.breakpoints.up("lg")]: {
+    [theme.breakpoints.up("md")]: {
       colName: {
-        width: "auto"
+        minWidth: 200
       },
       colPrice: {
         width: 300
--- a/src/products/components/ProductList/ProductList.tsx
+++ b/src/products/components/ProductList/ProductList.tsx
@@ -147,9 +147,7 @@ export const ProductList: React.FC<ProductListProps> = props => {
       <ResponsiveTable className={classes.table}>
         <colgroup>
           <col />
-          <DisplayColumn column="productName" displayColumns={settings.columns}>
-            <col className={classes.colName} />
-          </DisplayColumn>
+          <col className={classes.colName} />
           <DisplayColumn column="productType" displayColumns={settings.columns}>
             <col className={classes.colType} />
           </DisplayColumn>
@@ -328,7 +326,6 @@ export const ProductList: React.FC<ProductListProps> = props => {
                     />
                   </TableCell>
                   <TableCellAvatar
-                    className={classes.colName}
                     thumbnail={maybe(() => product.thumbnail.url)}
                   >
                     {product?.productType ? (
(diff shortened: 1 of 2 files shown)


patrys:
Task linked: [SALEOR-4249 Products table content breaks when item is selected](https://app.clickup.com/t/2549495/SALEOR-4249)

dominik-zeglen (changes requested):
It doesn't fix root cause which is layout shift after selecting, which is caused by setting column width on `td` element containing product name, not `col`

kamilpastuszka:
The existing solution also fixes bug with layout when nothing is selected.
Please provide your solution and I will test it.

jwm0 (changes requested):
Texts no longer collide but there's a visible layout shift when item is selected, there should be none.

src/products/components/ProductList/ProductList.tsx
```diff
@@ -147,7 +147,9 @@ export const ProductList: React.FC<ProductListProps> = props => {
       <ResponsiveTable className={classes.table}>
         <colgroup>
           <col />
-          <col className={classes.colName} />
+          <DisplayColumn column="productName" displayColumns={settings.columns}>
+            <col className={classes.colName} />
+          </DisplayColumn>
```
jwm0, line 152: The `DisplayColumn` component is meant to be used when the column can be hidden, product name should always be visible as opposed to the rest of the columns. Additionally, you wrapped the table head column but not the body itself, which would break the table if `productName` is supposed to be hidden.

src/storybook/__snapshots__/Stories.test.ts.snap
```diff
@@ -2107,7 +2107,7 @@ exports[`Storyshots Attributes / Attributes selected 1`] = `
                 class="MuiInputBase-input-id MuiOutlinedInput-input-id"
                 name="attribute:Date Time Attribute:time"
                 type="time"
-                value="00:15"
+                value="02:15"
```
jwm0, line 2110: Use the environment variable, if the command does not work on your system - set the env manually or modify the command to be cross-platform.

jwm0 (changes requested):
I see some issues:
1. When no columns are selected (only price & availability) the `productName` column unsticks from the left side (this is not present on current master.staging):
image

2.  When the table overflows horizontally and is selected, the last element is not within the highlight section:
image

src/products/components/ProductList/ProductList.tsx
kamilpastuszka: I know, it was discussed already. The fix will be improved.

src/storybook/__snapshots__/Stories.test.ts.snap
kamilpastuszka: Indeed, it is probably due to some problems with that command on Windows. That was the output after making the command cross-platform. Setting env manually did't work for me.

patrys: Call `set TZ=UTC` before running `jest`. Removing half of the command is not the same as making it cross-platform 😉

kamilpastuszka: I merely preceded the command with `cross-env`. I didn’t modify it in any other way.

jwm0 (changes requested):
I noticed one more issue with the table that might be worth fixing in this PR. When it overflows horizontally the table overlaps with navigation on the left.

Table without overflow:
image

Table with overflow:
image

dominik-zeglen (changes requested):
Layout shift is still visible

kamilpastuszka:
@dominik-zeglen  Just to confirm – are you talking about the initial scope of the ticket or the issue reported by @jwm0 ?

mmarkusik (approved):
Code-wise looks okay, when other issues mentioned are resolved

kamilpastuszka:
The other issue is covered in PR #1382

dominik-zeglen:
Zrzut ekranu 2021-09-13 o 14 52 21
Zrzut ekranu 2021-09-13 o 14 52 26

Notice how column with product name collapses, it shouldn't do it

jwm0 (dismissed):
You fixed the CLS when the columns overflow but now it breaks when there's no overflow

kamilpastuszka:
@jwm0 Can you provide more information?

jwm0 (approved):

dominik-zeglen (approved):

PR #1419, opened 2021-09-17
Fix issue with hidden country dropdown in Orders Section 

kamilpastuszka:
This PR fixes hidden country dropdown in Orders.

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/orders/components/OrderAddressEditDialog/OrderAddressEditDialog.tsx
+++ b/src/orders/components/OrderAddressEditDialog/OrderAddressEditDialog.tsx
@@ -27,7 +27,8 @@ import { FormattedMessage, useIntl } from "react-intl";
 const useStyles = makeStyles(
   {
     overflow: {
-      overflowY: "visible"
+      overflowY: "visible",
+      overflowX: "inherit"
     }
   },
   { name: "OrderAddressEditDialog" }


patrys:
Task linked: [SALEOR-4409 Orders - country dropdown is hidden](https://app.clickup.com/t/2549495/SALEOR-4409)

orzechdev (approved):

src/orders/components/OrderAddressEditDialog/OrderAddressEditDialog.tsx
```diff
@@ -27,7 +27,7 @@ import { FormattedMessage, useIntl } from "react-intl";
 const useStyles = makeStyles(
   {
     overflow: {
-      overflowY: "visible"
+      overflow: "visible"
```
jwm0, line 30: Instead of manually overriding the values in modals you should fix the core issue that was introduced in this commit: https://github.com/mirumee/macaw-ui/commit/73dcbb381ee9bb165d13695063a54299bb9ebf32

jwm0 (changes requested):

src/orders/components/OrderAddressEditDialog/OrderAddressEditDialog.tsx
kamilpastuszka: The value “hidden” there in that commit fixes overflow issue in a number of modals. 
So that value should stay or alternatively it can be passed dynamically.
But I don’t think so there would be much gain since it only breaks UI in just one place.
I think the real reason for this bug is dropdown opening outside of the modal.

jwm0: I believe the content should guard against unwanted overflows, not the dialog itself. What if you want to display a large horizontal table inside a modal? There are (rare) cases where overflow might be necessary and when it isn't expected it should be fixed on that particular use case. By default the modals shouldn't overflow horizontally if they're using proper CSS.

kamilpastuszka: I think it doesn’t matter. A modal should be responsive enough to handle such a big table. In case of that vertical overflow in Orders, the modal should extend itself to encompass the dropdown. In some cases, modals should go full viewport height if that’s required.

jwm0: In this PR you're fixing only one dialog that has been diagnosed with this bug, but if you CMD + F you'll see a buch more instances. We **do not want** to set explicit overflow on the dialog, the content itself should determine how and if it wants to overflow the content.

kamilpastuszka: Solving this the way you see it is currently blocked by a change that needs to be implemented in Dashboard. It relates to Macaw-UI. I’ve added relevant info to the ticket.

PR #1425, opened 2021-09-21
Add page type and attributes permissions handling

Cloud11PL:
I want to merge this change because I wanted to add handling for missing permissions for page type and attributes. 

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/configuration/utils.ts
+++ b/src/configuration/utils.ts
@@ -12,4 +12,4 @@ export const getConfigMenuItemsPermissions = (
         prev.concat(next.menuItems.map(menuItem => menuItem.permissions)),
       []
     )
-    .reduce((prev, next) => prev.concat(next), []);
+    .flat();
--- a/src/configuration/utils.ts
+++ b/src/configuration/utils.ts
@@ -8,8 +8,10 @@ export const getConfigMenuItemsPermissions = (
 ): PermissionEnum[] =>
   createConfigurationMenu(intl)
     .reduce(
-      (prev, next) =>
-        prev.concat(next.menuItems.map(menuItem => menuItem.permissions)),
+      (prev, { menuItems }) => [
+        ...prev,
+        ...menuItems.map(({ permissions }) => permissions)
+      ],
       []
     )
     .flat();
--- a/src/configuration/utils.ts
+++ b/src/configuration/utils.ts
@@ -1,0 +1,15 @@
+import { PermissionEnum } from "@saleor/types/globalTypes";
+import { IntlShape } from "react-intl";
+
+import { createConfigurationMenu } from ".";
+
+export const getConfigMenuItemsPermissions = (
+  intl: IntlShape
+): PermissionEnum[] =>
+  createConfigurationMenu(intl)
+    .reduce(
+      (prev, next) =>
+        prev.concat(next.menuItems.map(menuItem => menuItem.permissions)),
+      []
+    )
+    .reduce((prev, next) => prev.concat(next), []);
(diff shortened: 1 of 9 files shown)


patrys:
Task linked: [SALEOR-3164 [Permissions] Page Type & Attribute - user cannot see configuration menu.](https://app.clickup.com/t/2549495/SALEOR-3164)

src/configuration/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const getConfigMenuItemsPermissions = (
+  intl: IntlShape
+): PermissionEnum[] =>
+  createConfigurationMenu(intl)
+    .reduce(
+      (prev, next) =>
+        prev.concat(next.menuItems.map(menuItem => menuItem.permissions)),
+      []
+    )
```
jwm0, line 14: If you're using `reduce` and the 2 conditions are met:
1. Output is an array
2. The length of the input and output array is the same

then it means it can be replaced with `.map`:
```suggestion
    .map(menu => menu.menuItems.map(menuItem => menuItem.permissions))
```

src/configuration/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+    .reduce((prev, next) => prev.concat(next), []);
```
jwm0, line 15: There's a built-in method for that called `.flat()`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

jwm0 (changes requested):

Droniu (approved):

src/configuration/utils.ts
Cloud11PL: @jwm0 I think I initially wrote something like you suggested byt the length of the output array is different compared to input

jwm0: Ok, in this case the final `flat` (or reduce you used before) is not needed. I'd also suggest to use the spread operator as I believe it makes it more readable and inline with the rest of the codebase:
```js
.reduce(
      (prev, { menuItems }) =>
        [...prev, ...menuItems.map(({ permissions })=> permissions)],
      []
    )
.reduce() /// <-- remove this part!!!
```

src/configuration/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+export const getConfigMenuItemsPermissions = (
+  intl: IntlShape
+): PermissionEnum[] =>
+  createConfigurationMenu(intl)
+    .reduce(
+      (prev, next) =>
+        prev.concat(next.menuItems.map(menuItem => menuItem.permissions)),
+      []
+    )
+    .flat();
```
jwm0, line 15: This is no longer needed, you're already flattening the array in the step above.

src/index.tsx
```diff
@@ -303,9 +311,7 @@ const Routes: React.FC = () => {
                 path={channelsSection}
                 component={ChannelsSection}
               />
-              {createConfigurationMenu(intl).filter(menu =>
-                menu.menuItems.map(item => hasPermission(item.permission, user))
-              ).length > 0 && (
+              {hasAnyPermissions(getConfigMenuItemsPermissions(intl), user) && (
```
jwm0, line 314: Why not use the existing logic in `SectionRoute` component?

```jsx
<SectionRoute
  permissions={getConfigMenuItemsPermissions(intl))}
  matchPermission="any"
  exact
  path="/configuration"
  component={ConfigurationSection}
/>
```

jwm0 (changes requested):

jwm0 (approved):

eCzerniak (approved):

PR #1430, opened 2021-09-22
Add content type for file attribute

kamilpastuszka:
This PR adds content type for file uploads. Without it response from the server is incomplete.

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -78,7 +78,8 @@ errorTracker.init();
 // so we need to explicitly set them
 const linkOptions = {
   credentials: "include",
-  uri: API_URI
+  uri: API_URI,
+  type: "image/png"
 };
 const uploadLink = createUploadLink(linkOptions);
 const batchLink = new BatchHttpLink({


patrys:
Task linked: [SALEOR-2295 Dashboard should set up contentType for File attribute](https://app.clickup.com/t/2549495/SALEOR-2295)

orzechdev (approved):

src/index.tsx
```diff
@@ -78,7 +78,8 @@ errorTracker.init();
 // so we need to explicitly set them
 const linkOptions = {
   credentials: "include",
-  uri: API_URI
+  uri: API_URI,
+  type: "image/png"
```
jwm0, line 82: Files can have many content types, we shouldn't hardcore them

jwm0 (changes requested):

src/index.tsx
kamilpastuszka: Reasoning behind the proposed solution:
 - based on the ticket description it’s about images
 - there is no generic image media-type and Content-Type can accept only single media-type   
 - I tested it with other image extensions and it looked good
 - png file extension is in the server response example in the ticket, so the solution seemed    
   reasonable for me also based on that

PR #1382, opened 2021-09-13
Fix sidebar navigation and Products Page overlapping problem

kamilpastuszka:
This PR is to fix content overlapping issue between sidebar navigation and Products Page.

**PR intended to be tested with API branch:** 

### Pull Request Checklist

1. [ ] This code contains UI changes
2. [ ] All visible strings are translated with proper context including data-formatting
3. [ ] Attributes `[data-test-id]` are added for new elements
4. [ ] Changes are mentioned in the changelog
5. [ ] The changes are tested in different browsers and in light/dark mode

### Test environment config

API_URI=https://master.staging.saleor.cloud/graphql/

PR #1472, opened 2021-10-06
Fix sidebar navigation and Products Page overlapping problem (#1382)

jwm0:
Porting #1382


patrys:
Task linked: [SALEOR-4412 Sidebar navigation and Products Page overlapping problem](https://app.clickup.com/t/2549495/SALEOR-4412)

PR #1561, opened 2021-11-03
Add scrollable area for user addresses

Droniu:
I want to merge this change because it fixes a problem with address cards in order's add customer address dialog.

**PR intended to be tested with API branch:**  3.0

### Screenshots

Before:
image
After:
image

### Pull Request Checklist

1. [ ] This code contains UI changes
2. [ ] All visible strings are translated with proper context including data-formatting
3. [ ] Attributes `[data-test-id]` are added for new elements
4. [ ] Changes are mentioned in the changelog
5. [ ] The changes are tested in different browsers and in light/dark mode

### Test environment config

API_URI=https://master.staging.saleor.cloud/graphql/

PR #1564, opened 2021-11-03
Add scrollable area for user addresses (#1561)

Droniu:
Cherry pick of #1561 

* add scrollable area for user addresses

* fix billing address

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/orders/components/OrderCustomerAddressesEditDialog/styles.ts
+++ b/src/orders/components/OrderCustomerAddressesEditDialog/styles.ts
@@ -2,6 +2,14 @@ import { makeStyles } from "@saleor/macaw-ui";
 
 export const useStyles = makeStyles(
   {
+    scrollableContent: {
+      maxHeight: `calc(100vh - 250px)`,
+      overflow: "scroll"
+    },
+    scrollableWrapper: {
+      maxHeight: 400,
+      overflow: "scroll"
+    },
     container: {
       display: "block"
     },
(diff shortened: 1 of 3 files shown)


patrys:
Task linked: [SALEOR-4840 Issue with scrolling user address list](https://app.clickup.com/t/2549495/SALEOR-4840)

dominik-zeglen (approved):

src/orders/components/OrderCustomerAddressesEditDialog/styles.ts
```diff
@@ -2,6 +2,14 @@ import { makeStyles } from "@saleor/macaw-ui";
 
 export const useStyles = makeStyles(
   {
+    scrollableContent: {
+      maxHeight: `calc(100vh - 250px)`,
```
Cloud11PL, line 6: Where does `250px` come from? If it's used somewhere else (I suspect it is) maybe it would be better to make a const and use it here?

Droniu: It's not. Originally I thought about using theme.spacing but in this place 1 spacing is 0.8 rem, so to achieve desired result value needs to be something like `theme.spacing(26)`.

250px is arbitrary number leading to proper, good-looking margins on the top and bottom of the modal without making the css behaviour unexpected.

Cloud11PL (approved):

src/orders/components/OrderCustomerAddressesEditDialog/styles.ts
dominik-zeglen: Arbitrarily chosen value, it'll be removed in near future (this sprint), so I wouldn't overthink it

PR #1696, opened 2021-12-17
Use variant prices in add product modal

Cloud11PL:
I want to merge this change because I want to change pricing from channel to variant. 

3.1 PR: https://github.com/saleor/saleor-dashboard/pull/1698

### Screenshots
New "sale" price labels
image

### Pull Request Checklist

### Test environment config

API_URI=https://stable.staging.saleor.cloud/graphql/

### DIFF
--- a/src/misc.ts
+++ b/src/misc.ts
@@ -360,7 +360,7 @@ export function addressToAddressInput<T>(
   const { id, __typename, ...rest } = address;
   return {
     ...rest,
-    country: findInEnum(address?.country?.code, CountryCode)
+    country: findInEnum(address.country.code, CountryCode)
   };
 }
 
--- a/src/misc.ts
+++ b/src/misc.ts
@@ -13,6 +13,7 @@ import {
   orderStatusMessages,
   paymentStatusMessages
 } from "./intl";
+import { OrderDetails_order_shippingAddress } from "./orders/types/OrderDetails";
 import {
   MutationResultAdditionalProps,
   PartialMutationProviderOutput,
@@ -351,6 +352,16 @@ export function findInEnum<TEnum extends {}>(needle: string, haystack: TEnum) {
   }
 
   throw new Error(`Key ${needle} not found in enum`);
+}
+
+export function addressToAddressInput<T>(
+  address: T & OrderDetails_order_shippingAddress
+): AddressInput {
+  const { id, __typename, ...rest } = address;
+  return {
+    ...rest,
+    country: findInEnum(address?.country?.code, CountryCode)
+  };
 }
 
 export function findValueInEnum<TEnum extends {}>(
--- a/src/orders/components/OrderPriceLabel/OrderPriceLabel.tsx
+++ b/src/orders/components/OrderPriceLabel/OrderPriceLabel.tsx
@@ -2,6 +2,7 @@ import DiscountedPrice from "@saleor/components/DiscountedPrice/DiscountedPrice"
 import Money from "@saleor/components/Money";
 import { SearchOrderVariant_search_edges_node_variants_pricing } from "@saleor/orders/types/SearchOrderVariant";
 import * as React from "react";
+
 import { useStyles } from "./styles";
 
 interface OrderPriceLabelProps {
--- a/src/orders/components/OrderPriceLabel/OrderPriceLabel.tsx
+++ b/src/orders/components/OrderPriceLabel/OrderPriceLabel.tsx
@@ -15,10 +15,6 @@ const useStyles = makeStyles(theme => ({
     alignItems: "flex-end",
     justifyContent: "flex-end"
   },
-  subtitle: {
-    color: theme.palette.grey[500],
-    paddingRight: theme.spacing(1)
-  },
   strike: {
     textDecoration: "line-through",
     color: theme.palette.grey[400],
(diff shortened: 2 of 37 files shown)


patrys:
Task linked: [SALEOR-5073 Use variant prices in "Add product" modal](https://app.clickup.com/t/2549495/SALEOR-5073)

src/orders/components/OrderPriceLabel/OrderPriceLabel.tsx
```diff
@@ -0,0 +1,46 @@
...
+      justifyContent: "flex-end"
+    },
+    strike: {
+      textDecoration: "line-through",
+      color: theme.palette.grey[400],
+      fontSize: "smaller"
+    }
+  }),
+  { name: "OrderPriceLabel" }
+);
```
Droniu, line 26: Can be moved to seperate `styles.ts` file

src/misc.ts
```diff
@@ -353,6 +354,16 @@ export function findInEnum<TEnum extends {}>(needle: string, haystack: TEnum) {
...
+export function addressToAddressInput<T>(
+  address: T & OrderDetails_order_shippingAddress
+): AddressInput {
+  const { id, __typename, ...rest } = address;
+  return {
+    ...rest,
+    country: findInEnum(address?.country?.code, CountryCode)
+  };
+}
+
```
Droniu, line 366: So now in `misc` there are 3 functions which somehow transform address to address input (`transformFormToAddressInput`, `transformAddressToAddressInput`, `addressToAddressInput`). What are the differences between them, is it possible to merge them?

src/orders/components/OrderProductAddDialog/OrderProductAddDialog.tsx
```diff
@@ -363,9 +363,7 @@ const OrderProductAddDialog: React.FC<OrderProductAddDialogProps> = props => {
                             </div>
                           </TableCell>
                           <TableCell className={classes.textRight}>
-                            {variant?.channelListings[0]?.price && (
-                              <Money money={variant.channelListings[0].price} />
-                            )}
+                            <OrderPriceLabel pricing={variant.pricing} />
```
Droniu, line 366:
```suggestion
                            <OrderPriceLabel pricing={variant?.pricing} />
```
I'm pretty sure optional chaining should stay here, because variant field tend to be undefined in many cases

Droniu (changes requested):

src/misc.ts
Cloud11PL: `transformFormToAddressInput` returns `T & AddressInput`
`transformAddressToAddressInput` returns `AddressInput` and merges customer data
`addressToAddressInput` returns `AddressInput` without `id` and `__typename`

I would rather keep them separate

src/orders/components/OrderPriceLabel/OrderPriceLabel.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const OrderPriceLabel: React.FC<OrderPriceLabelProps> = ({ pricing }) => {
```
jwm0, line 28: It would be good to find out if this is not a reusable component we might use outside of `orders` module, can you please check with @pwgryglak to see if that's the case.

src/orders/utils/data.ts
```diff
@@ -208,3 +210,15 @@ export function mergeRepeatedOrderLines(
     return prev;
   }, Array<OrderDetails_order_fulfillments_lines>());
 }
+
+export const getVariantSearchAddress = (order: OrderDetails_order) => {
+  if (!!order.shippingAddress) {
```
jwm0, line 215: `!!` are not needed inside statements as they are automatically cast to booleans

src/orders/utils/data.ts
```diff
@@ -208,3 +210,15 @@ export function mergeRepeatedOrderLines(
...
+    return addressToAddressInput(order.shippingAddress);
+  }
+
+  if (!!order.billingAddress) {
+    return addressToAddressInput(order.billingAddress);
+  }
+
+  return { country: order.channel.defaultCountry.code } as AddressInput;
```
jwm0, line 223: Perhaps it would make more sense to type the outcome of this function instead? (assuming it's always AdressInput for all returns)
```
getVariantSearchAddress = (order: OrderDetails_order): AddressInput
```

jwm0 (changes requested):

src/orders/components/OrderProductAddDialog/OrderProductAddDialog.tsx
Cloud11PL: If `product.variants` is empty then the whole table row is not rendered

src/orders/components/OrderPriceLabel/OrderPriceLabel.tsx
Cloud11PL: There is a similar implementation in `TableLine.tsx` but the styles are different and the discounted price text acts as a link there. I'll extract the two prices together to a separate component as it seems to be most likely reused somewhere else - I'll keep `OrderPriceLabel` because it's a wrapper and it will vary from the needed implementation.

jwm0: I see you extracted the code but no changes were made to `TableLine.tsx` file

Cloud11PL: I used `TableLine.tsx` as an example that there was a single similar component but since it's a link I can't change it to use the new "reusable" component 😛 I extracted it for the future non-link "sale" prices

jwm0 (approved):

Droniu (approved):

eCzerniak (approved):

PR #1729, opened 2022-01-03
Fix switch account login and credentials API wrong login loop issues

orzechdev:

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/auth/hooks/useAuthProvider.ts
+++ b/src/auth/hooks/useAuthProvider.ts
@@ -48,6 +48,7 @@ export function useAuthProvider({
   } = useAuth();
   const { authenticated, authenticating, user } = useAuthState();
   const [error, setError] = useState<UserContextError>();
+  const [permitCredentialsAPI, setPermitCredentialsAPI] = useState(true);
 
   useEffect(() => {
     if (authenticating && error) {
@@ -56,15 +57,23 @@ export function useAuthProvider({
   }, [authenticating]);
 
   useEffect(() => {
-    if (!authenticated && !authenticating) {
+    if (!authenticated && !authenticating && permitCredentialsAPI) {
+      setPermitCredentialsAPI(false);
       loginWithCredentialsManagementAPI(handleLogin);
     }
   }, [authenticated, authenticating]);
 
   const userDetails = useQuery<UserDetails>(userDetailsQuery, {
     client: apolloClient,
-    skip: !authenticated
+    skip: !authenticated,
+    fetchPolicy: "cache-and-network"
   });
+
+  useEffect(() => {
+    if (authenticated) {
+      setPermitCredentialsAPI(true);
+    }
+  }, [authenticated]);
 
   const handleLogout = async () => {
     const result = await logout({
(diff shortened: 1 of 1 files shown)


patrys:
Task linked: [SALEOR-5597 Issues with authentication](https://app.clickup.com/t/2549495/SALEOR-5597)

src/auth/hooks/useAuthProvider.ts
```diff
@@ -56,16 +57,24 @@ export function useAuthProvider({
...
+      setPermitCredentialsAPI(false);
       loginWithCredentialsManagementAPI(handleLogin);
     }
   }, [authenticated, authenticating]);
 
   const userDetails = useQuery<UserDetails>(userDetailsQuery, {
     client: apolloClient,
-    skip: !authenticated
+    skip: !authenticated,
+    fetchPolicy: "cache-and-network"
```
jwm0, line 69: Shouldn't this be the default "network-only"? Do we ever fetch this from cache?

src/auth/hooks/useAuthProvider.ts
```diff
@@ -56,16 +57,24 @@ export function useAuthProvider({
...
   });
 
+  useEffect(() => {
+    if (authenticated) {
+      setPermitCredentialsAPI(true);
```
jwm0, line 74: Would it make more sense to trigger this on user logout? (inside `isCredentialsManagementAPISupported` if statement)

src/auth/hooks/useAuthProvider.ts
```diff
@@ -48,6 +48,7 @@ export function useAuthProvider({
   } = useAuth();
   const { authenticated, authenticating, user } = useAuthState();
   const [error, setError] = useState<UserContextError>();
+  const [permitCredentialsAPI, setPermitCredentialsAPI] = useState(true);
```
jwm0, line 51: Internal values that do not correspond to UI updates should not be kept inside state. This can be safely extracted outside component or wrapped in useRef:
```
let permitCredentialsAPI = true;

export function useAuthProvider({
```

jwm0 (changes requested):

Cloud11PL (approved):

src/auth/hooks/useAuthProvider.ts
orzechdev: Theoretically, the cache could be used on page refresh, this is why I left "cache-and-network".

jwm0: Apollo cache is only there for the duration of runtime, it's gone when you do a full page refersh (it's not persisted in local/session storage)

src/auth/hooks/useAuthProvider.ts
orzechdev: Not sure, because what if token expires? It won't be logout this way.

jwm0 (approved):

eCzerniak (approved):

src/auth/hooks/useAuthProvider.ts
orzechdev: However, change again to "cache-and-network" might be needed in https://github.com/saleor/saleor-dashboard/pull/1792,

PR #1742, opened 2022-01-11
Redirect user back to the first landing page after external login

Cloud11PL:
I want to merge this change because it redirects the user back to the subpage they first landed on. 

3.0 PR: https://github.com/saleor/saleor-dashboard/pull/1743

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/auth/views/Login.tsx
+++ b/src/auth/views/Login.tsx
@@ -42,6 +42,11 @@ const LoginView: React.FC<LoginViewProps> = ({ params }) => {
     setRequestedExternalPluginId
   ] = useLocalStorage("requestedExternalPluginId", null);
 
+  const [fallbackUri, setFallbackUri] = useLocalStorage(
+    "externalLoginFallbackUri",
+    null
+  );
+
   const handleSubmit = async (data: LoginFormData) => {
     const result = await login(data.email, data.password);
     const errors = result?.errors || [];
@@ -50,6 +55,8 @@ const LoginView: React.FC<LoginViewProps> = ({ params }) => {
   };
 
   const handleRequestExternalAuthentication = async (pluginId: string) => {
+    setFallbackUri(location.pathname);
+
     const result = await requestLoginByExternalPlugin(pluginId, {
       redirectUri: urlJoin(
         window.location.origin,
@@ -70,7 +77,8 @@ const LoginView: React.FC<LoginViewProps> = ({ params }) => {
       state
     });
     if (result && !result?.errors?.length) {
-      navigate(APP_DEFAULT_URI);
+      navigate(fallbackUri ?? APP_DEFAULT_URI);
+      setFallbackUri(null);
     }
   };
 
--- a/src/auth/views/Login.tsx
+++ b/src/auth/views/Login.tsx
@@ -77,7 +77,7 @@ const LoginView: React.FC<LoginViewProps> = ({ params }) => {
       state
     });
     if (result && !result?.errors?.length) {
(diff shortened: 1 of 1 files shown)


patrys:
Task linked: [SALEOR-3445 Dropped redirection on OIDC login](https://app.clickup.com/t/2549495/SALEOR-3445)

orzechdev (approved):

src/auth/views/Login.tsx
```diff
@@ -50,6 +55,8 @@ const LoginView: React.FC<LoginViewProps> = ({ params }) => {
   };
 
   const handleRequestExternalAuthentication = async (pluginId: string) => {
+    setFallbackUri(location.pathname);
```
jwm0, line 58: Can we not add it as a query parameter to `redirectUri` below?
```
      redirectUri: urlJoin(
        window.location.origin,
        APP_MOUNT_URI === APP_DEFAULT_URI ? "" : APP_MOUNT_URI,
        loginCallbackPath,
        `?fallback=${location.pathname}`
      )
```

jwm0 (changes requested):

src/auth/views/Login.tsx
Cloud11PL: I tried this solution, it would require backend changes as backend's response includes only the callback url, additionally auth0 is strict on the callback path - from what i've seen it cannot be dynamic. Auth0, in our case, recommends storing the path we call `fallback` in the browser storage.

jwm0: In that case let's just put it in the url. `Login` view uses a non-exact path match, so it will match all pathnames that look like this: `/login/*`. We can leverage this and append our url directly to the main pathname:
```
   redirectUri: urlJoin(
       location.origin,
        APP_MOUNT_URI === APP_DEFAULT_URI ? "" : APP_MOUNT_URI,
        loginCallbackPath,
       location.pathname
      )
```

Then we can extract it from the url when `handleExternalAuthentication` is called. No changes are needed to path detection as:
```
const isCallbackPath = location.pathname.includes(loginCallbackPath);
```
will continue to work as it used to

Cloud11PL: The solutions work the same as in your previous comment - passing `location.pathname` to the `requestLoginByExternalPlugin`. `requestLoginByExternalPlugin` sends it off to the auth0 and it won't accept an uri that is not predefined in the auth0 dashboard - it cannot be dynamic. 
It causes the following error:
```
Callback URL mismatch.
The provided redirect_uri is not in the list of allowed callback URLs.
```
It would work on our side but not on the auth0 side, therefore we will have to stick with storing the pathname in browser storage as it is recommended by auth0 (eg. here https://community.auth0.com/t/how-do-i-set-up-a-dynamic-allowed-callback-url/60268 but most importantly here https://auth0.com/docs/secure/attack-protection/state-parameters)

src/auth/views/Login.tsx
```diff
@@ -70,7 +77,8 @@ const LoginView: React.FC<LoginViewProps> = ({ params }) => {
       state
     });
     if (result && !result?.errors?.length) {
-      navigate(APP_DEFAULT_URI);
+      navigate(fallbackUri ?? APP_DEFAULT_URI);
```
jwm0, line 80: When using `react-router` navigate do not include the `APP_DEFAULT_URI` as it's treated as default url basename. You can see it here:
```
<BrowserRouter basename={APP_MOUNT_URI}>
```

jwm0 (changes requested):

jwm0 (approved):

PR #1772, opened 2022-01-19
Fix double authentication

orzechdev:
I want to merge this change because it fixes issue with double authentication (signing using browser's navigator.credentials while already logged-in via external provider)

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/auth/hooks/useAuthProvider.ts
+++ b/src/auth/hooks/useAuthProvider.ts
@@ -1,5 +1,6 @@
 import { IMessageContext } from "@saleor/components/messages";
 import { APP_DEFAULT_URI, APP_MOUNT_URI, DEMO_MODE } from "@saleor/config";
+import useLocalStorage from "@saleor/hooks/useLocalStorage";
 import useNavigator from "@saleor/hooks/useNavigator";
 import { commonMessages } from "@saleor/intl";
 import {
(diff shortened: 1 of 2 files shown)


patrys:
Task linked: [SALEOR-5692 Dashboard allows for double-authentication when using external login and navigator.credentials api](https://app.clickup.com/t/2549495/SALEOR-5692)

src/auth/hooks/useAuthProvider.ts
```diff
@@ -65,7 +70,12 @@ export function useAuthProvider({
   }, [authenticated]);
 
   useEffect(() => {
-    if (!authenticated && !authenticating && permitCredentialsAPI.current) {
+    if (
+      !authenticated &&
+      !authenticating &&
+      !requestedExternalPluginId &&
```
jwm0, line 76:
```suggestion
      !requestedExternalPluginId &&
```
doesn't this condition always resolve to `true` when coupled with:
```
    setRequestedExternalPluginId(null);
```
on sucess external authentication? Because the initial value from localStorage will be null and the value you get from `useLocalStorage` is not reactive afaik (it does not listen on changes).

jwm0: Hmm this logic makes sense between redirects, so if that was your intention it seems 👍

orzechdev: Yes, it should work between redirects

dominik-zeglen: I guess `requestedExternalPluginId` it's used by sdk?

orzechdev: It is used by `LoginView`

orzechdev: ^ To determine what external plugin was used to log in, after redirect

jwm0 (approved):

dominik-zeglen (approved):

PR #1784, opened 2022-01-23
Add hover tooltip on sorting disabled columns in lists

Droniu:
I want to merge this change because it adds an info tooltip when active filter is necessary to sort particular column in lists.

Currently only `channel` filter affects sorting in any way, but it will also be `currency` in gift cards.

**PR intended to be tested with API branch:**  3.0

### Screenshots

image

### Pull Request Checklist

### Test environment config

API_URI=https://stable.staging.saleor.cloud/graphql/


patrys:
Task linked: [SALEOR-3637 Add hover tooltip on sorting disabled columns](https://app.clickup.com/t/2549495/SALEOR-3637)

src/components/TooltipTableCellHeader/TooltipTableCellHeader.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <Tooltip
+      title={intl.formatMessage(
+        !!overrideMessage ? overrideMessage : messages.noFilterSelected,
+        {
+          filterName: filterDependency
+        }
+      )}
+      placement="top"
+      disabled={tooltipDisabled()}
+    >
```
orzechdev, line 43: Can we improve the interface a bit? I think, ideally, I should be able to say what this particular component does only based on the interface it provides like its name and props without knowing anything about the implementation and without knowing in what context it is used. `TooltipTableCellHeader` looks like generic name, while `filterDependency` looks not connected to it in any way.
I guess it would be cleaner to move message outside of component permanently. What do you think about it? Do you agree with me? 
```suggestion
  const {
    children,
    tooltip,
    disabled,
    ...rest
  } = props;

  const intl = useIntl();

  const tooltipDisabled = () => {
    if (!tooltip) {
      return true;
    }
    return !disabled;
  };

  return (
    <Tooltip
      title={tooltip}
      placement="top"
      disabled={tooltipDisabled()}
    >
```

Droniu: That sounds like a good idea, however this message is repeated across multiple lists and I don't want to duplicate it. Maybe putting it in `src/intl.ts` would be a good solution here?

Cloud11PL: I think `src/intl.ts` is a bit too general - I think I would just put it in the same folder as this tooltip.

src/components/TooltipTableCellHeader/RefTableCellHeader.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+import TableCellHeader, { TableCellHeaderProps } from "../TableCellHeader";
+
+// Workaround because child of <Tooltip /> must be able to hold ref
```
Cloud11PL, line 5: Why?

Droniu: This is just how MUI Tooltip works - in order to show, its child has to be able to access DOM node and it is done by giving a ref to the component. Custom function components like `TableCellHeader` can't hold refs so `React.forwardRef` is needed. Further reading: https://v4.mui.com/guides/composition/#caveat-with-refs

src/components/TooltipTableCellHeader/TooltipTableCellHeader.tsx
orzechdev: Yep, imo might be in put it in the same folder as well

orzechdev (approved):

Cloud11PL (approved):

src/components/TooltipTableCellHeader/RefTableCellHeader.tsx
jwm0: Directly modify TableCellHeader component and wrap it inside forwardRef instead

package.json
```diff
@@ -28,7 +28,7 @@
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@material-ui/styles": "^4.11.4",
-    "@saleor/macaw-ui": "^0.3.0",
+    "@saleor/macaw-ui": "github:saleor/macaw-ui#6a8a1e8",
```
jwm0, line 31: Let's release a new version before merging this in

jwm0 (changes requested):

jwm0 (approved):

PR #1785, opened 2022-01-24
Support apps in navigation

orzechdev:
I want to merge this change because it adds support for extensions placed as sidebar menu items.

Uses macaw-ui changes: https://github.com/saleor/macaw-ui/pull/106, https://github.com/saleor/macaw-ui/pull/112
Backend: https://github.com/saleor/saleor/pull/8959

**PR intended to be tested with API branch:**   add-new-app-extensions

### Screenshots

Zrzut ekranu 2022-02-1 o 15 06 04

### Pull Request Checklist

### Test environment config

API_URI=https://add-new-app-extensions.api.saleor.rocks/graphql/


patrys:
Task linked: [SALEOR-4918 Extend dashboard navigation to support Apps](https://app.clickup.com/t/2549495/SALEOR-4918)

schema.graphql
```diff
@@ -5,39 +5,27 @@ schema {
 
 type AccountAddressCreate {
   user: User
-  accountErrors: [AccountError!]!
```
bmigirl: Is this formatting on purpose?

src/apps/views/App/App.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <AppPage
+      data={data?.app}
+      url={appCompleteUrl}
+      navigateToAbout={() => navigate(appDetailsUrl(id))}
+      onBack={() => navigate(appsListPath)}
+      onError={() =>
+        notify({
+          status: "error",
+          text: intl.formatMessage({
```
bmigirl, line 52: messages file

bmigirl (changes requested):

src/apps/views/App/App.tsx
orzechdev: I didn't change that file, only relocated it. But I'll change it also 👍

schema.graphql
orzechdev: It is copy-paste formatting from the backend. I'll discuss which one is better 👍

src/apps/useExtensions.ts
```diff
@@ -40,35 +62,72 @@ export const mapToMenuItems = (extensions: Extension[]) =>
     onSelect: open
   }));
 
-export const useExtensions = (
-  view: AppExtensionViewEnum,
-  type: AppExtensionTypeEnum
-): Record<Target, Extension[]> => {
+export const mapToExtensionsItems = (
```
jwm0, line 65: What is _extension items_ exactly? The `mapToMenuItems` makes sense because it applied to all menus, but I don't see how this function output make it useful in places other than navigation itself. I think it makes more sense to put specific one-off mappers inside places that actually use them, rather than end up with a list of mappers in one place. If this can be useful for other scenarios, it might make sense to keep it here but I can't think of one.

src/apps/useExtensions.ts
```diff
@@ -40,35 +62,72 @@ export const mapToMenuItems = (extensions: Extension[]) =>
...
+      permissions
+    })
+  );
+  if (items.length) {
+    items.unshift(header);
+  }
+  return items;
+};
+
+export const getMenuItemExtension = (
```
jwm0, line 85: The same goes here

src/apps/views/App/App.tsx
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  const appCompleteUrl = getAppCompleteUrlFromDashboardUrl(
+    location.pathname,
+    data?.app.appUrl,
+    id
+  );
+
+  return (
+    <AppPage
```
jwm0, line 45: It looks like we use same AppPage component for both app itself and configuration, this means we have the same breadcrumb: "App Name / **Configuration**". Change it accordingly

src/apps/index.tsx
```diff
@@ -36,9 +38,13 @@ const AppDetails: React.FC<RouteComponentProps<{ id: string }>> = ({
...
 };
 
-const AppDetailsSettings: React.FC<RouteComponentProps<{ id: string }>> = ({
+const AppSettings: React.FC<RouteComponentProps<{ id: string }>> = ({
   match
-}) => <AppDetailsSettingsView id={decodeURIComponent(match.params.id)} />;
+}) => <AppSettingsView id={decodeURIComponent(match.params.id)} />;
+
+const App: React.FC<RouteComponentProps<{ id: string }>> = ({ match }) => (
+  <AppView id={decodeURIComponent(match.params.id)} />
```
jwm0, line 46: There should be changes inside AppView to only display "Edit settings" button if configurationUrl is provided.

jwm0 (changes requested):

src/apps/views/App/App.tsx
orzechdev: Breadcrumbs are set automatically by the app in `useSettingsBreadcrumbs`

src/apps/index.tsx
orzechdev: Do you mean AppDetailsView? I've changed it in AppDetailsView.

schema.graphql
orzechdev: As discussed, I've added `schema.graphql` to `.prettierignore` to prevent schema from unintended formatting in future and persist it the same as on backend.

src/apps/views/App/App.tsx
jwm0: Still, it needs to be changed to display proper breadcrumb for app main view (not "configuration")

jwm0 (commented):
When navigated to app page from extension, the navigation does not show correct tab selection.
1. Click on Catalog => Extension w/ app page
2. Catalog menu item is shown as active, but it should be App navigation item that's selected (since this view is part of this route)

src/apps/views/App/App.tsx
orzechdev: But where should bbe changed? In app or in dashboard?

jwm0: It's the app that sets the breadcrumb? Can't really see what's the use case here 🤔

orzechdev: App sets breadcrumb via the event
```
if (event.data.type === ExtensionMessageType.BREADCRUMB_SET) {
  setBreadcrumbs(event.data.breadcrumbs);
}
```

jwm0 (approved):

eCzerniak (approved):

PR #1917, opened 2022-03-10
Prevent form submit if country is not selected

Cloud11PL:
I want to merge this change because it prevents form submit if country is not selected.

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/orders/components/OrderCustomerAddressesEditDialog/utils.ts
+++ b/src/orders/components/OrderCustomerAddressesEditDialog/utils.ts
@@ -2,6 +2,7 @@ import { SingleAutocompleteChoiceType } from "@saleor/components/SingleAutocompl
 import {
   AccountErrorFragment,
   AddressFragment,
+  AddressInput,
   AddressTypeEnum,
   Node,
   OrderErrorFragment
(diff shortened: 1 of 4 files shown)


orzechdev (approved):

src/orders/components/OrderCustomerAddressesEditDialog/utils.ts
```diff
@@ -48,6 +49,21 @@ export function validateDefaultAddress<T extends AddressFragment>(
...
+
+interface ShippingAddresses {
+  shippingAddress: AccountErrorFragment[] | AddressInput;
+  billingAddress: AccountErrorFragment[] | AddressInput;
+}
+
+export const hasPreSubmitErrors = (input: ShippingAddresses) =>
+  Object.values(input)
+    .flat()
+    .some(el => (el as AccountErrorFragment).code);
```
jwm0, line 65: It looks like you're having to type cast because `AddressInput ` does not have a `code` property and you might be getting the `property does not exist on type ...`. To solve this you can use a type safe check instead:
```suggestion
    .some(el => "code" in el);
```

This is also a natively built type guard (automatic type assertion in union types), here's a short example on how it can be useful:
```
type A = { foo: string, ok: boolean };
type B = { bar: string  }

const someFn = (x: A | B) => {
  if ("foo" in x) {
    return x.ok; // does not throw type error
  }
}
```

jwm0 (changes requested):

jwm0 (approved):

dominik-zeglen (approved):

PR #1954, opened 2022-03-25
Ignore more than 2 decimal numbers in price input

taniotanio7:
I want to merge this change because it fixes an issue where user could insert prices that would have more than 2 numbers after the decimal point, which would cause error when trying to save any form containing such prices.

### Screenshots

Before:

https://user-images.githubusercontent.com/4144459/160163445-8194c9fb-d31f-4b86-9d1b-73da8c80f346.mp4

After:

https://user-images.githubusercontent.com/4144459/160163458-1ff7dedb-e2c0-4d3c-9a28-3674ffe05744.mp4

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/components/PriceField/PriceField.tsx
+++ b/src/components/PriceField/PriceField.tsx
@@ -1,4 +1,4 @@
-import { InputAdornment, TextField } from "@material-ui/core";
+import { InputAdornment, TextField, TextFieldProps } from "@material-ui/core";
 import { InputProps } from "@material-ui/core/Input";
 import { makeStyles } from "@saleor/macaw-ui";
 import React from "react";
@@ -43,6 +43,8 @@ interface PriceFieldProps {
   onChange(event: any);
 }
 
+const MAX_DECIMAL_LENGTH = 2;
+
 export const PriceField: React.FC<PriceFieldProps> = props => {
   const {
     className,
@@ -61,6 +63,17 @@ export const PriceField: React.FC<PriceFieldProps> = props => {
 
   const classes = useStyles(props);
   const minValue = 0;
+
+  const handleChange: TextFieldProps["onChange"] = e => {
+    const [, decimalPart] = e.target.value.split(".");
+
+    if (decimalPart?.length > MAX_DECIMAL_LENGTH) {
+      return;
+    }
+
+    onChange(e);
+  };
+
   return (
     <TextField
       className={className}
@@ -88,6 +101,7 @@ export const PriceField: React.FC<PriceFieldProps> = props => {
         ),
         inputProps: {
           min: 0,
+          step: "0.01",
           ...InputProps?.inputProps
         },
         type: "number"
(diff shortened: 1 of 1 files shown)


patrys:
Task linked: [SALEOR-6312 All price fields should be 2 decimals](https://app.clickup.com/t/2549495/SALEOR-6312)

Droniu (approved):

Cloud11PL (approved):

taniotanio7:
Updated the PR to handle currencies that have other requirements for decimal point value.
For example: no decimal numbers - like South-Korean Won (KRW) or 3 decimal numbers - like Kuwaiti Dinar (KWD).

The data for each currency is provided to the component as `currencySymbol` prop (ex. `"KRW"`). That symbol is then passed to [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) that has the correct formatting values according to the [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) specification.

Currently, backend doesn't provide any data about currencies other than its symbol.

taniotanio7:
If there would be more requirements for `PriceField` in the future (ex. formatting separators), we could use an existing solution like [react-currency-input-field](https://www.npmjs.com/package/react-currency-input-field)

Cloud11PL:
Opening product "Apple Juice" throws (I think related to this PR) error:
image

taniotanio7:
Ignore any character other than `,` and `.` (for example `e` is problematic because it's used for exponents)

taniotanio7:
The [failing test `cypress-run-critical`](https://github.com/saleor/saleor-dashboard/runs/5752814288?check_suite_focus=true) is fixed by https://github.com/saleor/saleor-dashboard/pull/1957

src/components/PriceField/PriceField.tsx
```diff
@@ -101,7 +129,7 @@ export const PriceField: React.FC<PriceFieldProps> = props => {
         ),
         inputProps: {
           min: 0,
-          step: "0.01",
+          step: 1 / Math.pow(10, maxDecimalLength),
```
Droniu, line 132:
```suggestion
          step: 1 / 10 ** maxDecimalLength,
```
there isn't any implementation differences between these, but imo we should use es7 syntax, as it's cleaner

Droniu (approved):
Currencies with triple decimals and exponent input are crazy edge cases, very interesting PR. Good work

src/components/PriceField/PriceField.tsx
```diff
@@ -61,6 +64,44 @@ export const PriceField: React.FC<PriceFieldProps> = props => {
 
   const classes = useStyles(props);
   const minValue = 0;
+
+  const maxDecimalLength = useMemo(
+    () => getCurrencyDecimalPoints(currencySymbol),
+    [currencySymbol]
+  );
+
+  const handleChange: TextFieldProps["onChange"] = e => {
```
dominik-zeglen, line 73:
```suggestion
  const handleChange: React.ChangeEventHandler = e => {
```

dominik-zeglen: Just to keep consistency

dominik-zeglen (approved):

PR #1960, opened 2022-03-29
Change shipping edit dialog

Cloud11PL:
I want to merge this change because it:
- Changes the order address edit flow
- Makes address edit to be only done in one dialog window as opposed to multiple different ones
- Implements new UI component - Select Switch

### Screenshots

image
image

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/orders/components/OrderAddressFields/OrderAddressFields.tsx
+++ b/src/orders/components/OrderAddressFields/OrderAddressFields.tsx
@@ -26,14 +26,14 @@ interface OrderAddressFieldsProps {
   errors: OrderErrorFragment[];
 }
 
-const getVariant = (action: string) => {
+const getVariant = (action: string): AddressEditDialogVariant => {
   switch (action) {
     case "edit-customer-address":
-      return AddressEditDialogVariant.CHANGE_CUSTOMER_ADDRESS;
+      return "CHANGE_CUSTOMER_ADDRESS";
     case "edit-shipping-address":
-      return AddressEditDialogVariant.CHANGE_SHIPPING_ADDRESS;
+      return "CHANGE_SHIPPING_ADDRESS";
     case "edit-billing-address":
-      return AddressEditDialogVariant.CHANGE_BILLING_ADDRESS;
+      return "CHANGE_BILLING_ADDRESS";
     default:
       return undefined;
   }
@@ -64,9 +64,11 @@ const OrderAddressFields: React.FC<OrderAddressFieldsProps> = ({
     onConfirm
   };
 
+  const isAddressEditDialogVariantSelected = !!getVariant(action);
+
   return (
     <OrderCustomerAddressesEditDialog
-      open={!!getVariant(action)}
+      open={isAddressEditDialogVariantSelected}
       variant={getVariant(action)}
       {...addressFieldCommonProps}
     />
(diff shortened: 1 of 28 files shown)


patrys:
Task linked: [SALEOR-6155 Change dialog of the shipping edit](https://app.clickup.com/t/2549495/SALEOR-6155)

src/orders/components/OrderAddressFields/OrderAddressFields.tsx
```diff
@@ -62,26 +65,11 @@ const OrderAddressFields: React.FC<OrderAddressFieldsProps> = ({
...
-        {...addressFieldCommonProps}
-      />
-      <OrderCustomerAddressesEditDialog
-        open={action === "edit-billing-address"}
-        variant={AddressEditDialogVariant.CHANGE_BILLING_ADDRESS}
-        {...addressFieldCommonProps}
-      />
-    </>
+    <OrderCustomerAddressesEditDialog
+      open={!!getVariant(action)}
```
Droniu, line 69: This isn't easily readable for me, I'd add explicit `const isAddressEditDialogVariantSelected = !!getVariant(action)` or something similar

src/orders/components/OrderCustomerAddressesEditDialog/OrderCustomerAddressesEditDialog.tsx
```diff
@@ -239,11 +176,22 @@ const OrderCustomerAddressesEditDialog: React.FC<OrderCustomerAddressesEditDialo
...
 
+  const getVariant = (action: AddressEditDialogVariant) => {
+    switch (action) {
+      case AddressEditDialogVariant.CHANGE_SHIPPING_ADDRESS:
+        return "shipping";
+      case AddressEditDialogVariant.CHANGE_BILLING_ADDRESS:
+        return "billing";
+    }
+  };
+
```
Droniu, line 187: What about change customer, change customer address variants? I think it would be good idea to define explicit union type as return type for this function (and also in utils).

Right now when I change customer, this falls back to billing address as this is how `getAddressEditProps` handles variants other than shipping. E. g. when I change customer I see a checkbox "Set the same for billing address" which suggests that dialog is in shipping mode. However when i disable this checkbox and change address, billing changes together with shipping. Shipping is first address from array, I think. It has to be investigated further.

src/orders/views/OrderDetails/OrderDraftDetails/index.tsx
```diff
@@ -114,7 +107,7 @@ export const OrderDraftDetails: React.FC<OrderDraftDetailsProps> = ({
     variables: {
       id: order?.user?.id
     },
-    skip: !order?.user?.id || !isAnyAddressEditModalOpen(params.action)
```
Droniu, line 117: Why is it deleted? Do we need addresses on draft view, when modal is not opened?

Droniu (changes requested):

src/orders/views/OrderDetails/OrderDraftDetails/index.tsx
Cloud11PL: Yeah, you're right, after some digging I realised it was unnecessary

PR #1967, opened 2022-04-01
Fix deploying proper version to staging on new tag release

GrzegorzKowalik:

### Pull Request Checklist

### Test environment config

### DIFF
--- a/.github/workflows/deploy-latest-staging.yaml
+++ b/.github/workflows/deploy-latest-staging.yaml
@@ -56,12 +56,6 @@ jobs:
             curl --request GET --url ${{ secrets.VAULT_URL}} --header "Authorization: JWT ${{ secrets.VAULT_JWT }}" | jq -r .token \
           )
           
-          if [[ -n "${{ github.event.inputs.git_ref }}" ]]; then 
-            version="${{ github.event.inputs.git_ref }}"; 
-          else
-            version="$(git describe --tags)"
-          fi
-          
           gh api /repos/saleor/saleor-cloud-deployments/dispatches \
             --input - <<< '{
               "event_type": "open-release-pull-request",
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-latest-staging.yaml
```diff
@@ -55,15 +55,22 @@ jobs:
...
-              "client_payload": {
-                "project": "DASHBOARD",
-                "environment": "SANDBOX",
-                "version": "${{ github.event.inputs.git_ref }}"
+            --input - <<< "{
+              \"event_type\": \"open-release-pull-request\",
+              \"client_payload\": {
+                \"project\": \"DASHBOARD\",
+                \"environment\": \"SANDBOX\",
+                \"version\": \"${version}\"
```
cmiacz, line 71: Use `${{ env.version }}` and do not introduce `\"`

cmiacz (changes requested):

.github/workflows/deploy-latest-staging.yaml
GrzegorzKowalik: This will not work, as the env var is set inside the script. I think github templates values from env context before executing the script

To avoid escaping those quotes inside json, the variable would need to be defined in previous step and saved to `GITHUB_ENV`. No proper place for that in this workflow, I would need to create a separate step just for that to keep everything clean

.github/workflows/deploy-latest-staging.yaml
```diff
@@ -55,13 +55,20 @@ jobs:
           export GITHUB_TOKEN=$( \
             curl --request GET --url ${{ secrets.VAULT_URL}} --header "Authorization: JWT ${{ secrets.VAULT_JWT }}" | jq -r .token \
           )
+          
+          if [[ -n "${{ github.event.inputs.git_ref }}" ]]; then
```
cmiacz, line 59: Do we still need that part?

cmiacz (approved):

cmiacz:
Please remember to cherry-pick it to 3.1 branch

PR #1969, opened 2022-04-01
Use links instead of onClick navigate function

witoszekdev:
I want to merge this change because it replaces imperative `navigate()` function calls in `onClick` with react router's `Link` component that uses `<a>` element. This allows users to open links in new tab, which improves productivity.

## On hold

Waiting on Macaw v0.5.1 release

## Changes

Pleaces where links were added:
- **Lists** - ex. product list (except Plugins, see below)
- **SortableTables** - ex. product variants
- **Sidebar**
- **Buttons** that open new page - ex. "Create product"
- **Backlinks**
- **Menus** - ex. "Account Settings"
- **Links** that actually used onClick - ex. warehouse shipping zone, reset password

Places where links weren't added:
- **Plugins list** - issues with Cloud deployments (works fine if navigated imperatively, but breaks when link is opened in a new tab)
- **Pagination** - will be added in a separate ticket
- **Buttons** that open modal - this action happens on the same page, so link wouldn't be appropriate
- **Modals** that open page (ex. Create variants when no variant is added to product) - those modals use Forms with `onSubmit` event handlers. Adding links instead of `<button type="submit>` doesn't really makes sense here, it could also introduce additional unexpected behavior (ex. we send mutation while redirecting imperatively to that page) and a11y issues

### Screenshots

- **Lists**

https://user-images.githubusercontent.com/4144459/165544455-9304c746-db41-4ec4-a19d-6faf8edd1ea1.mp4

- **SortableTables**

https://user-images.githubusercontent.com/4144459/165544724-e83f128c-bddf-48b6-87c7-a735e8c31b65.mp4

- **Sidebar**

https://user-images.githubusercontent.com/4144459/165544111-1c51ff1c-3af6-4f30-9419-12d909fa770a.mp4

- **Buttons** that open new page

https://user-images.githubusercontent.com/4144459/165544972-b89772da-289a-4780-b853-1934a7f4e361.mp4

- **Backlinks**

https://user-images.githubusercontent.com/4144459/165545156-2c8c1a07-3d7a-4b0d-99d0-e6423c0c6912.mp4

- **Menus**

https://user-images.githubusercontent.com/4144459/165545517-52fcb524-f2d3-4973-944c-54e06aa384c6.mp4

- **Links**

https://user-images.githubusercontent.com/4144459/165545704-71bad75e-db1e-49e3-8026-07e7b5da7ce0.mp4

### Pull Request Checklist

### Test environment config


patrys:
Task linked: [SALEOR-6324 Add links to rows](https://app.clickup.com/t/2549495/SALEOR-6324)

adrian-potepa (approved):

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,16 @@
+import { OverridableComponent } from "@material-ui/core/OverridableComponent";
+import { Button as MacawButton, ButtonTypeMap } from "@saleor/macaw-ui";
+import { isExternalURL } from "@saleor/utils/urls";
+import React from "react";
+import { Link } from "react-router-dom";
+
+const _Button: React.FC<any> = React.forwardRef(({ href, ...props }) => {
```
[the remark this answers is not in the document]
taniotanio7: Changed the name to just `Button`. If I used named export, I would still need these shenanigans because of type assertion.

We're not using named exports in dashboard (I know, I know) but since that's the existing convention I think it's better to stick to what's already there rather than to have different components using named exports and the other ones using default export.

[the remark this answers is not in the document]
taniotanio7: done

jwm0 (approved):
Make sure to bump macaw version before merging this in

src/components/TableRowLink/TableRowLink.tsx
```diff
@@ -0,0 +1,50 @@
...
+import clsx from "classnames";
+import React from "react";
+
+import Link from "../Link";
+
+type MaterialTableRowPropsType = TableRowTypeMap["props"];
+
+export interface TableRowLinkProps
+  extends Omit<MaterialTableRowPropsType, "onClick"> {
+  children?: React.ReactNode;
```
jwm0, line 12: Does it make sense to have this component without any children? I'd make it required

jwm0, line 12: Does it make sense to have this component without any children? I'd make it required

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,16 @@
...
+const _Button: React.FC<any> = React.forwardRef(({ href, ...props }, ref) => {
+  if (href && !isExternalURL(href)) {
+    return <MacawButton {...props} to={href} component={Link} ref={ref} />;
+  }
+
+  return <MacawButton href={href} {...props} ref={ref} />;
+});
+_Button.displayName = "Button";
+
+export const Button = _Button as OverridableComponent<ButtonTypeMap>;
```
jwm0, line 16: What's the reason for having 2 constants here (_Button and Button)? You can do the type assertion directly on the first one, if it's a readability thing and was intended then I'd also change it in SortableTableRow.tsx

witoszekdev: Yes, that was my initial implementation, however this doesn't work with TypeScript:
CleanShot 2022-05-04 at 16 24 13@2x
CleanShot 2022-05-04 at 16 24 05@2x

If we make the assertion in the same constant, then TypeScript infers the type from the component declaration, and it doesn't work nicely with the `OverridableComponent` from Material-UI. We can't use `as unknown as OverridableComponent` as well because this on the other hand breaks types from `React.fowardRef`.

However, if we use this hack, then TypeScript infers the type from our declaration `React.FC<any>` and it allows us to cast this type.

src/components/TableRowLink/TableRowLink.tsx
witoszekdev: Yup, it does make sense. Fixed that

dominik-zeglen (approved):

PR #1989, opened 2022-04-14
Remove store name, store URL and Store description from dashboard settings

taniotanio7:
I want to merge this change because it removes fields Store name, Store URL and Store description that weren't used anywhere (Django leftover). In Cloud instances those fields were uneditable.

**PR intended to be tested with API branch:** main 

### Pull Request Checklist

### Test environment config


patrys:
Task linked: [SALEOR-6040 Changes not saved after updating store name](https://app.clickup.com/t/2549495/SALEOR-6040)

adrian-potepa (approved):

src/siteSettings/components/SiteDetailsSettingsCard/SiteDetailsSettingsCard.tsx
```diff
@@ -59,18 +59,17 @@ const SiteDetailsSettingsCard: React.FC<SiteDetailsSettingsCardProps> = ({
         />
         <FormSpacer />
         <TextField
-          disabled={disabled}
+          disabled={disabled || IS_CLOUD_INSTANCE}
```
jwm0, line 62: Should it be disabled? This looks like a standard use case for a read-only input field - we want you to freely copy and see this field just not be able to edit it directly.

jwm0 (commented):
Was the API updated to reflect these changes? `shopDomainUpdate` mutation was blocked on cloud environments

taniotanio7:
This PR needs API change - after saving we still get PermissionDenied error on cloud instances even when saving just the store name

taniotanio7:
> Was the API updated to reflect these changes? `shopDomainUpdate` mutation was blocked on cloud environments

@jwm0 Yup, you were right - we need to skip the entire mutation to not get the PermissionDenied error. Reverted it back to the initial mutation 👍

src/siteSettings/components/SiteDetailsSettingsCard/SiteDetailsSettingsCard.tsx
taniotanio7: I've asked @timuric if we should make this field disabled. Initialy for me it looked like a bug, the field looked exactly the same as the rest but the input was ignored. I only figured it out that it was read-only when I looke at the code 😆.

Disabled is more clear to me that "nope, you can't edit that". You can still copy the value if you need to.

I guess we can leave it as read-only but it should be more clear to the user - maybe some label ("This field is read-only") or CSS class that makes it look like disabled but allows cursor (so that the user can do Cmd+A and Cmd+C to copy the entire field).

taniotanio7: Nevermind we've decided to remove those fields altogether

jwm0 (approved):

PR #2035, opened 2022-05-09
Fix redirect when modal is opened on apps list

witoszekdev:
I want to merge this change because it fixes an issue with apps list when user clicks on the delete button or "About" button an unexpected redirect to app iframe page happens due to event propagation

### Screenshots
Before
CleanShot 2022-05-09 at 17 42 16@2x

After
CleanShot 2022-05-09 at 17 41 46@2x

### Pull Request Checklist

### Test environment config

### DIFF
--- a/src/apps/components/AppPermissions/AppPermissions.tsx
+++ b/src/apps/components/AppPermissions/AppPermissions.tsx
@@ -1,0 +1,49 @@
+import { AppsListQuery } from "@saleor/graphql";
+import {
+  IconButton,
+  makeStyles,
+  PermissionsIcon,
+  Tooltip
+} from "@saleor/macaw-ui";
+import React from "react";
+import { FormattedMessage } from "react-intl";
+
+const useStyles = makeStyles(
+  theme => ({
+    list: {
+      margin: 0,
+      paddingLeft: "16px"
+    }
+  }),
+  { name: "AppPermissions" }
+);
+
+interface AppPermissionsProps {
+  permissions: AppsListQuery["apps"]["edges"][0]["node"]["permissions"];
+}
+
+export const AppPermissions = ({ permissions }: AppPermissionsProps) => {
+  const classes = useStyles();
+  return (
+    <Tooltip
+      header={
+        <FormattedMessage
+          defaultMessage="App permissions"
+          id="xNfh4L"
+          description="app permissions tooltip header"
+        />
+      }
+      title={
+        <ul className={classes.list}>
+          {permissions.map(permission => (
+            <li key={permission.code}>{permission.name}</li>
+          ))}
+        </ul>
+      }
+    >
+      <IconButton variant="secondary" color="primary">
+        <PermissionsIcon />
+      </IconButton>
+    </Tooltip>
+  );
+};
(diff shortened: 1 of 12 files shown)


patrys:
Task linked: [SALEOR-6729 App list page](https://app.clickup.com/t/2549495/SALEOR-6729)

src/apps/components/AppPermissions/AppPermissions.tsx
```diff
@@ -0,0 +1,49 @@
...
+    list: {
+      margin: 0,
+      paddingLeft: "16px"
+    }
+  }),
+  { name: "AppPermissions" }
+);
+
+interface AppPermissionsProps {
+  permissions: AppsListQuery["apps"]["edges"][0]["node"]["permissions"];
```
jwm0, line 22: Permission should already be a fragment (if it's not, add one) so you can use the generated type from that

src/apps/components/InstalledApps/InstalledApps.tsx
```diff
@@ -22,14 +21,15 @@ import React from "react";
 import { FormattedMessage, useIntl } from "react-intl";
 
 import { useStyles } from "../../styles";
+import { AppPermissions } from "../AppPermissions/AppPermissions";
 import AppsSkeleton from "../AppsSkeleton";
-import DeactivatedText from "../DeactivatedText";
+
+type App = AppsListQuery["apps"]["edges"][0]["node"];
```
jwm0, line 27: Create fragment and its generated type

src/apps/context.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export interface AppListContextValues {
+  activateApp: (appId: string) => void;
+  deactivateApp: (appId: string) => void;
+}
+
+export const AppListContext = React.createContext<AppListContextValues>({
+  activateApp: () => {},
+  deactivateApp: () => {}
+});
```
jwm0, line 12: The `!context` check you're doing inside `useAppListContext` won't work unless you omit the values on context instantiation. 

If you simply omit it here, it will work as expected without any other changes (and you can remove the eslint-disable comment as well!)
```suggestion
export const AppListContext = React.createContext<AppListContextValues | undefined>();
```

src/apps/views/AppsList/AppsList.tsx
```diff
@@ -145,6 +152,46 @@ export const AppsList: React.FC<AppsListProps> = ({ params }) => {
...
+      const errors = data?.appActivate?.errors;
+      if (errors?.length === 0) {
+        notify({
+          status: "success",
+          text: intl.formatMessage(messages.appActivated)
+        });
+        refetch();
+        closeModal();
+      } else {
+        errors.forEach(error =>
```
jwm0, line 166: I could be mistaken here, but don't we have a generic error notifier for these types of errors now? cc @Cloud11PL

Cloud11PL: Yes, I think this is unnecessary - checkout `makeMutation.ts` it should cover most of the cases

src/apps/context.ts
witoszekdev: Thanks, I forgot to delete the default data 😅

orzechdev (approved):

jwm0 (approved):

Cloud11PL (approved):

PR #2128, opened 2022-06-30
Add note about .env file

timuric:

This readme file can be better

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ $ npm i
 
 ### Configuration
 
-Create ```.env``` file in a root directory with the following variables:
+Create ```.env``` file in a root directory or set environment variables with following values:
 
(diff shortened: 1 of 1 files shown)


dominik-zeglen:
I don't know about this, maybe instead or replacing info about env variables with dotenv file we should just add info that we support dotenv?

timuric:
> I don't know about this, maybe instead or replacing info about env variables with dotenv file we should just add info that we support dotenv?

This PR is not replacing info about env variables, it only clarifies where you should paste variables. For the user it doesn't matter much what we use under the hood

Droniu (approved):

timuric:
@dominik-zeglen done

dominik-zeglen (approved):

PR #2299, opened 2022-09-12
Handle form errors before product creation

orzechdev:
It introduces on product creation page:
• Save button is always enabled (also in product update page)
• Display errors in inputs or in toasts when the form is incomplete
• Show leave without saving dialogue if changes failed

There is also added product type picker before opening product creation page.

### Screenshots

Product type picker:
Screenshot 2022-09-12 at 17 09 21
Screenshot 2022-09-13 at 11 14 51

### Pull Request Checklist

### Test environment config

CONTAINERS=1


patrys:
Task linked: [SALEOR-7221 Enable Save button](https://app.clickup.com/t/2549495/SALEOR-7221)

dominik-zeglen (approved):

src/products/utils/validation.ts
```diff
@@ -1,5 +1,44 @@
...
+        code: ProductErrorCode.REQUIRED,
+        field: "name",
+        message: null,
+        attributes: [],
+      },
+    ];
+  }
+
+  return errors;
+};
```
andrzejewsky, line 44: Isn't the same?

```suggestion
const createEmptyRequiredError = (field: string) => ({
    __typename: "ProductError",
     code: ProductErrorCode.REQUIRED,
     field,
     message: null,
     attributes: [],
})

export const validateProductCreateData = (data: ProductCreateData) => {
  if (!data.productType) {
   return [createEmptyRequiredError('productType')]
  }

  if (!data.name) {
      return [createEmptyRequiredError('name')]
  }

  return [];
};
```

Why do you create an empty array (errors) and then merge everything else to it? Do you actually need that?

src/products/views/ProductCreate/ProductCreate.tsx
```diff
@@ -282,6 +288,16 @@ export const ProductCreateView: React.FC<ProductCreateProps> = ({ params }) => {
...
+  const channelsErrors = [
+    ...(updateVariantChannelsOpts.data?.productVariantChannelListingUpdate
+      ?.errors || []),
+    ...(updateChannelsOpts.data?.productChannelListingUpdate?.errors || []),
+  ];
+  const errors = [
+    ...(productCreateOpts.data?.productCreate.errors || []),
+    ...(productVariantCreateOpts.data?.productVariantCreate.errors || []),
+  ];
+
```
andrzejewsky, line 300: can you split this into functions that will extract your errors by a given "opts"? Perhaps it can be useful for later (a utility?) eg. a function that will take an array of opts and return for you merged errors of each entry

andrzejewsky (approved):

src/products/views/ProductCreate/ProductCreate.tsx
orzechdev: Sure, I've found out that we have `getMutationErrors` which I can use here

src/products/utils/validation.ts
orzechdev: Yes, it is the same, Ill change it

PR #2333, opened 2022-09-23
Hide shipping zones when you do not have access

andrzejewsky:
### Pull Request Checklist

### Test environment config

Tests will be re-run only when the "run e2e" label is added.

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation

CONTAINERS=1


patrys:
Task linked: [SALEOR-3162 [Permissions] Channels - user with Channel Management cannot see the list of shipping zones](https://app.clickup.com/t/2549495/SALEOR-3162)

src/channels/components/AssignmentList/styles.ts
```diff
@@ -20,18 +20,30 @@ export const useExpanderStyles = makeStyles(
   { name: "Expander" },
 );
 
-export const useHeaderStyles = makeStyles(
-  theme => ({
+export const useHeaderContainerStyles = makeStyles(
```
dominik-zeglen, line 23: Y tho
It's like creating three different `.css` files (each hook creates separate `<styles/>` tag) so it's slower and seems to me like over the top separation of concerns. There are cases where I'd advice to split stylesheets, eg. reusing the same styles in different components or during component splitting, but having each class in different hooks is not one these cases.

dominik-zeglen (changes requested):

src/channels/components/AssignmentList/styles.ts
andrzejewsky: I did that because when you pass a class key that is not supported by material UI it raises errors in the console. However, you are right here, and to avoid these errors I can just pass the only classes that are needed

dominik-zeglen:
Don't forget to run `npm test -- -u` and see if diff matches expected output.

dominik-zeglen (approved):

PR #2336, opened 2022-09-26
SALEOR-7904: Split payments - Grant refund

witoszekdev:
This PR adds Grant Refund views to order details page.

**What are refund grants?**

They are a business decision that a refund should be given to the user, without actually sending the refund (this will be done in a separate step, PR: https://github.com/saleor/saleor-dashboard/pull/2359).

Grant refunds don't change the total amount of the order, they also don't change payment status. A button to give a grant refund is visible when there's at least 1 transaction in an order.

Grant refund view allows users to select the amount to refund, but the amount can be any arbitrary value provided by the user larger than 0, they don't need to select any lines from the order to grant a refund. Providing a reason is optional.

When grant refund is saved we only save the amount and reason, items selected in grant refund view are not saved. This means that when editing grant refund, none of the items will be selected - only provided amount and reason will be visible.

**What happens after granting refund?**

Granting refund changes `order.totalRemainingGrant` - until a refund isn't actually send, this amount will be visible in the list of granted refunds as unsettled.

## Links

- [Live preview](https://split-payments-grant-refund.dashboard.saleor.rocks/orders/T3JkZXI6MTdkNmM3MjctYzcwZC00MmQ1LWFmNjktYTM2NjIxYzRlODcz)

- [Storybook preview](https://split-payments-grant-refund.dashboard.saleor.rocks/storybook/index.html?path=/story/views-orders-grant-refund-order--grant-refund)

## TODO

- [x] Add grant refund
- [x] Edit grant refund
- [x] Granted refunds list
- [x] Improve loading state of grant refund page
- [x] Refunds stats in order details
- [x] UI fixes
  - [x] don't pre fill refund amount
  - [x] `NaN` when amount is removed from item in order
  - [x] remove warning text in refund card
  - [x] add more spacing in refund card

### Screenshots

Granted refunds list

Grant refund view

### Test environment config

API_URI=https://support-refund-for-transaction.api.saleor.rocks/graphql/

Tests will be re-run only when the "run e2e" label is added.

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [x] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation

CONTAINERS=1


patrys:
Task linked: [SALEOR-7904 Grant refund](https://app.clickup.com/t/2549495/SALEOR-7904)

package.json
```diff
@@ -29,7 +29,7 @@
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@material-ui/styles": "^4.11.4",
     "@reach/auto-id": "^0.16.0",
-    "@saleor/macaw-ui": "git+https://github.com/saleor/macaw-ui.git#transaction-list-improvements",
+    "@saleor/macaw-ui": "git+https://github.com/saleor/macaw-ui.git#main",
```
andrzejewsky, line 32: is that because it's not yet released?

witoszekdev: yes

src/orders/components/OrderGrantRefundPage/reducer.ts
```diff
@@ -0,0 +1,127 @@
...
+      type: "setMaxQuantity";
+      lineIds: string[];
+    }
+  | {
+      type: "initState";
+      state: GrantRefundState;
+    }
+  | {
+      type: "toggleRefundShipping";
+    };
```
andrzejewsky, line 32: love it, so that's actually a discrimination union pattern you can also add an exhaustiveness check (optional)

src/orders/components/OrderPayment/components/PaymentsSummary.tsx
```diff
@@ -0,0 +1,87 @@
...
+import SummaryLine from "../../OrderSummaryCard/SummaryLine";
+import { SummaryList } from "../../OrderSummaryCard/SummaryList";
+import { orderPaymentMessages } from "../messages";
+import { useStyles } from "../styles";
+
+interface PaymentsSummaryProps {
+  order: OrderDetailsFragment;
+}
+
+export const getShouldDisplayAmounts = (order: OrderDetailsFragment) => {
```
andrzejewsky, line 15: idea: how about returning here also a discriminating union where the discriminant property is the name of the state? We can distinguish here a few states (ordering from the top, example names):
- WRONG_ORDER
- DIFFERENT_AMOUNTS
- NOT_AUTHORIZED
- PARTIAL_CAPTURED
- FULLY_PAID

then in the if statements, you can explicitly say in what state you return null (PaymentsSummary component). 

_happy type safe!_

andrzejewsky (approved):

src/orders/components/OrderGrantRefundPage/components/RefundCard.tsx
```diff
@@ -0,0 +1,133 @@
...
+  submitDisabled,
+}: RefundCardProps) => {
+  const intl = useIntl();
+  const { locale } = useLocale();
+  const classes = useRefundCardStyles();
+  const id = useId();
+
+  const { state, dispatch, form, totalSelectedPrice } = useGrantRefundContext();
+
+  const currency = order?.total?.gross?.currency ?? "USD";
```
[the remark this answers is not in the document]
witoszekdev: yup, that doesn't really make sense to use USD as default, it just needs to be a string, so I've replaced the default value (used when data is still fetching) to just an empty string

src/orders/components/OrderPayment/components/PaymentsSummary.tsx
witoszekdev: that's really nice, much more readable than just comments in this function

Droniu (approved):

PR #2472, opened 2022-10-26
Spacing nudges

timuric:

### Pull Request Checklist

### Test environment config

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants

CONTAINERS=1

### DIFF
--- a/src/orders/components/OrderPayment/styles.ts
+++ b/src/orders/components/OrderPayment/styles.ts
@@ -39,6 +39,8 @@ export const useStyles = makeStyles(
     actions: {
       gap: theme.spacing(1),
       display: "flex",
+      flexWrap: "wrap",
+      justifyContent: "right",
     },
   }),
   { name: "OrderPayment" },
--- a/src/orders/components/OrderPayment/styles.ts
+++ b/src/orders/components/OrderPayment/styles.ts
@@ -36,6 +36,11 @@ export const useStyles = makeStyles(
     success: {
       color: theme.palette.success.dark,
     },
+    actions: {
+      "& > *:not(:last-child)": {
+        marginRight: theme.spacing(2),
+      },
+    },
   }),
   { name: "OrderPayment" },
 );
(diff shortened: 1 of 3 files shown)


src/orders/components/OrderPayment/styles.ts
```diff
@@ -36,6 +36,11 @@ export const useStyles = makeStyles(
     success: {
       color: theme.palette.success.dark,
     },
+    actions: {
+      "& > *:not(:last-child)": {
+        marginRight: theme.spacing(2),
+      },
+    },
```
orzechdev, line 43: Wouldn't be the following cleaner?
```suggestion
    actions: {
      gap: theme.spacing(2),
      display: "flex",
      flexWrap: "wrap",
      justifyContent: "right"
    },
```
Gap for flex is not only supported by IE 11, but it is by Edge.

timuric: why is it cleaner? :D

orzechdev: The purpose of `gap` is clear, we want to create a gap between items, while when writing `marginRight` we require from dev to understand it's purpose. With `gap` browser takes care about not appending space at the very end, so if we one day decide to add specific margin to one item, it will not collide with "& > *:not(:last-child)" rule, that rule is implicitly in `gap` property.

Besides that with `flexWrap: "wrap"` we also apply gap between rows which may wrap on small resolutions and justify them to right with `justifyContent: "right"`. Otherwise, we should need to write another margin rule.

orzechdev (dismissed):

andrzejewsky (approved):

Issue #2471, opened 2022-10-26
Improper mixed use of me object and user object


orzechdev:
### What I'm trying to achieve

I'm trying to get user details on the staff user details page while being logged in via an external auth provider.

Log in to the dashboard does at the very beginning me query, which returns userPermissions. It works in both log-in method, either with external provider or with Saleor account. But when being logged in dashboard makes another user query, then it returns none permissions in userPermissions, although me previously returned them - only when logged in with external provider. When logged in with Saleor account, userPermissions are returned correctly, the same as in me query. What is the reason for it?

### Steps to reproduce the problem

1. Log in to the dashboard with an external auth provider.
2. Go to the staff user details page.
3. (fixed in https://github.com/saleor/saleor-dashboard/pull/2478, but requires further work) 404 error occurs due to the missing `MANAGE_STAFF` permission in userPermissions returned by query `user`.

### What I expected to happen

I expected to be able to open staff user details page, without 404 error. Query `user` does not know about user permissions in the external system, only about the ones assigned explicitly in Saleor, what means userPermissions in `user` query might not reflect real permissions when we are logged in with external provider). So,:
~1. we should not query for userPermissions via `user` query, only via `me` query, and...~ (fixed in https://github.com/saleor/saleor-dashboard/pull/2478)
2. we should not treat `user` query object as `me` query object, they should be saved as distinct objects in the Apollo Cache (either in Dashboard and in SDK), so their userPermissions fields will not be overwritten, like it is now (`me` query with userPermissions -> `user` without userPermissions -> 404 error).

### Screenshots

<img width="1426" alt="Screenshot 2022-10-26 at 14 44 12" src="https://user-images.githubusercontent.com/9825562/198032430-852deebc-0292-42ee-a30e-bc2edb42c573.png">

### HTTP request and response from inspector

```js
// Me request:
{
    "operationName": "UserDetails",
    "variables": {},
    "query": "query UserDetails {\n  me {\n    ...User\n    __typename\n  }\n}\n\nfragment User on User {\n  id\n  email\n  firstName\n  lastName\n  isStaff\n  userPermissions {\n    ...UserPermission\n    __typename\n  }\n  avatar {\n    url\n    __typename\n  }\n  __typename\n}\n\nfragment UserPermission on UserPermission {\n  code\n  name\n  __typename\n}\n"
}
// Me response:
{
    "me": {
        "id": "uhHkLusodSkj",
        "email": "john.smith@example.com",
[...]

PR #2478, opened 2022-10-26
Fix account settings page when logged in with external auth

orzechdev:
I want to merge this change because it fixes the 404 crash that occurs on user account page mentioned in https://github.com/saleor/saleor-dashboard/issues/2471 (the issue with not separated `Query.me` cache from `Query.user` cache still needs to be solved).

### Pull Request Checklist

### Test environment config

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants

CONTAINERS=1

### DIFF
--- a/src/staff/components/StaffProperties/StaffProperties.tsx
+++ b/src/staff/components/StaffProperties/StaffProperties.tsx
@@ -1,12 +1,10 @@
 import photoIcon from "@assets/images/photo-icon.svg";
 import { Card, CardContent, TextField, Typography } from "@material-ui/core";
 import CardTitle from "@saleor/components/CardTitle";
-import {
-  StaffErrorFragment,
-  StaffMemberDetailsFragment,
-} from "@saleor/graphql";
+import { StaffErrorFragment } from "@saleor/graphql";
 import { commonMessages } from "@saleor/intl";
 import { makeStyles } from "@saleor/macaw-ui";
+import { StaffMemberDetails } from "@saleor/staff/types";
 import { getFormErrors } from "@saleor/utils/errors";
 import getStaffErrorMessage from "@saleor/utils/errors/staff";
 import React from "react";
@@ -105,7 +103,7 @@ interface StaffPropertiesProps {
   };
   errors: StaffErrorFragment[];
   disabled: boolean;
-  staffMember: StaffMemberDetailsFragment;
+  staffMember: StaffMemberDetails;
   onChange: (event: React.ChangeEvent<any>) => void;
   onImageDelete: () => void;
   onImageUpload: (file: File) => void;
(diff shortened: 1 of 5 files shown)


src/staff/types.ts
```diff
@@ -0,0 +1,8 @@
+import { StaffMemberDetailsFragment } from "@saleor/graphql";
+
+type WithOptional<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
+
+export type StaffMemberDetails = WithOptional<
```
krzysztofzuraw, line 5: shouldn't core make those as optional in schema? 🤔

src/staff/types.ts
```diff
@@ -0,0 +1,8 @@
+import { StaffMemberDetailsFragment } from "@saleor/graphql";
+
+type WithOptional<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
```
krzysztofzuraw, line 3: do we have place for such utils in our codebase? I imagine that we would like to reuse this type in the future 🤔

src/staff/components/StaffProperties/StaffProperties.tsx
```diff
@@ -105,7 +103,7 @@ interface StaffPropertiesProps {
   };
   errors: StaffErrorFragment[];
   disabled: boolean;
-  staffMember: StaffMemberDetailsFragment;
+  staffMember: StaffMemberDetails;
```
andrzejewsky, line 106: `StaffMemberDetails` is a mapped type from `StaffMemberDetailsFragment` that makes two fields (isActive, permissionGroups) optional. `StaffMemberDetailsFragment` is the api-related thing, is actually a contract we rely on, why do you change that contract? Could you cover cases as the API indicates?

andrzejewsky (changes requested):

src/staff/components/StaffProperties/StaffProperties.tsx
orzechdev: Ok, I changed to types from API. It requires checking if fields exist in objects to prevent type errors, but anyway it might be worth doing it gives the advantage of using API types, not custom ones.

src/staff/types.ts
orzechdev: Removed 👌

krzysztofzuraw (approved):

andrzejewsky (changes requested):

andrzejewsky (approved):

krzysztofzuraw (approved):

szczecha (approved):

Issue #2422, opened 2022-10-20
Update dockerfile


andrzejewsky:
### What I'm trying to achieve

Our dockers are a bit outdated, we are able to optimize that and are likely to have just one. Additionaly would be nice to have published docker images, ready to consume

### Describe a proposed solution

- Create one production docker file
- use build stages to achieve tiny image (just built code, nothing else)
- create CI action that will publish docker images to ghcr (make it visible [as packages](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry))
- consider using alpine
- allow the dashboard to switch API URL without rebuilding the image

Refs: https://github.com/saleor/saleor-app-template/issues/95

andrzejewsky:
quick testing, seems like this dockerfile will work fine:

```dockerfile
# 1. dependencies
FROM node:16-alpine AS deps
RUN apk add --no-cache libc6-compat
WORKDIR /app
COPY package.json package-lock.json* ./
RUN npm ci

# 2. build
FROM node:16-alpine AS builder
WORKDIR /app
COPY --from=deps /app/node_modules ./node_modules
COPY . .
RUN yarn build


# 3. runner
FROM nginx:stable-alpine AS runner
WORKDIR /app

ENV NODE_ENV=production

COPY ./default.conf /etc/nginx/conf.d/default.conf
COPY --from=builder /app/build/ /app/
```

@orzechdev please have a look

tofran:
@andrzejewsky I would merge the step `1. dependencies` with `2. build` as suggested on #2281
Splitting these seems that it is just creating additional overhead with no gain. Both images are `node:16-alpine` that will be thrown away, so having it in the same build step will allow for cleaner, easier to understand code and possibly slightly faster builds.

tofran:
I am really interesting in this:

> allow the dashboard to switch API URL without rebuilding the image

What the current idea to design this? Is it already discussed somewhere other than #2271?

andrzejewsky:
@tofran so the idea is similar to yours, we also will _sed'ing_ the codebase, but not the js files - just index.html.

We will put our "config" within the window object, which will be initialized in the index.html, now we can _sed_ this based on the variable name. Modifying index.html in the runtime is safer, we don't touch the code itself which is a bit risky.

tofran:
> Modifying index.html in the runtime is safer

Yup I agree, that's much nicer. (Obviously requires code changes that I didn't intend to do initially)

[...]

PR #2516, opened 2022-11-03
Update dockefile with dynamic API_URI

krzysztofzuraw:
I want to merge this change because it allows setting dynamic `API_URI` from docker.

## How to test it?
```shell
docker build --tag saleor-dashboard .
docker run --publish 8080:80 --env "API_URL=https://automation-dashboard.staging.saleor.cloud/graphql/" saleor-dashboard
open http://localhost:8080
```

Issue linked: #2422.

### Pull Request Checklist

### Test environment config

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants

CONTAINERS=1


Dockerfile
```diff
@@ -1,19 +1,26 @@
...
-RUN STATIC_URL=${STATIC_URL} API_URI=${API_URI} MARKETPLACE_URL=${MARKETPLACE_URL} SALEOR_APPS_ENDPOINT=${SALEOR_APPS_ENDPOINT} APP_MOUNT_URI=${APP_MOUNT_URI} npm run build
 
-FROM nginx:stable
+RUN npm run build
+
+FROM nginx:stable-alpine as runner
 WORKDIR /app
+
 COPY ./nginx/default.conf /etc/nginx/conf.d/default.conf
+COPY ./nginx/replace-api-url.sh /docker-entrypoint.d/replace-api-url.sh
```
tofran, line 25: Please follow the convention, filenames in `/docker-entrypoint.d` should start with a two digit number. Ex: `01-something.sh`

As suggested in https://github.com/tofran/saleor-dashboard/pull/3, please do:
`COPY ./nginx/replace-api-url.sh /docker-entrypoint.d/50-replace-api-url.sh`

I choose 50 since it runs after the built-in scripts and is flexible for expandability (end-users of your image).

nginx/replace-api-url.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env sh
```
tofran, line 1: This is debatable, but i would avoid this shebang. It makes it look that this is compatible with POSIX `sh`.
But in reality (ln 10), we have `[[ ]]` that is undefined: https://mywiki.wooledge.org/BashFAQ/031

That's why it is safer to have the following `#!/usr/bin/env bash`

Dockerfile
krzysztofzuraw: 👋🏻 Hi @tofran and thank you for your comment. Can you send me a link to a convention about 2-digit numbers? So I will know for the future 😁

nginx/replace-api-url.sh
krzysztofzuraw: We have Alpine Linux as base docker images - there is no `bash` there, so I've used `sh` instead.

andrzejewsky (approved):

nginx/replace-api-url.sh
tofran: Okay makes sense. Then one should do `if [ -z "${API_URL}" ]; then` instead of using `[[ ]]`

orzechdev (approved):

Dockerfile
tofran: First, you can see that in the `/docker-entrypoint.d` directory they already have a few scripts that follow this pattern:

```
$ docker run -it nginx:stable-alpine sh
/ # ls docker-entrypoint.d/
10-listen-on-ipv6-by-default.sh 
30-tune-worker-processes.sh
20-envsubst-on-templates.sh
```

Then if you have a look into their `docker-entrypoint.sh`, you will see that they intentionally sort scripts: 
```
find "/docker-entrypoint.d/" -follow -type f -print | sort -V | while read -r f; do
```

This is a way to control the execution flow if entry-point scrips. Alone should be enough, we should respect the upstream, but in Linux this scripting ordering is quite common.

PR #2523, opened 2022-11-03
Update Dockerfile

krzysztofzuraw:

### Pull Request Checklist

### Test environment config

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants

CONTAINERS=1


.github/workflows/publish-containers.yml
```diff
@@ -0,0 +1,113 @@
...
+            NAMED_VERSION="snapshot"
+          fi
+          TAGS=$"\
+          ${{ steps.image.outputs.image_name }}:${NAMED_VERSION},\
+          ${{ steps.image.outputs.image_name }}:${VERSION}\
+          "
+          echo "
+          CONTAINER_TAGS=${TAGS}
+          NAMED_VERSION=${NAMED_VERSION}
+          " >> "${GITHUB_ENV}"
```
NyanKiyoshi, line 50: This workflow is copying the current known issues with core's CI

## Issue no.1
This assumes there is only one minor version maintained at a time, it's not true. I would recommend not to tag `latest` and not to tag `snapshot`, only tag the version

## Issue no.2
It doesn't support minor tags, only patch tags. When we push `1.2.3`, we should push `1.2` as well.

.github/workflows/publish-containers.yml
```diff
@@ -0,0 +1,113 @@
...
+          TAGS=$"\
+          ${{ steps.image.outputs.image_name }}:unstable-${CLEAN_BRANCH_NAME}
+          "
+          # Set version name for open-containers version label as:
+          # <branch>-<drift>-g<6-digit-hash>
+          NAMED_VERSION=$(git describe --all --long | tr -d $'\n')
+          echo "
+          CONTAINER_TAGS=${TAGS}
+          NAMED_VERSION=${NAMED_VERSION}
+          " >> "${GITHUB_ENV}"
```
NyanKiyoshi, line 68: Do we want to push `unstable-main` as an image? There were comments from @patrys against doing this

.github/workflows/publish-containers.yml
```diff
@@ -0,0 +1,113 @@
...
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Cache Docker layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-${{ hashFiles('requirements_dev.txt') }}
```
NyanKiyoshi, line 85: This is node.js a project, not a Python project, thus point to `package-lock.json`

Dockerfile
```diff
@@ -1,19 +1,26 @@
-FROM node:14 as builder
+FROM node:14-alpine as builder
 WORKDIR /app
 COPY package*.json ./
-RUN npm install
+RUN npm ci
 COPY . .
```
NyanKiyoshi, line 5: Only copy what you actually need

NyanKiyoshi (changes requested):

.github/workflows/publish-containers.yml
```diff
@@ -0,0 +1,113 @@
...
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ secrets.GHCR_USERNAME }}
+          password: ${{ secrets.GHCR_PASSWORD }}
```
NyanKiyoshi, line 92: That's a really outdated and insecure way of authenticating to GHCR 
```suggestion
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
```

And add to the job definitions:
```
    permissions:
      contents: read
      packages: write
```

NyanKiyoshi (changes requested):

Droniu (approved):

.github/workflows/publish-containers.yml
```diff
@@ -0,0 +1,73 @@
...
+          DASHBOARD_VERSION=${GITHUB_REF#refs/tags/}
+          echo "
+          DASHBOARD_VERSION=${DASHBOARD_VERSION}
+          CONTAINER_TAGS=${{ steps.image.outputs.image_name }}:${DASHBOARD_VERSION}
+          " >> "${GITHUB_ENV}"
+
+      - name: Set up Docker QEMU
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: all
```
NyanKiyoshi, line 42: Only ARM64 is needed:
```suggestion
          platforms: arm64
```

Dockerfile
```diff
@@ -1,19 +1,47 @@
...
-RUN npm install
-COPY . .
-ARG APP_MOUNT_URI
+RUN npm ci
+
+
+COPY nginx/ nginx/
+COPY assets/ assets/
+COPY locale/ locale/
+COPY testUtils testUtils/
```
NyanKiyoshi, line 10: Is `testUtils` actually needed? Sounds like it's something about a test suite?

.github/workflows/publish-containers.yml
```diff
@@ -0,0 +1,73 @@
...
+        with:
+          builder: ${{ steps.buildx.outputs.name }}
+          context: ./
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: ${{ env.CONTAINER_TAGS }}
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache
+          build-args: |
+            COMMIT_ID=${{ github.sha }}
```
NyanKiyoshi, line 67: `Dockerfile` defines this:
```
org.opencontainers.image.version="$PROJECT_VERSION" \
```

But it wasn't getting passed:
```suggestion
            COMMIT_ID=${{ github.sha }}
            PROJECT_VERSION=$DASHBOARD_VERSION
```

Dockerfile
```diff
@@ -1,19 +1,47 @@
...
+FROM node:14-alpine as builder
 WORKDIR /app
 COPY package*.json ./
-RUN npm install
-COPY . .
-ARG APP_MOUNT_URI
+RUN npm ci
+
+
+COPY nginx/ nginx/
```
NyanKiyoshi, line 7:
```suggestion
RUN npm ci

COPY nginx/ nginx/
```

Dockerfile
krzysztofzuraw: It is required to build Docker - without `testUtils` Docker build yields an error:
```#24 78.95 TS2307: Cannot find module '../../testUtils/ApolloMockedProvider' or its corresponding type declarations.```

Comment (patrys) [reply Dockerfile:9]: Maybe it's worth investigating why production code tries to import something designated for testing.

Comment (andrzejewsky) [reply Dockerfile:9]: yeah, let's investigate that, mocks should not be mandatory for building..

**Comment by andrzejewsky** [comment]
@krzysztofzuraw  takeaway: please create an issue to move cypress to dev deps, I don't know why it's placed there, but worth investigating before we take a step.

**Comment by krzysztofzuraw** [comment]
@andrzejewsky issue created at https://github.com/saleor/saleor-dashboard/issues/2575

**Review by NyanKiyoshi — APPROVED** [review]

**Review by orzechdev — APPROVED** [review]

File: Dockerfile (hunk 7 above)
Comment (andrzejewsky) [reply Dockerfile:9]: sure let's take this as a next thing and close the docker/building topic

File: README.md [hunk 10]
```diff
@@ -145,6 +145,22 @@ You are ready to run cypress commands like:
...
+```
+
+Run nginx from docker and bind it to port on your machine (in this example 8080):
+
+```shell
+docker run --publish 8080:80 --env "API_URL=<YOUR_API_URL>" saleor-dashboard
+```
+
+Enter `http://localhost:8080/` to use dashboard.
+
```
andrzejewsky, line 163: would be nice to attach also a command for replacing API URL within runtime, I think it will be something like:

```
docker exec -it -e API_URL=some url [container id] /replace-api-url.sh
```

nginx/browser will cache index.html so I think the user will see that the codebase has changed (notification on the dashboard side), but after refreshing (clicking reload on the notification) the API will be changed - would be nice also to improve that afterward (change API applies instantly)

andrzejewsky (approved):

Issue #2553, opened 2022-11-14
Lack of scroll bar in media picker


andrzejewsky:
### What I'm trying to achieve

Add media scroll-bar for media picker

### Steps to reproduce the problem

1. Add images for a product until you overflow the picker section

### What I expected to happen

I am supposed to see a scroll bar and be able to navigate over the images

### System information

Dashboard version: 3.8

Core version: 3.8

Operating system:

- [ ] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Other

Browser:

- [ ] Safari
- [ ] Chrome
- [ ] Edge
- [ ] Other

karolm-saleor:
Steps to reproduce the problem are not clear

karolm-saleor:
(/) QA Passed

PR #2625, opened 2022-11-21
Fix scroll bar in media picker

orzechdev:
I want to merge this change because... it fixes https://github.com/saleor/saleor-dashboard/issues/2553

### Pull Request Checklist

### Test environment config

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants
23. [ ] payments

CONTAINERS=1


src/storybook/config.js
```diff
@@ -2,6 +2,10 @@
 import requireContext from "require-context.macro";
 import { configure } from "@storybook/react";
 
+window.__SALEOR_CONFIG__ = {
```
krzysztofzuraw, line 5: I think we should fix case that `__SALEOR_CONFIG__` can be undefined in `webpack.d.ts` and add fix inside this function: https://github.com/saleor/saleor-dashboard/blob/3247131570213dc246385b382ee337f20bb27a5c/src/config.ts#L7

krzysztofzuraw (changes requested):

src/storybook/webpack.d.ts
```diff
@@ -0,0 +1,5 @@
+declare interface Window {
+  __SALEOR_CONFIG__: {
```
krzysztofzuraw, line 2:
```suggestion
  __SALEOR_CONFIG__?: {
   APP_MOUNT_URI: string;
  };
}
```

src/storybook/config.js
andrzejewsky: well.... storybook should not even touch API endpoints but we will manage that later

src/storybook/webpack.d.ts
andrzejewsky: why optional? can you run the APP without config?

andrzejewsky (approved):

src/storybook/config.js
krzysztofzuraw: if we provide default values for storybook we should just have:
```js
window.__SALEOR_CONFIG__ = {
	APP_MOUNT_URI: "/"
}
```

krzysztofzuraw: and if we want to make sure to fallback when there is no `__SALEOR_CONFIG__` present we should fix function in `config.ts`

src/storybook/webpack.d.ts
krzysztofzuraw: maybe it is not need to have this file at all if it is only for storybook

orzechdev: Actually it is not needed for more then storybook and even the file where we define config is in javascript, as well as storybook version is quite old. I removed this file.

andrzejewsky (approved):

krzysztofzuraw (approved):

Issue #2904, opened 2022-12-21
[Fix] Expected to find element: [name="countryArea"]


szczecha:
We need to update cypress tests due to: [ Implement dropdown for province/country area selection based on allowedFields from the API #2833
](https://github.com/saleor/saleor-dashboard/pull/2833)

<img width="1127" alt="Screenshot 2022-12-21 at 17 02 03" src="https://user-images.githubusercontent.com/30683248/208951628-97db32f0-3a0c-4cc0-846f-7afde420688a.png">

Update data test id 
Update command fillUpBasicAddress
Update fixtures

PR #2905, opened 2022-12-21
E2E fixes after countryArea changes

szczecha:
I want to merge this change because it fixes the tests where we add or update addresses
Fixes: https://github.com/saleor/saleor-dashboard/issues/2904

### Pull Request Checklist

### Test environment config

1. [x] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants
23. [ ] payments

CONTAINERS=4

### DIFF
--- a/cypress/fixtures/addresses.json
+++ b/cypress/fixtures/addresses.json
@@ -23,6 +23,7 @@
     "postalCode": "70957",
     "country": "US",
     "countryArea": "LA",
+    "countryAreaFullName": "Louisiana",
     "phone": "+12025550189",
     "currency": "USD",
     "countryFullName": "United States of America"
@@ -37,6 +38,7 @@
     "postalCode": "37745",
     "country": "US",
     "countryArea": "TN",
+    "countryAreaFullName": "Tennessee",
     "phone": "+12025550169",
     "currency": "USD",
     "countryFullName": "United States of America"
--- a/cypress/elements/shared/addressForm.js
+++ b/cypress/elements/shared/addressForm.js
@@ -8,6 +8,6 @@ export const ADDRESS_SELECTORS = {
   city: '[name="city"]',
   postalCode: '[name="postalCode"]',
   country: '[data-test-id="address-edit-country-select-field"]',
-  countryArea: '[name="countryArea"]',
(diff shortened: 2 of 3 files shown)


krzysztofzuraw (approved):

andrzejewsky (approved):

cypress/fixtures/addresses.json
```diff
@@ -37,6 +38,7 @@
     "postalCode": "37745",
     "country": "US",
     "countryArea": "TN",
+    "countryAreaFullName": "Tennessee",
```
karola312, line 41: Why adding new filed "countryAreaFullName", not just change "countryArea" value from "TN", to "Tennessee"? In these way, we don't have to change anything in fillUpBasicAddress and in expectCorrectBasicAddress commands

szczecha: Right now country area is a select type so I need to make changes in fillUpBasicAddress and use fillAutocompleteSelect there.

I was thinking about only changing the country area from code like "TN" to full name "Tennessee" but when I ran tests I noticed that it fails on the assertion for the country area - the backend returns the country area as code (This is not a bug, it's just data normalization on BE)
image

karola312: In customer it work in other way. The backend returns the country area as full name. So tests for adding address in customer are not working after this changes. I think API should return the same value type in all places

karola312: In customer it work in other way. The backend returns the country area as full name. So tests for adding address in customer are not working after this changes. I think API should return the same value type in all places

szczecha: You're right, I asked the backend team yesterday about this case and they confirmed that they should return the code here as in other mutations with the address: https://github.com/saleor/saleor/issues/11572

PR #2933, opened 2023-01-02
Update dashboard OrderDetails after changes in core

witoszekdev:
I want to merge this change because it updates dashboard OrderDetails view after changes in core (added & deprecated fields)

## TODO

- [x] Add `user`/`app` to Transactions' events list
- [x] Test if mapping of old payments API to transactions list is correct
  - Note: checked for dummy payments only, we still need to test Adyen, Stripe plugins
- [x] Display pending amounts (pendingRefund, pendingCapture) in `TransactionCard`
  - [x] Add total cancelled
- [x] Rewrite mutations for adding manual refunds
  - Add additional fields for pspReferecene (should be optional) and description (event's `message`)
  - Send amounts
- [x] Remove usage of `transactionEvent` on `transactionCreate`
- [x] Remove `TransactionEvent.name` -> `message`
  - Add as a tooltip on hover
- [x] Add link from `TransactionEvent.externalUrl` to the Transaction's events list
  - [x] Always display copy button
  - [x] Show link when externalUrl exists on event
- [x] Rewrite mapping for Transaction event type + name to use new `TransactionEvent.type` - `TransactionEventTypeEnum`
- [x] Use `createdBy` from `TransactionEvent` instead of `app` or `user` fields
- [x] Remove deprecated input fields from `orderFulfillmentReturnProducts` mutation

### Pull Request Checklist

### Test environment config

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants
23. [ ] payments

CONTAINERS=1


src/orders/components/OrderAddTransaction/OrderAddTransaction.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import React from "react";
+
+import OrderAddTransaction from "./OrderAddTransaction";
+
+describe("OrderAddTransaction", () => {
+  it("renders skeleton when order is loading", () => {
+    render(
+      <Wrapper>
```
Droniu, line 12: **issue**: Using Wrapper makes this test non-isolated which defies the purpose of unit test. Values required by contexts should be mocked (example for reference: `ExitFormDialog.test.tsx`). This happens in multiple tests in this PR.

src/orders/components/OrderPaymentSummaryCard/components/PaymentsSummary/types.ts
```diff
@@ -0,0 +1,8 @@
+export enum PaymentState {
+  NO_DATA,
+  AMOUNTS_MISMATCH,
+  PARTIAL_CAPTURE,
+  PARTIAL_AUTHORIZED,
+  FULLY_SETTLED,
+  IS_PENDING,
+}
```
Droniu, line 8: **question:** shouldn't this come from the api and be in generated types?

src/orders/types.ts
```diff
@@ -0,0 +1,40 @@
...
+  "type" | "__typename"
+> & {
+  __typename: "TransactionFakeEvent";
+  mappedResult: TransactionMappingResult;
+};
+
+export type FakeTransaction = Omit<
+  TransactionItemFragment,
+  "events" | "__typename"
+> & { __typename: "FakeTransaction" };
```
Droniu, line 40: **question**: what's the reasoning behind splitting fake and real transaction events?

Droniu (approved):
overall it looks pretty good

src/orders/components/OrderManualTransactionForm/components/PspReferenceField.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const {
+    submitState,
+    pspReference,
+    handleChangePspReference,
+  } = useManualTransactionContext();
+
+  return (
+    <TextField
+      variant="outlined"
+      {...props}
```
poulch, line 19: suggestion: IMO spread of props should be at the beginning or a the end

src/orders/components/OrderPaymentSummaryCard/components/PaymentsSummary/PaymentsSummary.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export const PaymentsSummary: React.FC<PaymentsSummaryProps> = ({ order }) => {
+  const classes = useStyles();
+  const shouldDisplay = getShouldDisplayAmounts(order);
+
+  if (
+    [PaymentState.FULLY_SETTLED, PaymentState.NO_DATA].includes(
+      shouldDisplay.state,
+    )
```
poulch, line 25: suggestion: I'd wrap this into function

src/orders/components/OrderPaymentSummaryCard/components/PaymentsSummary/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+interface ShouldDisplayResult {
+  state: PaymentState;
+  authorized: boolean;
+  captured: boolean;
+  cancelled: boolean;
+  pending: boolean;
+}
+
+export const getShouldDisplayAmounts = (
+  order: OrderDetailsFragment,
```
poulch, line 14: question: should be order undefined or   order: OrderDetailsFragment ?

src/orders/components/OrderTransaction/OrderTransaction.tsx
```diff
@@ -19,14 +21,27 @@ export interface OrderTransactionProps {
   cardFooter?: React.ReactNode;
 }
 
+const getEvents = (
```
poulch, line 24: suggestion: shouldn't be move to utils ?

src/orders/components/OrderTransaction/components/TransactionEvents/TransactionEvents.tsx
```diff
@@ -31,93 +26,57 @@ const useStyles = makeStyles(
...
   }),
   {
     name: "OrderTransactionEvents",
   },
 );
 
+const isFakeEventsList = (
+  events: TransactionEventFragment[] | TransactionFakeEvent[],
+): events is TransactionFakeEvent[] =>
+  events[0]?.__typename === "TransactionFakeEvent";
```
poulch, line 41: question: shouldn't you check if you have at least one element in array ?

src/orders/components/OrderTransaction/components/TransactionEvents/components/EventCreatedBy.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+    expect(screen.queryByRole("link")).not.toBeInTheDocument();
+  });
+
+  it("displays a link to the app if app is passed", () => {
+    render(
+      <MemoryRouter>
+        <Wrapper>
+          <EventCreatedBy createdBy={appAvatar} />
+        </Wrapper>
+      </MemoryRouter>,
```
poulch, line 24: suggestion: try to test this component in solution and mock everything

src/orders/components/OrderTransaction/components/TransactionEvents/components/EventItem.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { EventItem } from "./EventItem";
+
+describe("EventItem", () => {
+  it("displays correct event data", () => {
+    render(
+      <MemoryRouter>
+        <Wrapper>
+          <EventItem
+            event={transactionEvent}
+            onHover={() => undefined}
```
poulch, line 17: suggestion: use jest.fn()

src/orders/components/OrderTransaction/components/TransactionEvents/components/EventItem.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+      minute: "2-digit",
+    });
+
+    expect(row).toHaveTextContent("Success"); // Transaction event
+    expect(row).toHaveTextContent("Capture");
+    expect(row).toHaveTextContent(transactionEvent.amount.amount.toString());
+    expect(row).toHaveTextContent(transactionEvent.amount.currency);
+    expect(row).toHaveTextContent(transactionEvent.pspReference);
+    expect(row).toHaveTextContent(
+      intl.format(new Date(transactionEvent.createdAt)),
```
poulch, line 40: suggestion: IMO you should put here raw string not use extra function to calculate result

src/orders/components/OrderTransaction/utils.ts
```diff
@@ -16,3 +23,121 @@ export const mapActionToMessage: Record<
   CANCEL: transactionActionMessages.cancel,
   REFUND: null,
 };
+
+export const mapTransactionEvent = (
+  event: TransactionEventFragment | TransactionFakeEvent,
```
poulch, line 28: suggestion: event should be undefined | TransactionEventFragment | TransactionFakeEvent

src/orders/components/OrderTransaction/utils.ts
```diff
@@ -16,3 +23,121 @@ export const mapActionToMessage: Record<
...
+      type: null,
+      status: null,
+    };
+  }
+
+  if (event.__typename === "TransactionFakeEvent") {
+    return event.mappedResult;
+  }
+
+  switch (event.type) {
```
poulch, line 41: suggestion: istead of having this long switch, IMO simple map object will solve problem

poulch (commented):
Replace all `@saleor` import with `@dashboard`

src/orders/components/OrderPaymentSummaryCard/components/PaymentsSummary/types.ts
witoszekdev: Nope, this doesn't have to do anything with the API (except with similar names I guess). It's a description for transaction amount states (e.g. **amounts mismatch** - it doesn't exist in the API, but we need to know that the amounts are different, because in such case we want to display them; other example: **fully settled** - it exists in the API, but in our case it just means we don't want to display the amounts not that the order is paid)

src/orders/components/OrderAddTransaction/OrderAddTransaction.test.tsx
witoszekdev: I would argue the opposite: our components have to be used within a context provider, so mocking import of the hook instead of creating a fake context provider gives us less confidence in our tests.

In the `Wrapper` we have context providers for all needed use cases:
- fake apollo provider
- intl provider with only 1 locale: EN
- locale provider with locale set to EN
- date provider, so that date doesn't change between test runs
- timezone provider, same as above
- theme provider - without this, our components simply won't render, because they expect a theme from macaw

src/orders/types.ts
witoszekdev: Fake transaction and events are used for old Payments API and Gift cards API to look exactly like transactions on the order details page.

Obviously, they are not a real transaction, so we need to map data from Payments and Gift cards into Transaction shape. I wanted more flexibility for that mapping - TransactionEvents require many fields (e.g. amount) that those APIs simply don't have. It's not an issue - our component is made so that it can display those events even when some data is missing.

Another reason is that TransactionEvent `type` enum field (e.g. AUTHORIZATION_SUCCESS) holds two pieces of information that we need to get in order to display them: status (e.g. FAILED / SUCCEEDED) and kind of event (e.g. CHARGE / AUTHORIZATION). In fake events, we only pass those mapped result (e.g. FAILED + CHARGE) because some events don't have a 1:1 mapping to `TransactionEvent.type`

src/orders/components/OrderPaymentSummaryCard/components/PaymentsSummary/utils.ts
witoszekdev: yes, there should be `undefined` in that type, thx

src/orders/components/OrderTransaction/components/TransactionEvents/TransactionEvents.tsx
witoszekdev: nope, this is enough - if the array is empty `events[0]` returns undefined and `undefined?.__typename` is also undefined and this definitely doesn't equal `TransactionFakeEvent` string ;)

src/orders/components/OrderTransaction/components/TransactionEvents/components/EventCreatedBy.test.tsx
witoszekdev: To be frank I don't see how it benefits our tests in any meaningful way - in the app we wrap the components in those providers from `Wrapper` anyway, just with real data + it sets the data to be always the one specified, so it doesn't differ between test runs. I think this is against the main philosophy of react-testing-library:

> "The more your tests resemble the way your software is used, the more confidence they can give you." [from KCD](https://twitter.com/kentcdodds/status/977018512689455106)

If we mock everything, then what's the point of writing tests?

We need `MemoryRouter` because otherwise we wouldn't be able to render `Link` component from react-router. Mocking `Link` instead of using this fake router fails to actually do any meaningful test - we want to test the real component if it uses correct `href`.

src/orders/types.ts
witoszekdev: I've added comments in code to explain better this separation 😅

src/orders/components/OrderTransaction/components/TransactionEvents/TransactionEvents.tsx
witoszekdev: This is used to check the evetns array + as a custom type guard, if the array is empty this doesn't change anything because in such a case we won't render `EventItem`

src/orders/components/OrderTransaction/OrderTransaction.tsx
```diff
@@ -21,26 +24,36 @@ export interface OrderTransactionProps {
...
+        amounts={{
+          authorizedAmount: transaction.authorizedAmount,
+          authorizePendingAmount: transaction.authorizePendingAmount,
+          chargedAmount: transaction.chargedAmount,
+          chargePendingAmount: transaction.chargePendingAmount,
+          refundedAmount: transaction.refundedAmount,
+          refundPendingAmount: transaction.refundPendingAmount,
+          canceledAmount: transaction.canceledAmount,
+          cancelPendingAmount: transaction.cancelPendingAmount,
+        }}
```
poulch, line 52: suggestion: can this be reduce to 
```
    amounts={{...transaction}}
```

src/orders/components/OrderTransaction/components/TransactionEvents/components/EventCreatedBy.test.tsx
poulch: In this case, you have to use Memory Router, sorry for confusion, but you can get rid of Wrapper as the test passed without it.

src/orders/components/OrderTransaction/components/TransactionEvents/components/EventStatus.tsx
```diff
@@ -1,41 +1,43 @@
...
           label={intl.formatMessage(statusMessages.failure)}
         />
       );
-    case TransactionEventStatus.REQUEST:
+    case "REQUEST":
       return (
         <Pill color="info" label={intl.formatMessage(statusMessages.request)} />
       );
+    default:
+      return null;
```
poulch, line 41: suggestion: you can add exustivnes check at the end of switch to check if all item from union were checked
```
 const _exhaustiveCheck: never = status;
      return _exhaustiveCheck;
```

poulch (approved):
LGTM great job 🚀 
Remember to fix import aliases from `@saleor` to `@dashboard`

src/orders/components/OrderAddTransaction/OrderAddTransaction.test.tsx
andrzejewsky: That's no longer a unit test, but rather an integration/functional. You should break down your transaction component into smaller pieces and test them separately by unit tests. Instead, you are wrapping the entire functionality, providing many contexts just for testing your functionality as one piece.

It's ok, you can keep that because you put so much effort into this, but... think about it: you don't use everything from your wrapper, you use only specifying things, you rely only on certain fields from given context, and so on, and you should test your components against them first (interface segregation), and if that works - then integration/functional test can add some value.

Whatever breaks now either in your code or these wrappers, it will be difficult to find where the bug comes from your code, or that wrapper - as you tightly couple it

Droniu (approved):

PR #2940, opened 2023-01-03
Add redirect to GraphiQL from product & order

zaiste:
I want to merge this change because it adds an ability to open a GraphiQL playground from the product & order details pages along with the automatically filled query and object ID.

**PR intended to be tested with API branch:** `main`

### Screenshots

| Before | After |
| --- | --- |
| CleanShot 2023-01-09 at 16 27 03@2x | CleanShot 2023-01-09 at 16 26 46@2x |

### Pull Request Checklist

### Test environment config

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [x] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants
23. [ ] payments

CONTAINERS=1

### DIFF
--- a/src/orders/components/OrderDetailsPage/OrderDetailsPage.tsx
+++ b/src/orders/components/OrderDetailsPage/OrderDetailsPage.tsx
@@ -25,12 +25,13 @@ import useNavigator from "@dashboard/hooks/useNavigator";
 import { sectionNames } from "@dashboard/intl";
 import OrderChannelSectionCard from "@dashboard/orders/components/OrderChannelSectionCard";
 import { orderListUrl } from "@dashboard/orders/urls";
+import { encodeGraphQLStatement } from "@dashboard/utils/graphql";
 import { mapMetadataItemToInput } from "@dashboard/utils/maps";
 import useMetadataChangeTrigger from "@dashboard/utils/metadata/useMetadataChangeTrigger";
 import { Typography } from "@material-ui/core";
-import { ConfirmButtonTransitionState } from "@saleor/macaw-ui";
+import { Button, ConfirmButtonTransitionState } from "@saleor/macaw-ui";
 import React from "react";
-import { useIntl } from "react-intl";
+import { FormattedMessage, useIntl } from "react-intl";
 
 import { getMutationErrors, maybe } from "../../../misc";
 import OrderCustomer from "../OrderCustomer";
@@ -46,6 +47,22 @@ import { messages } from "./messages";
 import { useStyles } from "./styles";
 import Title from "./Title";
 import { filteredConditionalItems, hasAnyItemsReplaceable } from "./utils";
+
+// FIXME should be moved elsewhere eventually
+const DefaultTestQuery = `query OrderDetails($id: ID!) { 
+  order(id: $id) { 
+    id
+    number
+    status
+    isShippingRequired
+    canFinalize
+    created
+    customerNote
+    paymentStatus
+    userEmail
+    isPaid
+  } 
+}`;
 
 export interface OrderDetailsPageProps {
   order: OrderDetailsFragment;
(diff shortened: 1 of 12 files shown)


zaiste:
I didn't know how to make a horizontal space between these buttons in Macaw UI :P

witoszekdev:
Please generate messages files: `npm run extract-messages` and commit ;)

src/orders/components/OrderDetailsPage/OrderDetailsPage.tsx
```diff
@@ -47,6 +48,22 @@ import { useStyles } from "./styles";
...
+    isShippingRequired
+    canFinalize
+    created
+    customerNote
+    paymentStatus
+    userEmail
+    isPaid
+  } 
+}`;
+
```
witoszekdev, line 66: You could move that to `src/orders/queries.ts`

src/orders/components/OrderDetailsPage/OrderDetailsPage.tsx
```diff
@@ -211,9 +228,30 @@ const OrderDetailsPage: React.FC<OrderDetailsPageProps> = props => {
...
+                    onClick={() => {
+                      const playgroundURL = new URL(process.env.API_URI);
+                      playgroundURL.hash = encodeGraphQLStatement({
+                        query: DefaultTestQuery,
+                        headers: "",
+                        operationName: "",
+                        variables: `{ "id": "${order.id}" }`,
+                      });
+                      window.open(playgroundURL, "_blank").focus();
+                    }}
```
witoszekdev, line 243: Why are we using `onClick` handler rather than providing `href`? Is there some performance concern?

FYI: You can change import of button to `components/Button` it uses `a` tag instead of `button` when `href` is passed (it still looks like a button)

src/products/components/ProductUpdatePage/ProductUpdatePage.tsx
```diff
@@ -66,6 +67,16 @@ import {
...
+// FIXME should be moved elsewhere eventually
+const DefaultTestQuery = `query ProductDetails($id: ID!) {
+  product(id: $id) {
+    id
+    name
+    slug
+    description
+  }
+}`;
+
```
witoszekdev, line 79: Please move to `products/queries.ts`

src/orders/components/OrderDetailsPage/OrderDetailsPage.tsx
zaiste: @witoszekdev The intention was to defer the URL compression (from `encodeGraphQLStatement`) to the time the button is clicked; also, eventually this default query could be dynamic. What's your take?

witoszekdev: My take was that links are better for accessibility: they behave differently than buttons (triggered on Enter rather than Space) + most a11y tools like screen readers can list every link on a page for faster navigation (e.g. VoiceOver on macOS can list all links on a page in rotor).
But on the contrary, if generating a link on every page visit would be too expensive performance wise then maybe using a button is a better idea - it still opens the GraphiQL in a new window, so the users should know what is going on. We could make it even more obvious to screen reader users by adding aria-label that explains what will happen if you click this button (e.g. "Opens example query in GraphiQL for this order in a new browser window")

witoszekdev (approved):

src/products/components/ProductUpdatePage/ProductUpdatePage.tsx
```diff
@@ -320,7 +322,32 @@ export const ProductUpdatePage: React.FC<ProductUpdatePageProps> = ({
...
+                      onClick={() => {
+                        const playgroundURL = new URL(process.env.API_URI);
+                        playgroundURL.hash = encodeGraphQLStatement({
+                          query: defaultGraphiQLQuery,
+                          headers: "",
+                          operationName: "",
+                          variables: `{ "id": "${product.id}" }`,
+                        });
+                        window.open(playgroundURL, "_blank").focus();
+                      }}
```
orzechdev, line 340: It looks like potential function handler, e.g.:
```suggestion
                      onClick={openPlayground}
```
and somewhere in component:
```ts
const openPlaygroundURL = playgroundOpenHandler({
  query: defaultGraphiQLQuery,
  headers: "",
  operationName: "",
  variables: `{ "id": "${product.id}" }`,
});
```

src/orders/components/OrderDetailsPage/OrderDetailsPage.tsx
zaiste: @witoszekdev the button has been converted into a link and moved under the details submenu, i.e. `CardMenu`

andrzejewsky (approved):

Droniu (approved):

PR #2983, opened 2023-01-13
Poc mocked functional testing

karolm-saleor:
This PR is a Proof Of Concept with mocked api request from GraphQL

How to run:

Start the dashboard app locally and copy the token from devtools - application - cookies - refreshToken values
Paste the token value in:
cypress/support/customCommands/user/index.js  53:57
cypress/fixtures/mockedData/app/refreshTokenWithUser.json  4:23

Start Cypress with parameter: `--config-file=cypress-integration.config.js`

### Pull Request Checklist

### Test environment config

APPS_MARKETPLACE_API_URI=https://marketplace-gray.vercel.app/api/v2/saleor-apps

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants
23. [ ] payments

CONTAINERS=1


karola312 (changes requested):
All tests failed

karolm-saleor:
> All tests failed

Added additional instruction in PR description

cypress/support/customCommands/user/index.js
```diff
@@ -47,3 +47,9 @@ Cypress.Commands.add(
     });
   },
 );
+
+Cypress.Commands.add("loginUserViaMockedToken", () => {
+  cy.window().then(appWindow => {
+    appWindow.localStorage.setItem("_saleorCSRFToken", "#token#");
```
poulch, line 53: question: should we have one mock token with some random data inside? Right now to run test required pass own token there, but this will be problem in CI, so IMO better will be use mocked token.

karola312: @karolm-saleor @poulch . So for now, I have to log in manually, then copy the token and paste it here. Won't it be better to just use the existing Command - loginUserViaRequest? This command sets _saleorCSRFToken in browser. Or we could use the [session](https://docs.cypress.io/api/commands/session), with login via API, and cache it across all tests. I don't see value in tests where you need to copy token

karolm-saleor: @andrzejewsky

cypress/support/api/utils/graphqlMockUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+
+// Alias mutation if operationName matches
+export const aliasMutation = (
+  req: CyHttpMessages.IncomingHttpRequest,
+  operationName: string,
+) => {
+  if (hasOperationName(req, operationName)) {
+    req.alias = operationName;
+  }
+};
```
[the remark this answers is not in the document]
karolm-saleor: done, deleted

karola312 (commented):
Can we shorten a bit mocked data files? ShopInfo.json has > 5000 lines. Do we really need that?

cypress-integration.config.js
```diff
@@ -0,0 +1,40 @@
...
+            }
+          });
+        }
+      });
+      return config;
+    },
+    baseUrl: "http://localhost:4173/",
+    specPattern: "cypress/functional/**/*.{js,jsx,ts,tsx}",
+  },
+});
```
karola312, line 40: Do we need to have two config files, there are almost same. Wouldn't be easier to just add environment variable functional with default false, and base on that create spec pattern? It could be run then via "npx cypress open --env functional=true"

cypress/support/customCommands/user/index.js
poulch: @karola312 we want to mock all requests. We don't need any unique token, the only requirement is that token contains proper structure but the data doesn't have to be accurate, it means such a user may not exists and it's ok.

cypress-integration.config.js
michalina-graczyk: Please don't create 2 config files if it is not a must.

michalina-graczyk (changes requested):
Please don't create 2 config files if it is not a must. In this case I can see that the only difference is a base url

PR #3040, opened 2023-01-24
Improve Polly.js config

witoszekdev:
I want to merge this change because it adds more options to run tests with Polly.js and fixes issues when running them locally:

- You can set 2 new env variables that determine the Polly.js behaviour: `POLLY_MODE` (replay, record, passthrough, stopped; by default it's `replay`) and `POLLY_RECORD_IF_MISSING` (by default: `true` if run **locally**, `false` on **CI**).
- Adapters and persister are imported dynamically (no need to use `.register`)

### Pull Request Checklist

### Test environment config

APPS_MARKETPLACE_API_URI=https://marketplace-gray.vercel.app/api/v2/saleor-apps

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants
23. [ ] payments

CONTAINERS=1


testUtils/api.ts
```diff
@@ -1,30 +1,44 @@
...
-import FSPersister from "@pollyjs/persister-fs";
 import { createFetch } from "@saleor/sdk";
+import { isCI } from "ci-info";
+import { get } from "env-var";
 import path from "path";
 import { setupPolly } from "setup-polly-jest";
 
+const POLLY_MODES = ["replay", "record", "passthrough", "stopped"] as const;
+
+const POLLY_MODE = get("POLLY_MODE")
```
andrzejewsky, line 12: **issue**: do you need a library for getting env value? we access process.env actually, node.js runs this, if we introduce a new library we also have to plan it to apply everywhere to have one way of accessing vars

package.json
```diff
@@ -32,19 +33,20 @@
...
     "@saleor/macaw-ui": "^0.8.0-pre.1",
     "@saleor/sdk": "^0.4.4",
-    "@floating-ui/react-dom-interactions": "^0.5.0",
     "@sentry/react": "^6.0.0",
     "@types/faker": "^5.1.6",
     "@uiw/react-color-hue": "0.0.34",
     "@uiw/react-color-material": "^0.1.0",
     "@uiw/react-color-saturation": "0.0.34",
     "apollo-upload-client": "^17.0.0",
+    "ci-info": "^3.7.1",
```
andrzejewsky, line 42: issue: we already have `is-ci` please use that instead

andrzejewsky (changes requested):

testUtils/api.ts
witoszekdev: The main idea here for this library was to abstract away the tricky part which is parsing of those values:
- we want `"false"` and `"true"` to be mapped to their correct boolean types
- we want to check if `POLLY_MODE` has the correct value declared in the union (otherwise raise exception)
- we want to provide a default value if none is provided by env var

Of course, we could do that by accessing raw env vars, like [what we have in saleor-app-checout](https://github.com/saleor/react-storefront/blob/main/apps/saleor-app-checkout/test-utils.ts#L181). But comparing that to the implementation with `env-var` it is much easier to understand and less error-prone in my opinion (we had to fix the implementation in checkout a couple of times 🙈).

We don't need to use that everywhere we use an env var, just where you need to parse it in some way.

package.json
witoszekdev: I don't think this is a great idea:
1. `is-ci` isn't used anywhere, so we don't need to limit ourselves ;)
2. `is-ci` actually uses... [`ci-info` under the hood](https://github.com/watson/is-ci/blob/master/index.js) and an old version which wasn't updated in 2 years 🙈

krzysztofzuraw (approved):

poulch (approved):

Droniu (approved):

PR #3359, opened 2023-03-14
Add job for cleaning environments

karola312:
I want to merge this change because we want to clean release testing environments and master staging on weekends

### Pull Request Checklist

### Test environment config

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants
23. [ ] payments

CONTAINERS=1

### DIFF
--- a/.github/workflows/clean-envs.yml
+++ b/.github/workflows/clean-envs.yml
@@ -1,8 +1,7 @@
-name: Execute nightly tests
-
+name: Clean environments
 on:
   schedule:
-    - cron: "00 0 * * 7"
+    - cron: "0 0 * * 0"
   workflow_dispatch:
     branches: ["main"]
 
--- a/.github/workflows/clean-envs.yml
+++ b/.github/workflows/clean-envs.yml
@@ -1,0 +1,34 @@
+name: Execute nightly tests
+
+on:
+  schedule:
+    - cron: "00 0 * * 7"
+  workflow_dispatch:
+    branches: ["main"]
+
+jobs:
+  cleanEnvironments:
+    runs-on: ubuntu-latest
+    env:
+      TOKEN: ${{ secrets.CLOUD_ACCESS_TOKEN }}
+      SNAPSHOT: 9oUFxT7f
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: ".nvmrc"
+
+      - name: Install dependencies
+        run: |
+          cd .github/workflows
+          npm ci
+
+      - name: clean environments
+        id: clean-environments
+        run: |
+          node .github/workflows/cleanEnvironments.js \
+          --token "$TOKEN" \
+          --snapshot "$SNAPSHOT"
--- a/.github/workflows/cleanEnvironments.js
+++ b/.github/workflows/cleanEnvironments.js
@@ -1,0 +1,113 @@
+const { Command } = require("commander");
+const fetch = require("node-fetch");
+
+const program = new Command();
+
+program
+  .name("cleanEnvironments")
+  .description("Clean environments")
+  .option("--token <token>", "token fo login to cloud")
+  .option("--snapshot <snapshot>", "snapshot to revert to")
+  .action(async options => {
+    const token = options.token;
+    const snapshot = options.snapshot;
+    const tasksArray = [];
+    const environmentsToClean = await getEnvironmentsForReleaseTesting(token);
+    environmentsToClean.forEach(environment => {
+      tasksArray.push(cleanEnvironment(environment, snapshot, token));
+    });
+    const throwErrorAfterTimeout = setTimeout(function () {
+      throw new Error("Environment didn't upgraded after 30 minutes");
+    }, 1800000);
+    Promise.all(tasksArray).then(tasksList => {
+      tasksList.forEach(task => {
+        if (task.error) {
+          console.warn(`${task.environment}: ${task.error}`);
+        } else {
+          waitUntilTaskInProgress(task.taskId, task.environment, token);
+        }
+      });
+    });
+    clearTimeout(throwErrorAfterTimeout);
+  })
+  .parse();
+
+async function getEnvironmentsForReleaseTesting(token) {
+  const environments = await getEnvironments(token);
+  const environmentsForReleaseTesting = environments.filter(environment => {
+    return (
+      environment.domain.match(/^v\d*.staging/) ||
+      environment.domain == "master.staging.saleor.cloud"
+    );
+  });
+  return environmentsForReleaseTesting;
+}
+
+async function getEnvironments(token) {
+  const response = await fetch(
+    `https://staging-cloud.saleor.io/api/organizations/saleor/environments/`,
+    {
+      method: "GET",
+      headers: {
+        Authorization: `Token ${token}`,
+        Accept: "application/json",
+        "Content-Type": "application/json;charset=UTF-8",
+      },
+    },
+  );
+  return await response.json();
+}
+
+async function cleanEnvironment(environment, snapshot, token) {
+  const response = await fetch(
+    `https://staging-cloud.saleor.io/api/organizations/saleor/environments/${environment.key}/restore/`,
+    {
+      method: "PUT",
+      body: `{
+        "restore_from": "${snapshot}"
+      }`,
+      headers: {
+        Authorization: `Token ${token}`,
+        Accept: "application/json",
+        "Content-Type": "application/json;charset=UTF-8",
+      },
+    },
+  );
+  const responseInJson = await response.json();
+  return {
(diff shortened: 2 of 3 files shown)


.github/workflows/clean-envs.yml
```diff
@@ -0,0 +1,34 @@
+name: Execute nightly tests
+
+on:
+  schedule:
+    - cron: "00 0 * * 7"
```
magul, line 5: is there a reason we have `00` doubled here?

magul: also that's `7` at 5th position here? From Github Actions docs:
```
┌───────────── minute (0 - 59)
│ ┌───────────── hour (0 - 23)
│ │ ┌───────────── day of the month (1 - 31)
│ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
│ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
│ │ │ │ │
│ │ │ │ │
│ │ │ │ │
* * * * *
```
they accept 0 to 6 (see: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule)

.github/workflows/cleanEnvironments.js
```diff
@@ -0,0 +1,113 @@
...
+    const snapshot = options.snapshot;
+    const tasksArray = [];
+    const environmentsToClean = await getEnvironmentsForReleaseTesting(token);
+    environmentsToClean.forEach(environment => {
+      tasksArray.push(cleanEnvironment(environment, snapshot, token));
+    });
+    const throwErrorAfterTimeout = setTimeout(function () {
+      throw new Error("Environment didn't upgraded after 30 minutes");
+    }, 1800000);
+    Promise.all(tasksArray).then(tasksList => {
```
magul, line 22: don't we need to `await` here before `clearTimeout`? Are we sure, that above error will be triggered after 30 minutes, if we will not wait to clear it for promises to be resolved?

.github/workflows/cleanEnvironments.js
```diff
@@ -0,0 +1,113 @@
...
+  );
+  return await response.json();
+}
+
+async function cleanEnvironment(environment, snapshot, token) {
+  const response = await fetch(
+    `https://staging-cloud.saleor.io/api/organizations/saleor/environments/${environment.key}/restore/`,
+    {
+      method: "PUT",
+      body: `{
```
magul, line 66: do we need to manually render that as a template? cannot we use `JSON.stringify`?

.github/workflows/cleanEnvironments.js
```diff
@@ -0,0 +1,113 @@
...
+    environmentsToClean.forEach(environment => {
+      tasksArray.push(cleanEnvironment(environment, snapshot, token));
+    });
+    const throwErrorAfterTimeout = setTimeout(function () {
+      throw new Error("Environment didn't upgraded after 30 minutes");
+    }, 1800000);
+    Promise.all(tasksArray).then(tasksList => {
+      tasksList.forEach(task => {
+        if (task.error) {
+          console.warn(`${task.environment}: ${task.error}`);
```
magul, line 25: isn't that something we want action to fail?

.github/workflows/cleanEnvironments.js
```diff
@@ -0,0 +1,113 @@
...
+  .name("cleanEnvironments")
+  .description("Clean environments")
+  .option("--token <token>", "token fo login to cloud")
+  .option("--snapshot <snapshot>", "snapshot to revert to")
+  .action(async options => {
+    const token = options.token;
+    const snapshot = options.snapshot;
+    const tasksArray = [];
+    const environmentsToClean = await getEnvironmentsForReleaseTesting(token);
+    environmentsToClean.forEach(environment => {
```
magul, line 16: do we need these to arrays? wouldn't that be more readable as:
```js
tasksArray = environmentsToClean.map(
  environment => cleanEnvironment(environment, snapshot, token));
```

.github/workflows/cleanEnvironments.js
magul: also it felt that's something belonging to `cleanEnvironment` function.

.github/workflows/cleanEnvironments.js
```diff
@@ -0,0 +1,113 @@
...
+  ) {
+    await new Promise(resolve =>
+      setTimeout(function () {
+        resolve(waitUntilTaskInProgress(taskId, environment, token));
+      }, 10000),
+    );
+  } else if (responseInJson.status == "SUCCEEDED") {
+    return responseInJson.status;
+  } else if (responseInJson.status !== "SUCCEEDED") {
+    console.warn(
```
magul, line 109: that probably should also make pipeline to fail? (and it's deal inside that function, not in that anonymous function in `.then()`.

.github/workflows/cleanEnvironments.js
```diff
@@ -0,0 +1,113 @@
...
+    },
+  );
+  const responseInJson = await response.json();
+  console.log(`${environment} - Job status - ${responseInJson.status}`);
+  if (
+    responseInJson.status == "PENDING" ||
+    responseInJson.status == "IN_PROGRESS"
+  ) {
+    await new Promise(resolve =>
+      setTimeout(function () {
```
magul, line 102: I ultimately understood, what you want to achieve here (as a side note - that's terrible, that there's not simpler construct to wait for particular time, or is it?), but would recommend to not rely on recursion here (in my personal opinion is more difficult to read/reason about) and just implement simple `while(true) {}`

.github/workflows/cleanEnvironments.js
```diff
@@ -0,0 +1,113 @@
...
+      method: "GET",
+      headers: {
+        Accept: "application/json",
+        "Content-Type": "application/json;charset=UTF-8",
+      },
+    },
+  );
+  const responseInJson = await response.json();
+  console.log(`${environment} - Job status - ${responseInJson.status}`);
+  if (
```
magul, line 97: instead of series of `if` and `else if` statements I would consider using:
```js
switch (responseInJson.status) {
  case "PENDING":
  case "IN_PROGRESS":
     wait10Seconds();
     break;
   case "SUCCEEDED":
     return responseInJson.status;
     break; // it's unnecessary as we return just above
   default:
     throw new Error(`${environment}: Job ended with status - ${responseInJson.status} - ${responseInJson.job_name}`);
}
```

.github/workflows/tests-nightly.yml
```diff
@@ -75,7 +75,7 @@ jobs:
         env:
           CI: true
           SALEOR_CLI_ENV: staging
-        run: npx saleor backup restore 9S4PckyI --skip-webhooks-update
+        run: npx saleor backup restore W5yhqHkw --skip-webhooks-update
```
magul, line 78: kind of magic number, don't really know, why it changes and why it's different then `SNAPSHOT` in cleaning pipeline.

.github/workflows/cleanEnvironments.js
karola312: I don't want to the whole workflow fail, eg. if cleaning the environment with version 3.9 fail because of some reason I still want to try to clean up environments 3.10, 3.11, 3.12...

.github/workflows/tests-nightly.yml
karola312: In this file, we are cleaning env which is in "QA" organization so with snapshot from "QA" org. Pipeline for cleaning environments is for "SALEOR" organization so it is using snapshot from that organization

.github/workflows/cleanEnvironments.js
karola312: Not sure if I want to fail whole pipeline, I think in the future it will be better to have all of the statuses and if any of this won't be 'SUCCEEDED' just send a slack notification to qa, which of this envs was not cleaned

michalina-graczyk (approved):

.github/workflows/cleanEnvironments.js
```diff
@@ -0,0 +1,113 @@
...
+      }`,
+    );
+  } else {
+    await waitUntilTaskInProgress(responseInJson.task_id, environment.name);
+  }
+}
+
+async function waitUntilTaskInProgress(taskId, environment) {
+  const throwErrorAfterTimeout = setTimeout(function () {
+    throw new Error("Environment didn't upgraded after 30 minutes");
```
NyanKiyoshi, line 80:
```suggestion
    throw new Error("Environment didn't upgrade after 30 minutes");
```

NyanKiyoshi (approved):

michalina-graczyk (approved):

PR #3789, opened 2023-06-22
Update app-sdk and pass saleor versions to handshake event for apps

lkostrowski:
Now handshake even will send saleor and dashboard versions to apps

Apps can easily use this to disable features which are not available in some Saleor versions, without calling API

- [x] Release stable SDK and update pr

Zrzut ekranu 2023-06-22 o 14 47 00

closes #...

### Pull Request Checklist

### Test environment config

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants
23. [ ] payments

CONTAINERS=1

### DIFF
--- a/src/apps/components/AppFrame/AppFrame.tsx
+++ b/src/apps/components/AppFrame/AppFrame.tsx
@@ -4,6 +4,7 @@ import { useUpdateAppToken } from "@dashboard/apps/components/AppFrame/useUpdate
 import { AppDetailsUrlQueryParams } from "@dashboard/apps/urls";
 import { useAllFlags } from "@dashboard/featureFlags";
 import { CircularProgress } from "@material-ui/core";
+import { DashboardEventFactory } from "@saleor/app-sdk/app-bridge";
 import clsx from "clsx";
 import React, { useCallback } from "react";
 
@@ -19,6 +20,8 @@ interface Props {
   className?: string;
   params?: AppDetailsUrlQueryParams;
   refetch?: () => void;
+  dashboardVersion: string;
+  coreVersion: string;
   onError?(): void;
 }
 
@@ -32,6 +35,8 @@ export const AppFrame: React.FC<Props> = ({
   params,
   onError,
   refetch,
+  dashboardVersion,
+  coreVersion,
 }) => {
   const frameRef = React.useRef<HTMLIFrameElement | null>(null);
   const classes = useStyles();
@@ -46,6 +51,10 @@ export const AppFrame: React.FC<Props> = ({
     appOrigin,
     appId,
     appToken,
+    {
+      core: coreVersion,
+      dashboard: dashboardVersion,
+    },
   );
 
   /**
@@ -62,13 +71,12 @@ export const AppFrame: React.FC<Props> = ({
      * Move handshake to notifyReady, so app is requesting token after it's ready to receive it
      * Currently handshake it 2 times, for compatibility
      */
-    postToExtension({
-      type: "handshake",
-      payload: {
-        token: appToken,
-        version: 1,
-      },
-    });
+    postToExtension(
+      DashboardEventFactory.createHandshakeEvent(appToken, 1, {
+        core: coreVersion,
+        dashboard: dashboardVersion,
+      }),
+    );
 
     setHandshakeDone(true);
   }, [appToken, postToExtension, setHandshakeDone]);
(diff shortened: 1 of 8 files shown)


src/apps/components/AppPage/AppPage.tsx
```diff
@@ -24,38 +26,48 @@ export const AppPage: React.FC<AppPageProps> = ({
...
-        )}
-      </Box>
-    </DetailPageLayout.Content>
-  </DetailPageLayout>
-);
+}) => {
+  const shop = useShop();
+
+  if (!shop?.version) {
+    return null; // loading?
```
krzysztofzuraw, line 33: **suggestion**: what about converting this to an issue - if it is a problem 🤔

krzysztofzuraw (dismissed):

src/apps/components/AppFrame/useAppActions.ts
```diff
@@ -3,11 +3,18 @@ import { usePostToExtension } from "@dashboard/apps/components/AppFrame/usePostT
 import { Actions, DispatchResponseEvent } from "@saleor/app-sdk/app-bridge";
 import React, { useState } from "react";
 
+/**
+ * TODO Refactor to named attributes
+ */
```
poulch, line 8: suggestion: leaving todo in code means nothing, wouldn't it be better to create an issue for this and link it here?

poulch (dismissed):

src/apps/components/AppPage/AppPage.tsx
```diff
@@ -24,38 +26,48 @@ export const AppPage: React.FC<AppPageProps> = ({
...
+  }
+
+  return (
+    <DetailPageLayout gridTemplateColumns={1} withSavebar={false}>
+      <AppPageNav
+        goBackUrl={AppUrls.resolveAppListUrl()}
+        appId={data.id}
```
Droniu, line 40:
```suggestion
        appId={data?.id}
```

src/apps/components/AppFrame/AppFrame.tsx
```diff
@@ -19,6 +20,8 @@ interface Props {
   className?: string;
   params?: AppDetailsUrlQueryParams;
   refetch?: () => void;
+  dashboardVersion: string;
+  coreVersion?: string;
```
Droniu, line 24: **question**: why is dashboard version required, and core optional? I don't see fallback in `postToExtension`

Droniu (dismissed):

src/apps/components/AppFrame/AppFrame.tsx
lkostrowski: For me, all should be required, but dashboard version is static  - from ENV, while core is async - fetched. 

IMO useShop() should block dashboard rendering, but I can't do that because it breaks ~60 tests.

I will revisit this - maybe I will just prevent rendering in Apps context only

src/apps/components/AppFrame/useAppActions.ts
lkostrowski: From my experience TODO comments in code are easier to track and fix, because IDEs track them and I see them every time near the code.

If I create an issue in the backlog for every small case like this, they will be forever gone in the backlog.

Here, I want to continue with refactoring src/apps so I left myself comments for future places I should refactor, so I don't mix refactors with features like here

src/apps/components/AppPage/AppPage.tsx
lkostrowski: I will revisit this

src/apps/components/AppPage/AppPage.tsx
lkostrowski: Fixed, but its something that should be refactored. We should not pass optional raw queries down the tree...

src/apps/components/AppFrame/AppFrame.tsx
lkostrowski: I checked this. I can't do it other way right now. AppFrame is used by ExternalAppProvider. I can't "early return" there, because it will break tests. This requires large refactor

src/apps/components/AppPage/AppPage.tsx
lkostrowski: https://github.com/saleor/apps/issues/682

poulch (approved):

krzysztofzuraw (approved):

PR #3800, opened 2023-06-23
Feature flags strategies

andrzejewsky:
### Pull Request Checklist

6. [ ] I added changesets and [read good practices](/.changeset/README.md)

### Test environment config

1. [ ] stable
2. [ ] giftCard
3. [ ] category
4. [ ] collection
5. [ ] attribute
6. [ ] productType
7. [ ] shipping
8. [ ] customer
9. [ ] permissions
10. [ ] menuNavigation
11. [ ] pages
12. [ ] sales
13. [ ] vouchers
14. [ ] homePage
15. [ ] login
16. [ ] orders
17. [ ] products
18. [ ] app
19. [ ] plugins
20. [ ] translations
21. [ ] navigation
22. [ ] variants
23. [ ] payments

CONTAINERS=1


src/featureFlags/FlagContent.ts
```diff
@@ -0,0 +1,21 @@
...
+    } catch (e) {
+      return FlagContent.empty();
+    }
+  }
+
+  public static empty() {
+    return new FlagContent(false, "");
+  }
+
+  public asString() {
```
lkostrowski, line 18: you can overwrite toString as well which is basically the same now I think

src/featureFlags/FlagsResolver.ts
```diff
@@ -0,0 +1,76 @@
...
+      .reduce(toSingleObject, {} as AvailableFlags.FlagList);
+  }
+
+  public getResult() {
+    return this.results;
+  }
+}
+
+const createInstances = (strategies: AvailableStrategies[]) =>
+  strategies.map(Constrctor => new Constrctor());
```
lkostrowski, line 68: Are you sure it should create constructors? You can't pass any configuration now. Why don't just accept instances of the interface?

src/featureFlags/useFlag.ts
```diff
@@ -0,0 +1,14 @@
+import { Name as FlagName } from "./availableFlags";
+import { useFeatureFlagContext } from "./context";
+import { FlagContent } from "./FlagContent";
+
+export const useFlag = (name: FlagName) => {
```
lkostrowski, line 5: I'd be good if you provide example of testing flags, maybe some simple Wrapper with static strategies list provided?

src/featureFlags/availableFlags.ts
```diff
@@ -0,0 +1,37 @@
+import { FlagContent } from "./FlagContent";
+
+const AVAILABLE_FLAGS = [
```
lkostrowski, line 3: I think available flags should come from ENV

andrzejewsky: I think it could be even a separated config, env are not enough for store quite sophisticated information (it's more than primitive values)

We are definite flags that can be used in the code (in this array), which also can be stored within the metadata in the future. Using these definition we are able to display some user-friendly information on the UI. Let's start with it for now.

lkostrowski: Please remember that flags are _not_ configuration, while its often tempting to model them like that. And they should be temporary - so likely flags are very simple strings or booleans. After a feature is merged, they are gone. I hope we don't overengineer them.

What is this non-primitive value you need?

krzysztofzuraw:
What about `Flagsmith` - should it not be removed?

krzysztofzuraw (dismissed):

src/featureFlags/availableFlags.ts
andrzejewsky: I'm renaming this to "payload"

krzysztofzuraw (approved):

src/featureFlags/FlagContent.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    expect(instance).toEqual({ enabled: false, payload: "" });
+  });
+
+  it("returns a JSON string", async () => {
+    // ARRANGE
+    const instance = new FlagValue(true, "some value");
+
+    const jsonString = instance.asString();
+
+    expect(jsonString).toEqual('{"enabled":true,"payload":"some value"}');
```
poulch, line 29: I would suggest keeping AAA naming using pascal case
```suggestion
    // Arrange
    const instance = new FlagValue(true, "some value");

    // Act
    const jsonString = instance.asString();

    // Assert
    expect(jsonString).toEqual('{"enabled":true,"payload":"some value"}');
```

src/featureFlags/FlagContent.ts
poulch: Agree ☝🏻

src/featureFlags/FlagsResolver.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (!AvailableFlags.isSupported(name)) {
+    return p;
+  }
+
+  if (!p[name]) {
+    p[name] = value;
+  }
+
+  return p;
+};
```
poulch, line 35: suggestion: I'd use function and put all of them below FlagsResolver

poulch (dismissed):

krzysztofzuraw (approved):

krzysztofzuraw (approved):

poulch (approved):

PR #4168, opened 2023-09-05
Playwright e2e setup

### LINKED ISSUE #4240: Migrate product type critical tests to Playwright
Sort and migrate e2e tests for product type from Cypress to Playwright.

AC:

- [ ] Migrated tests can be run locally

michalina-graczyk:
closes #4240 
Add first  Playwright setup
### Pull Request Checklist

6. [ ] I added changesets and [read good practices](../.changeset/README.md)

### Test environment config

1. [ ] stable
2. [ ] app
3. [ ] attribute
4. [ ] category
5. [ ] collection
6. [ ] customer
7. [ ] giftCard
8. [ ] homePage
9. [ ] login
10. [ ] menuNavigation
11. [ ] navigation
12. [ ] orders
13. [ ] pages
14. [ ] payments
15. [ ] permissions
16. [ ] plugins
17. [ ] productType
18. [ ] products
19. [ ] sales
20. [ ] shipping
21. [ ] translations
22. [ ] variants
23. [ ] vouchers

CONTAINERS=2


playwright/tests/auth.setup.ts
```diff
@@ -0,0 +1,22 @@
...
+  await loginPage.typeEmail(process.env.CYPRESS_USER_NAME!);
+  await loginPage.typePassword(process.env.CYPRESS_USER_PASSWORD!);
+  await loginPage.clickSignInButton();
+
+  const homePage = new HomePage(page);
+  await expect(homePage.welcomeMessage).toContainText("Hello there,");
+
+  // End of authentication steps.
+  await page.context().storageState({ path: adminFile });
+});
```
karola312, line 22: Do we need to log in via frontend in setup?

wojteknowacki: This is the main and recommended way by PW https://playwright.dev/docs/auth

wojteknowacki: You do it just once per role before all tests start and then all tests uses same storage and cookies 🍪  - looks as really nice solution

wojteknowacki: Plus you can treat it as sanity test - if you can't log in why bother testing rest of the stuff

karola312: Authenticate with API request is also recommended. From playwright docs:
When to use

Your web application supports authenticating via API that is easier/faster than interacting with the app UI.

I think in most cases test start to fail when somthing Has changed on the UI, not when it's broken. So in the case of missing data-test-id on some input, all tests will be broken, not just login. You will need to fix it asap, on order to run tests

playwright/api/basics.ts
```diff
@@ -0,0 +1,43 @@
+import {
+  APIResponse,
+  request,
+} from "@playwright/test";
+
+const URL = process.env.API_URI || "";
+interface DATA {
```
Droniu, line 7: Types and interfaces should be written in pascal case
```suggestion
interface Data {
```

playwright/api/basics.ts
```diff
@@ -0,0 +1,43 @@
...
+  query: string;
+}
+
+interface USER {
```
Droniu, line 11:
```suggestion
interface User {
```

playwright/tests/auth.setup.ts
wojteknowacki: @karola312 I see your point 🤔 but still I think it will not be a problem, it is beneficial since we have a copy of the browser state which is much more correlated to the end user. If this will cause trouble then we need to fix it in one place (setup) so let's give it a shot - if won't do the trick I will change it to API. Is that fair enough?

playwright.config.ts
```diff
@@ -0,0 +1,39 @@
...
+  forbidOnly: !!process.env.CI,
+  retries: process.env.CI ? 2 : 0,
+  workers: process.env.CI ? 1 : undefined,
+  reporter: "html",
+  // webServer: {
+  //   command: "npm run dev",
+  //   url: "http://localhost:9000/",
+  //   reuseExistingServer: !process.env.CI,
+  // },
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
```
poulch, line 19: question: do you need those comments?

playwright.config.ts
```diff
@@ -0,0 +1,39 @@
...
+  use: {
+    baseURL: process.env.CYPRESS_baseUrl,
```
poulch, line 21: question: why not `CYPRESS_BASEURL`?

playwright/tests/orders.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+import {
+  expect,
+  test,
+} from "@playwright/test";
+
+test.use({ storageState: "playwright/.auth/admin.json" });
+
+test("Orders", async ({ page }) => {
+  await page.goto(URL_LIST.orders);
+  const homePage = new HomePage(page);
```
poulch, line 12: question: do you need homePage here?

poulch (approved):

szczecha (approved):

szczecha:
One thing I see for playwright you start using different test case ID, please make sure that it is reflected in testmo

wojteknowacki:
@szczecha When migrating critical tests is done I will remove cypress tests and in parallel clean testmo as you suggested 👍

playwright.config.ts
wojteknowacki: I left only those which can be helpful later :)

playwright.config.ts
wojteknowacki: TBH it was left for consistency - we use same names in workflows and tests so I did not want to introduce changes in this PR - I will create separate PR to clean it up

PR #4451, opened 2023-11-17
New deployment targets for multitenant dashboard

magul:
Relates saleor/saleor-cloud-infra#1029


.github/workflows/deploy-cloud.yaml
```diff
@@ -34,32 +40,51 @@ jobs:
       - name: Build
         run: |
           npm run build
+
+      - name: Configure AWS Credentials
```
magul, line 44: two comments here from me:
 * this is primary candidate to be abstarcted in workflow call
 * we could stop duplicating these steps and use https://github.com/actions/runner/issues/409#issuecomment-727565588, but, I believe, my current approach is more readable.

.github/workflows/deploy-cloud.yaml
```diff
@@ -34,32 +40,51 @@ jobs:
...
+              done
+              env:
+                DEPLOYMENT_BUCKET: ${{ secrets.AWS_CLOUD_US_DEPLOYMENT_BUCKET }}
+                CDN_DISTRIBUTIONS: ${{ secrets.AWS_CLOUD_US_CDN_DISTRIBUTIONS }}
 
-          aws s3 sync build/dashboard s3://${BUCKET}/${ENVIRONMENT}/static/
-          aws s3 cp build/dashboard/index.html s3://${BUCKET}/${ENVIRONMENT}/
-          for i in {1..3}; do aws cloudfront create-invalidation --distribution-id ${CF_ID} --paths "/dashboard*" && break || sleep 5; done
-          if [[ -n "$CF_2_ID" ]]; then
-            for i in {1..3}; do aws cloudfront create-invalidation --distribution-id ${CF_2_ID} --paths "/dashboard*" && break || sleep 5; done
```
magul, line 61: that allow us to invalidate all 4 distributions we currently have in EU, and if another will be introduced, we can just update secret and it will work just fine

andrzejewsky (dismissed):

.github/workflows/deploy-cloud.yaml
```diff
@@ -20,6 +20,12 @@ jobs:
       APPS_MARKETPLACE_API_URI: "https://apps.saleor.io/api/v2/saleor-apps"
       IS_CLOUD_INSTANCE: true
     steps:
+      - name: Check region
+        if: ${{ !contains(fromJSON('["eu", "us"]'), env.REGION) }}
+        uses: actions/github-script@v3
```
NyanKiyoshi, line 25: nodejs 12 is deprecated, use the latest version. But using an action is useless here, just use `echo 'something' >&2; exit 1`

.github/workflows/deploy-cloud.yaml
```diff
@@ -34,32 +40,51 @@ jobs:
...
+          aws s3 sync build/dashboard "s3://${DEPLOYMENT_BUCKET}/${ENVIRONMENT}/static/"
+          aws s3 cp build/dashboard/index.html "s3://${DEPLOYMENT_BUCKET}/${ENVIRONMENT}/"
+
+          IFS=',' read -ra DISTRIBUTIONS_ARRAY <<< "${CDN_DISTRIBUTIONS}"
+          for DISTRIBUTION in "${DISTRIBUTIONS_ARRAY[@]}"; do
+            for i in {1..3}; do aws cloudfront create-invalidation --distribution-id "$DISTRIBUTION" --paths "/dashboard*" && break || sleep 5; done
+          done
+          env:
+            DEPLOYMENT_BUCKET: ${{ secrets.AWS_CLOUD_EU_DEPLOYMENT_BUCKET }}
+            CDN_DISTRIBUTIONS: ${{ secrets.AWS_CLOUD_EU_CDN_DISTRIBUTIONS }}
```
NyanKiyoshi, line 72: Is `CDN_DISTRIBUTIONS` a secret or not? Secrets shouldn't mutate as they also are removed from logs if found (you are splitting the value)

.github/workflows/deploy-cloud.yaml
```diff
@@ -34,32 +40,51 @@ jobs:
...
+
+            - name: Deploy
+            if: env.REGION == 'us'
+            run: |
+              aws s3 sync build/dashboard "s3://${DEPLOYMENT_BUCKET}/${ENVIRONMENT}/static/"
+              aws s3 cp build/dashboard/index.html "s3://${DEPLOYMENT_BUCKET}/${ENVIRONMENT}/"
+
+              IFS=',' read -ra DISTRIBUTIONS_ARRAY <<< "${CDN_DISTRIBUTIONS}"
+              for DISTRIBUTION in "${DISTRIBUTIONS_ARRAY[@]}"; do
+                for i in {1..3}; do aws cloudfront create-invalidation --distribution-id "$DISTRIBUTION" --paths "/dashboard*" && break || sleep 5; done
```
NyanKiyoshi, line 82: Instead of duplicating code across all branches and workflows, you could use composite actions

NyanKiyoshi (changes requested):

.github/workflows/deploy-cloud.yaml
magul: Set
```
            echo "::add-mask::$DISTRIBUTION"
```

.github/workflows/deploy-cloud.yaml
magul: I agree with that, but it's outside of the scope of the ticket I'm working (and I know that you're last time pretty vocal about not doing additional work in ticket).

If that would be only SRE-maintained repository, I would probably try to refactor that, but as it's not maintained by SRE, I would prefer to leave that topic for @saleor/dashboard team to address.

.github/workflows/deploy-cloud.yaml
magul: I agree with that and if that would be only-SRE-maintaned project I would try to do that, as that's something that @saleor/dashboard team owns, I would prefer to leave that for them to address this duplication.

magul: This will also require consideration about main staging/fake production (where we deploy to two S3 targets, but invalidate cache only on one distribution.

andrzejewsky (dismissed):

.github/workflows/deploy-cloud.yaml
```diff
@@ -34,32 +38,53 @@ jobs:
...
+          aws s3 cp build/dashboard/index.html "s3://${DEPLOYMENT_BUCKET}/${ENVIRONMENT}/"
+
+          IFS=',' read -ra DISTRIBUTIONS_ARRAY <<< "${CDN_DISTRIBUTIONS}"
+          for DISTRIBUTION in "${DISTRIBUTIONS_ARRAY[@]}"; do
+            echo "::add-mask::$DISTRIBUTION"
+            for i in {1..3}; do aws cloudfront create-invalidation --distribution-id "$DISTRIBUTION" --paths "/dashboard*" && break || sleep 5; done
+          done
+          env:
+            DEPLOYMENT_BUCKET: ${{ secrets.AWS_CLOUD_EU_DEPLOYMENT_BUCKET }}
+            CDN_DISTRIBUTIONS: ${{ secrets.AWS_CLOUD_EU_CDN_DISTRIBUTIONS }}
```
NyanKiyoshi, line 71:
```suggestion
          done
        env:
          DEPLOYMENT_BUCKET: ${{ secrets.AWS_CLOUD_EU_DEPLOYMENT_BUCKET }}
          CDN_DISTRIBUTIONS: ${{ secrets.AWS_CLOUD_EU_CDN_DISTRIBUTIONS }}
```

.github/workflows/deploy-cloud.yaml
```diff
@@ -34,32 +38,53 @@ jobs:
...
+          aws s3 cp build/dashboard/index.html "s3://${DEPLOYMENT_BUCKET}/${ENVIRONMENT}/"
+
+          IFS=',' read -ra DISTRIBUTIONS_ARRAY <<< "${CDN_DISTRIBUTIONS}"
+          for DISTRIBUTION in "${DISTRIBUTIONS_ARRAY[@]}"; do
+            echo "::add-mask::$DISTRIBUTION"
+            for i in {1..3}; do aws cloudfront create-invalidation --distribution-id "$DISTRIBUTION" --paths "/dashboard*" && break || sleep 5; done
+          done
+          env:
+            DEPLOYMENT_BUCKET: ${{ secrets.AWS_CLOUD_US_DEPLOYMENT_BUCKET }}
+            CDN_DISTRIBUTIONS: ${{ secrets.AWS_CLOUD_US_CDN_DISTRIBUTIONS }}
```
NyanKiyoshi, line 86:
```suggestion
          done
        env:
          DEPLOYMENT_BUCKET: ${{ secrets.AWS_CLOUD_US_DEPLOYMENT_BUCKET }}
          CDN_DISTRIBUTIONS: ${{ secrets.AWS_CLOUD_US_CDN_DISTRIBUTIONS }}
```

NyanKiyoshi (approved):

NyanKiyoshi (approved):

andrzejewsky (approved):

PR #4499, opened 2023-11-28
#4128 category and collections columns

### LINKED ISSUE #4128: [Feature]: Category and Collections columns in products datagrid
### Description of the feature

It would be really helpful if there was a `Category` and `Collections` column in the `Products` page datagrid. There are columns for most of the other options that the product has but `Category` and `Collections` are missing.

### Acceptable criteria

- [ ] As a user, I should see the Categories column in the datagrid on the products page 
- [ ] As a user, I should see the Collections column in the datagrid on the  products page 
- [ ] As a user, I should be able to enable/disable those columns in the datagrid options

### Additional information

_No response_

pitkes22:
closes #4128

This pull request adds `Category` and `Collections` columns to the Product List Datagrid.
### Screenshots

The Category column uses `pill` cell to render, in order to match with the type column.
The Collections column uses `tags` cell to render because `pill` cell can't render multiple values.

The color of the elements in both columns is calculated based on the value. This is good for making the different values easily distinguishable. But on the other hand, it makes the screen look a bit too colorful. I am not sure if this is the best solution, or if it would be better to use a single color for all the elements. 

image

The columns are currently not enabled by default.

image

### Pull Request Checklist

4. [x] The changes are tested in Chrome/Firefox/Safari browsers and in light/dark mode
5. [x] Your code works with the latest stable version of the core
6. [x] I added changesets file (instructions in [contribution guide](https://github.com/saleor/saleor-dashboard/blob/main/.github/CONTRIBUTING.md)

### Test environment config

1. [ ] stable
2. [ ] app
3. [ ] attribute
4. [ ] category
5. [ ] collection
6. [ ] customer
7. [ ] giftCard
8. [ ] homePage
9. [ ] login
10. [ ] menuNavigation
11. [ ] navigation
12. [ ] orders
13. [ ] pages
14. [ ] payments
15. [ ] permissions
16. [ ] plugins
17. [ ] productType
18. [x] products
19. [ ] sales
20. [ ] shipping
21. [ ] translations
22. [ ] variants
23. [ ] vouchers

CONTAINERS=2


poulch (approved):

src/fragments/products.ts
```diff
@@ -110,6 +110,14 @@ export const productFragment = gql`
...
       hasVariants
     }
+    category {
+      id
+      name
+    }
+    collections {
+      id
+      name
+    }
```
andrzejewsky, line 120: We can use `@include(if: ...)` statements when the columns are toggled, otherwise we download the data even if we don't use them, especially when they are disabled by default

andrzejewsky (changes requested):

src/fragments/products.ts
pitkes22: 405b68c

poulch (approved):

PR #4631, opened 2024-01-23
[SHOPX-66] Select tax app on taxes configuration

krzysztofzuraw:
## What type of PR is this?
- [ ] 💅 Refactor
- [x] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- https://linear.app/saleor/issue/SHOPX-66

CleanShot 2024-01-29 at 15 33 16@2x
This PR can be merged after https://github.com/saleor/saleor/pull/15292 is merged to `saleor/saleor` main branch.
GraphQL endpoint for local testing: https://block-when-tax-failed.api.saleor.rocks/graphql/

## Have you written tests?
- [x] Yes!

## Description

This pull request allows the selection of a tax app to calculate tax for a specific channel. Previously, merchants could only choose between using a tax app or flat rates, which caused issues when there were multiple tax apps for a shop. This pull request solves the problem by allowing the selection of a specific tax app for each channel.


package.json
```diff
@@ -39,7 +39,7 @@
     "@material-ui/styles": "^4.11.4",
     "@reach/auto-id": "^0.16.0",
     "@saleor/macaw-ui": "npm:@saleor/macaw-ui@0.7.4",
-    "@saleor/macaw-ui-next": "npm:@saleor/macaw-ui@1.0.0-pre.18",
+    "@saleor/macaw-ui-next": "npm:@saleor/macaw-ui@1.0.0-pre.21",
```
lkostrowski, line 42: Is there a reason this PR bumps Macaw in this PR? I think its safer in general to avoid mixing too many things in single commit

src/apps/components/AppAvatar/AppAvatar.tsx
```diff
@@ -28,8 +28,8 @@ export const AppAvatar: React.FC<{
       borderRadius={2}
       borderWidth={1}
       borderColor="default1"
-      borderStyle={"solid"}
+      borderStyle="solid"
     >
-      <GenericAppIcon size="large" color="default2" />
+      <GenericAppIcon size="medium" color="default2" />
```
lkostrowski, line 33: This component is used in other contexts, like AppDetails / AppView, is this breaking UI?

src/taxes/components/TaxAppLabel.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { FormattedMessage } from "react-intl";
+
+interface Props {
+  app: TaxStrategyChoicesQuery["shop"]["availableTaxApps"][number];
+}
+
+export const TaxAppLabel: React.FC<Props> = ({ app }) => {
+  const logo = app.brand?.logo?.default
+    ? { source: app.brand?.logo?.default }
+    : undefined;
```
lkostrowski, line 16: I dont want to introduce any revolution but IMO this coupling is not necessary. This is pure UI level which is now heavily bound to how graphQL looks like. You need name, logo, some date. IMO it much clearer if component decalres these fields:

```
interface Props {
name: string;
logo?: string 
createdAt: string
}
```

Component, when used, will accept in props source that originally comes from graphQL... but it doesn't have to. Its easier, simpler and more testable in general

package.json
krzysztofzuraw: Good idea - I've backported this change in https://github.com/saleor/saleor-dashboard/pull/4643. We need version 21 as it fixes colors for warning banner while uninstalling apps.

src/apps/components/AppAvatar/AppAvatar.tsx
krzysztofzuraw: I've just double checked and it didn't break UI in list and details of an app. Only icon inside is a bit smaller:

| Before | After |
| ------- | ------- |
| CleanShot 2024-01-30 at 12 37 52@2x | CleanShot 2024-01-30 at 12 37 45@2x |

src/taxes/pages/TaxChannelsPage/useTaxStrategyChoices.tsx
```diff
@@ -0,0 +1,33 @@
...
+          version={app.version}
+          id={app.id}
+          identifier={app.identifier}
+          logoUrl={app.brand?.logo?.default}
+        />
+      ),
+    })) ?? [];
+
+  const choices = [...taxAppsChoices, flatTaxRateChoice];
+  return [choices, loading] as const;
```
lkostrowski, line 32: I think this API is worse than `return {choices, loading} as const`.

As a user of this hook I need to check its internal code to understand what is behind the tuple. **And I will name them "choices" and "loading" anyway**

This is not react "useState" → here we are specific what is behind the values

lkostrowski (dismissed):

lkostrowski (approved):

krzysztofzuraw:
I've got failing e2e with shipping method but my changes did not touch those pages - is it something known @andrzejewsky / @Droniu ?

andrzejewsky (approved):

src/taxes/pages/TaxChannelsPage/TaxChannelsPage.tsx
```diff
@@ -53,12 +58,19 @@ interface TaxChannelsPageProps {
...
+  TaxConfigurationPerCountryFragment,
+  "taxCalculationStrategy"
+> & {
+  taxCalculationStrategy: string;
+};
+
 export interface TaxConfigurationFormData {
   chargeTaxes: boolean;
-  taxCalculationStrategy: TaxCalculationStrategy;
+  taxCalculationStrategy: string;
```
poulch, line 70: question:  can we use stronger type here?

poulch (commented):
Could you write integration test for part that you added?

src/taxes/pages/TaxChannelsPage/TaxChannelsPage.tsx
krzysztofzuraw: We can have now multiple values here:
1) `flatRate` from `TaxCalculationStrategy`
2) id of an app
3) plugin id 

That is why I weaken type here

poulch (approved):

lkostrowski (approved):

PR #4705, opened 2024-03-04
Add shortcuts at bottom of sidebar

poulch:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!


src/components/Navigator/NavigatorProvider.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { ReactNode, useState } from "react";
+
+import { NavigatorContext } from "./useNavigatorContext";
+
+export const NavigatorProvider = ({ children }: { children: ReactNode }) => {
```
andrzejewsky, line 5: if you touch navigator components - don't you think to rename it? It is a search bar, perhaps we could rename it?

andrzejewsky: it can be SearchOverlay, NavigatorSearch, NavigatorSearchBar or something similar?

src/components/Sidebar/shortcuts/useShortcuts.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const controlKey = getControlKey();
+
+  const handleOpenPlayground = useCallback(() => {
+    devContext.setDevModeContent("");
+    devContext.setVariables("");
+    devContext.setDevModeVisibility(true);
+  }, []);
+
+  const handleOpenSearch = useCallback(() => {
+    setNavigatorVisibility(true);
```
andrzejewsky, line 25: the provider you created has its own state and you wrap the entire dashboard in it, so now when you make this call, it will re-render the whole application.

Both "navigator" and playground live within the app layout, so why not place them close to the logic that triggers them? Then you get rid of global context?

andrzejewsky (changes requested):

andrzejewsky (dismissed):

yellowee (approved):

andrzejewsky (approved):

PR #4708, opened 2024-03-05
[E2E][TEST] Add tests for thirdparty apps

szczecha:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!


playwright/tests/apps.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+test("TC: SALEOR_119 User should be able to install and configure app from manifest @e2e", async ({
+  page,
+}) => {
+  const appsPage = new AppsPage(page);
+  const installationPage = new AppInstallationPage(page);
+
+  await appsPage.gotoAppsList();
+  await expect(appsPage.installExternalAppButton).toBeVisible();
+  await appsPage.installExternalAppButton.click();
+  await appsPage.typeManifestUrl("https://klaviyo.saleor.app/api/manifest");
```
witoszekdev, line 21: question: Why Klaviyo app specifically? Maybe we should have a debug app that will return data parsed from JWT and check if it's valid by using `app-sdk`

witoszekdev (approved):

playwright/tests/apps.spec.ts
szczecha: @witoszekdev I chose an application that did not throw an error when entering a random value 😅

witoszekdev: We'll create a debug app just for that, that will validate JWT after clicking a button. Let's leave it as is for now

yellowee (approved):

PR #4738, opened 2024-03-19
Allow user without app permission access apps view

poulch:
## What type of PR is this?
This PR allows all users access to APPS tab. Since all users will have access there we moved checking required permission per app.

Users without `MANAGE_APPS` will be able to:
1. Visit the Apps tab 
2. See a list of all installed apps 
2. See a list of all available apps
3. See a list of upcoming apps
4. Enter each app and see its content only when the app allows for that
5. See each app settings (for this kind of user I rename the button label from `Manage app` to `App settings`)

Users without `MANAGE_APPS` will ***not*** be able to:
1. See list of app installations
2. Install any app 
3. Manage app (delete, active, reactive, change permissions)

- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [x] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!


yellowee (dismissed):
Tested manually, e2e tests will be added in a separate PR

peelar:
Please update the documentation

https://docs.saleor.io/docs/3.x/developer/extending/apps/architecture/app-permissions

src/apps/components/InstalledAppList/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+  appInstallationList,
+  appList,
+  hasManagedAppsPermission,
+}: {
+  appList?: InstalledApp[];
+  appInstallationList?: AppInstallation[];
+  hasManagedAppsPermission?: boolean;
+}): boolean {
+  if (!hasManagedAppsPermission) {
+    return !appList;
```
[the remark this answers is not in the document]
andrzejewsky: ahh ok you probably check if it is defined/undefined

andrzejewsky (changes requested):

lkostrowski:
@andrzejewsky please add apps guild "write" access so we can call team for a cr

src/apps/components/InstalledAppList/utils.ts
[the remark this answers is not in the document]
poulch: yes, because when there is no data we show loader, would be better to pass loading flag

src/components/ButtonWithTooltip/ButtonWithTooltip.tsx
```diff
@@ -0,0 +1,29 @@
...
+  children: React.ReactNode;
+}
+
+export const ButtonWithTooltip = ({
+  tooltip,
+  children,
+  ...props
+}: ButtonWithTooltipProps) => {
+  if (!tooltip) {
+    return <Button {...props}>{children}</Button>;
```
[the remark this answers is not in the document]
poulch: This is quite a common situation the tooltip is shown conditionally. The current implementation of tooltip shows a tooltip even when there is no content, so I have to prepare the component twice one with the wrapped tooltip and once without it.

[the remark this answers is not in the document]
andrzejewsky: Ok so perhaps the name is wrong?

andrzejewsky (dismissed):

.changeset/thin-avocados-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Allow user without MANAGE_APPS to access to apps page
```
lkostrowski, line 5: I wish we write better changelogs, not commit messages ;)

lkostrowski: People will see this changelogs when browsing app history, its empty. The context is in PR description, which is hard to retrieve browsing history

yellowee (approved):

src/apps/components/AllAppList/AllAppList.tsx
```diff
@@ -20,18 +19,24 @@ const AllAppList: React.FC<AllAppListProps> = ({
...
+      __gridTemplateColumns="repeat(2, 1fr)"
+      __gap={`${themeValues.spacing[8]} ${themeValues.spacing[5]}`}
+      padding={5}
+      marginTop={5}
+    >
+      {appList.map(app => (
         <AppListRow
-          key={appPair[0].name.en}
-          appPair={appPair}
+          key={app.name.en}
```
krzysztofzuraw, line 38: I think we allow multiple apps with the same name - maybe we can add `app.name.en-app.id`?

src/apps/components/AppListRow/AppListCardDescription.tsx
```diff
@@ -11,15 +11,7 @@ interface AppListCardDescriptionProps {
...
-  <Box
-    borderStyle="solid"
-    borderWidth={1}
-    borderTopLeftRadius={3}
-    borderTopRightRadius={3}
-    borderColor="default1"
-    borderBottomStyle="none"
-    padding={5}
-  >
+  <Box>
```
krzysztofzuraw, line 14: **Question:** do we need that `Box`?

src/apps/components/InstalledAppList/utils.test.ts
```diff
@@ -0,0 +1,248 @@
...
+
+        // Act
+        const isLoading = appsAreLoading({
+          appList,
+          appInstallationList,
+          hasManagedAppsPermission,
+        });
+
+        // Assert
+        expect(isLoading).toBe(true);
```
krzysztofzuraw, line 23: **Question:** shouldn't it be false?

src/apps/components/InstalledAppList/utils.test.ts
```diff
@@ -0,0 +1,248 @@
...
```
krzysztofzuraw, line 59: **Question:** shouldn't it be false?

src/apps/components/AppListRow/AppListCardDescription.tsx
poulch: Yes, as this is children in css grid as whole

krzysztofzuraw (approved):

andrzejewsky (approved):

PR #4740, opened 2024-03-20
Fix broken link to app details on tax configuration

krzysztofzuraw:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!

### DIFF
--- a/src/taxes/components/TaxAppLabel.tsx
+++ b/src/taxes/components/TaxAppLabel.tsx
@@ -1,7 +1,7 @@
 import { AppAvatar } from "@dashboard/apps/components/AppAvatar/AppAvatar";
 import { AppUrls } from "@dashboard/apps/urls";
 import useNavigator from "@dashboard/hooks/useNavigator";
-import { Box, ExternalLinkIcon, Text } from "@saleor/macaw-ui-next";
+import { Box, Button, ExternalLinkIcon, Text } from "@saleor/macaw-ui-next";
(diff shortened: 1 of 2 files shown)


poulch (approved):

src/taxes/components/TaxAppLabel.tsx
```diff
@@ -68,12 +75,11 @@ export const TaxAppLabel: React.FC<TaxAppLabelProps> = ({
       </Box>
       <Box
         as="a"
```
witoszekdev, line 77: suggestion: This is now an invalid link so, let's better use `button` instead:

```suggestion
        as="button"
```

witoszekdev (changes requested):
If that fixes the issue I guess it's fine (let's just change the `a` to be a `button`), but why is this even necessary? I don't like links that aren't actually links, since it breaks a lot of browsers functionality and accessibility, e.g. "Open in new tab" ;)

src/taxes/components/TaxAppLabel.tsx
krzysztofzuraw: I updated to be a button from new MacawUI

poulch (approved):

witoszekdev (approved):

PR #4745, opened 2024-03-21
Implement refund add modal

Droniu:
## What type of PR is this?
- [ ] 💅 Refactor
- [x] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

image

## Have you written tests?
- [x] Yes!

## [Optional] Description

You need to enable feature flag to test this modal:
```
localStorage.setItem('FF_improved_refunds', '{"enabled":true,"payload":"default"}')
```
It's available on any order that has transactions


package.json
```diff
@@ -37,6 +37,7 @@
     "@material-ui/icons": "^4.11.3",
     "@material-ui/lab": "^4.0.0-alpha.61",
     "@material-ui/styles": "^4.11.4",
+    "@radix-ui/react-radio-group": "^1.1.3",
```
poulch, line 40: suggestion: installing an additional library which is also used in Macaw might cause we will have a double package in bundle. That's one thing. The question in why can't you use RadioGroup component from macaw.

src/components/RadioTile/RadioTile.tsx
```diff
@@ -0,0 +1,94 @@
...
+        onMouseLeave={() => setHoverState(false)}
+      >
+        <Box
+          width={5}
+          height={5}
+          borderRadius="100%"
+          display="flex"
+          justifyContent="center"
+          alignItems="center"
+          backgroundColor={getHoverState({ checked, isHoverState })}
```
poulch, line 47: suggestion: I'd name function `getHoverState` in the way that suggest what it does

src/components/RadioTile/RadioTile.tsx
```diff
@@ -0,0 +1,94 @@
...
+            >
+              <svg
+                xmlns="http://www.w3.org/2000/svg"
+                width="6"
+                height="6"
+                fill="white"
+              >
+                <circle cx="3" cy="3" r="3" fill="currentColor" />
+              </svg>
+            </RadixRadioGroup.Indicator>
```
poulch, line 83: suggestion: this might be moved to component

src/orders/components/OrderRefundDatagrid/OrderRefundDatagrid.tsx
```diff
@@ -76,39 +80,40 @@ export const OrderRefundDatagrid: React.FC<OrderRefundDatagridProps> = ({
...
-          renderColumnPicker={() => (
-            <ColumnPicker
-              selectedColumns={selectedColumns}
-              staticColumns={staticColumns}
-              onToggle={handlers.onToggle}
-            />
-          )}
-        />
-      </DatagridChangeStateContext.Provider>
+      {grantedRefunds.length !== 0 && (
```
poulch, line 88: question: don't we want to show some empty state?

src/orders/views/OrderDetails/OrderNormalDetails/index.tsx
```diff
@@ -457,6 +459,13 @@ export const OrderNormalDetails: React.FC<OrderNormalDetailsProps> = ({
           })
         }
       />
+      <OrderRefundDialog
+        open={params.action === "add-refund"}
+        onClose={closeModal}
+        // TODO: Add redirect to refund view
+        onConfirm={() => null}
```
poulch, line 466: question: do you have any reason why didn't implemented it now

src/components/RadioTile/RadioTile.tsx
```diff
@@ -0,0 +1,94 @@
...
+export const RadioTile = ({
+  checked,
+  title,
+  description,
+  value,
+}: RadioTileProps) => {
+  const [isHoverState, setHoverState] = React.useState(false);
+
+  return (
+    <RadixRadioGroup.Item value={value} asChild>
```
poulch, line 23: question: how could I know that I have to wrapp this component into `<RadixRadioGroup.Root>`, Would it be better to create `RadioTiles` component that will be wrapping all children and `RadioTile` will be it command component like this

```
const RadioTilesBase = ({children}) => {
    return (
        <RadixRadioGroup.Root>
            {children}
      </RadixRadioGroup.Root>,
    )
}

export const RadioTiles = Object.assign(RadioTilesBase, {
    RadioTile
})
```

poulch (changes requested):

package.json
Droniu: I can't use RadioGroup from macaw-ui because it doesn't expose `<label>`.  Styling this part is a requirement for good UX in this feature. I've tried exposing Indicator (https://github.com/saleor/macaw-ui/pull/768) and using both RadioGroup and Indicator from macaw-ui, but custom implementation of RadioGroup.Item in the dashboard - this however didn't work due to the radix-ui context issue explained here -> https://github.com/radix-ui/primitives/issues/2273

I will try to remove the `radix-ui/react-radio-group` from the dashboard - import should work anyway because it is installed in macaw-ui.

src/orders/components/OrderRefundDatagrid/OrderRefundDatagrid.tsx
Droniu: We have no designs for this case, but I will try to come up with something

src/orders/views/OrderDetails/OrderNormalDetails/index.tsx
Droniu: Yes - refund view is not implemented (it's going to be done next in merx-55). We want to avoid big PR-s.

src/components/RadioTile/RadioTile.tsx
Droniu: That's a fair point. I'll update this component

poulch (approved):

poulch (approved):

andrzejewsky (approved):

.changeset/many-vans-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": minor
+---
+
+Implement refund add modal
```
Cloud11PL, line 5: Suggestion: Let's be more descriptive. `Now you can...` etc.

Cloud11PL (approved):

PR #4774, opened 2024-04-03
Manual order transaction refund

poulch:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!


src/orders/queries.ts
```diff
@@ -232,6 +232,22 @@ export const orderRefundData = gql`
   }
 `;
 
+export const orderTransationsData = gql`
+  query OrderTransationsData($orderId: ID!) {
+    order(id: $orderId) {
+      id
+      transactions {
+        ...TransactionItem
```
andrzejewsky, line 240: Do you use all of these fields? if not, please limit it to the ones you consume

src/orders/views/OrderManualTransationRefund/OrderManualTransationRefund.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const notify = useNotifier();
+  const navigate = useNavigator();
+  const intl = useIntl();
+
+  const { data, loading } = useOrderTransationsDataQuery({
+    displayLoader: true,
+    variables: {
+      orderId,
+    },
+  });
```
andrzejewsky, line 36: why do you need to load transactions here? does the entire view depend on it? seems like it is only one component down the component tree, so it can be perhaps colocated?

src/orders/views/OrderManualTransationRefund/OrderManualTransationRefund.tsx
```diff
@@ -0,0 +1,81 @@
...
+          });
+        } else {
+          notify({
+            status: "success",
+            text: intl.formatMessage(transactionRequestMessages.success),
+          });
+          navigate(orderUrl(orderId));
+        }
+      },
+    });
```
andrzejewsky, line 57: The view component isn't the one that should handle refund calls. The actual refund logic happens down the component tree, why do you need to create it here and pass it down? It looks like form handling while you render "Page"

andrzejewsky (changes requested):

andrzejewsky (approved):

src/orders/components/OrderManualTransationRefundPage/OrderManualTransationRefundPage.tsx
```diff
@@ -0,0 +1,78 @@
...
+  currency: string;
+}
+
+export const OrderManualTransationRefundPage = ({
+  orderId,
+  transactions,
+  currency,
+  loading,
+}: OrderManualTransationRefundProps) => {
+  const hasTransactionsToRefund = transactions.length > 0;
```
Droniu, line 28: In this case we probably shouldn't allow to enter this page

src/orders/views/OrderManualTransationRefund/OrderManualTransationRefund.tsx
```diff
@@ -0,0 +1,31 @@
...
+    displayLoader: true,
+    variables: {
+      orderId,
+    },
+  });
+
+  return (
+    <OrderManualTransationRefundPage
+      orderId={data?.order?.id ?? ""}
+      transactions={filterRefundTransactions(data?.order?.transactions ?? [])}
```
Droniu, line 24: I think transactions that are non-refundable should be displayed as disabled, instead of filtered out

Droniu (approved):

andrzejewsky (approved):

src/orders/views/OrderManualTransationRefund/OrderManualTransationRefund.tsx
poulch: I'll do this in separate PR

PR #4821, opened 2024-04-24
Always show charge and authorised amount in payment section of order details

poulch:
AC:
- User should see Authorized, Captured and Cancelled amounts in the Payment summary section when order with a transaction
- User should see pending payments amount only when have amount greater than 0
- User should see "Mark as paid" when the order does not have a charge and authorized amount, can not be refunded and it does not have a gift card, otherwise, we show Authorized, Captured and Cancelled amount

## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!

## [Optional] Description

Screenshot 2024-04-26 at 14 06 54


Droniu (dismissed):

Droniu (dismissed):

andrzejewsky (changes requested):
If the payment section now will show all payments summary, aggregated from all transactions in given order, we should inform the user about that:
- We can call section "Payments summery" instead of "Payment"
- We can add small text right below the title that explains what it presents eg. "A summary of all payments from registered transactions"

What do you think?

Droniu (approved):

witoszekdev (approved):

andrzejewsky (approved):

PR #4846, opened 2024-05-08
Show categories' parents in product view 

Cloud11PL:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

CleanShot 2024-05-13 at 10 23 33

CleanShot 2024-05-13 at 10 24 01

Selected category with ancestors:
CleanShot 2024-05-13 at 10 24 32

## Have you written tests?
- [ ] Yes!


src/products/components/ProductOrganization/ProductOrganization.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+      expect(screen.getByText("Shoes")).toBeInTheDocument();
+      expect(screen.getByText("Sneakers")).toBeInTheDocument();
+      expect(screen.getByText("Running")).toBeInTheDocument();
+    });
+  });
+  it("renders combobox with dropdown list with multiple ancestors", async () => {
+    // Arrange & Act
+    render(
+      <Router>
+        <ThemeProvider>
```
poulch, line 143: suggestion: create wrapper for this

src/searches/useCategorySearch.ts
```diff
@@ -14,6 +14,14 @@ export const searchCategories = gql`
         node {
           id
           name
+          ancestors(first: $first) {
```
poulch, line 17: question: what in case you will have more categories?

src/products/utils/utils.tsx
```diff
@@ -0,0 +1,33 @@
...
+type ChoiceWithAncestors = Choice & {
+  ancestors: {
+    edges: {
+      node: Choice;
+    }[];
+  };
+};
+
+export const getChoicesWithAncestors = (choices: ChoiceWithAncestors[]) =>
+  choices.map(category => {
```
poulch, line 19: suggestion: missing return type

src/searches/useCategorySearch.ts
Cloud11PL: I wouldn't worry about it now, it's a first iteration. We pass 10 here so it should be fine for most cases. If there are 10+ categories then the direct parents will not appear.

mirekm:
Parents should also be rendered once selected.
subcats

src/searches/useCategorySearch.ts
andrzejewsky, line 17: you pass here the same variable as for the root query (categories), which is wrong, this number should be either configured separately or even hardcoded so now the number of loaded categories indicates also amount of ancestors, which is not really expected here: having eg first 2 categories I still want eg. to have 3 ancestors (not 2)

**Consider:**
- keep number of ancestors as separate variable
- search categories query is used in several places, do you need ancestors in each? perhaps a separated query is more reasonable here? (think about it if you have time)

andrzejewsky (changes requested):

Cloud11PL:
@mirekm We've decided to reverse the order to "Grandparent / parent / *category*"

andrzejewsky (dismissed):

andrzejewsky (approved):

andrzejewsky (approved):

src/products/components/ProductOrganization/ProductOrganization.tsx
```diff
@@ -156,6 +159,47 @@ export const ProductOrganization: React.FC<ProductOrganizationProps> = props =>
...
+            startAdornment={val => {
+              if (categoryInputActive || disabled) {
+                return undefined;
+              }
+
+              let availableCategories = categories;
+
+              if (selectedProductCategory) {
+                availableCategories = [...categories, selectedProductCategory];
+              }
```
witoszekdev, line 183: suggestion: I think we should avoid using `let` if it can be solved with a ternary operator

```suggestion
              const availableCategories = selectedProductCategory ? [...categories, selectedProductCategory] : categories;
```

witoszekdev (changes requested):

src/products/components/ProductOrganization/ProductOrganization.tsx
Cloud11PL: you're right, thanks 😸

witoszekdev (approved):

andrzejewsky (approved):

PR #4866, opened 2024-05-18
Migrate order details refund datagrid to GridTable

Droniu:
## What type of PR is this?
- [ ] 💅 Refactor
- [x] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes MERX-359

This PR replaced the datagrid with new GridTable.

image

## Have you written tests?
- [x] Yes!

## [Optional] Description
Enable this feature with:
```
localStorage.setItem('FF_improved_refunds', '{"enabled":true,"payload":"default"}')
```
Go to order with at least one transaction, and create a new refund


poulch (approved):

src/hooks/useOverflowDetection/useOverflowDetection.ts
```diff
@@ -0,0 +1,26 @@
+import { useCallback, useRef } from "react";
+
+export const useOverflowDetection = <T extends HTMLElement>() => {
+  const elementsRef = useRef<T[]>([]);
+
+  const setRef = useCallback(
```
witoszekdev, line 6: suggestion: I got confused that `setRef` is a function that you can pass to `ref` in React, so maybe we should use some other name like, e.g. `getRefForIndex`?

```suggestion
  const getRefForIndex = useCallback(
```

witoszekdev (approved):

witoszekdev:
@Droniu I can't really test this on the preview deployment, because all orders are using old Payments API, and I cannot create a new order due to an error:

```
Something went wrong
Cannot mix operators with other filter inputs.
```

src/orders/components/OrderTransactionRefundPage/components/OrderTransactionRefundTable/RefundTableProductCell.tsx
```diff
@@ -9,25 +10,15 @@ interface RefundTableProductCellProps {
...
-    const productOverflow = productElement.scrollWidth > productElement.clientWidth;
-    const variantOverflow = variantElement.scrollWidth > variantElement.clientWidth;
-
-    return productOverflow || variantOverflow ? undefined : false;
-  };
+  const isAnyOverflowing = isProductOverflowing(index) || isVariantOverflowing(index);
 
   return (
-    <Tooltip open={isOverflowing(index)}>
+    <Tooltip open={isAnyOverflowing ? undefined : false}>
```
andrzejewsky, line 21: open is boolean so why not: open={isAnyOverflowing} ?

src/orders/components/OrderDetailsRefundTable/OrderDetailsRefundTable.tsx
```diff
@@ -0,0 +1,158 @@
...
+          <GridTable.Col __width="10%" />
+          <GridTable.Col __width="25%" />
+          <GridTable.Col __width="1%" />
+          <GridTable.Col __width="20%" />
+          <GridTable.Col __width="1%" />
+        </GridTable.Colgroup>
+        {mergedRefunds?.map((refund, index) => {
+          const isEditable = isRefundEditable(refund);
+
+          return (
```
andrzejewsky, line 82: suggestion: Could you create a dedicated component that renders a single row?

src/hooks/useOverflowDetection/useOverflowDetection.ts
```diff
@@ -0,0 +1,26 @@
+import { useCallback, useRef } from "react";
+
+export const useOverflowDetection = <T extends HTMLElement>() => {
+  const elementsRef = useRef<T[]>([]);
```
andrzejewsky, line 4: suggestion: I don't think you need to keep an array of all references that can have overflow. You should use your hook within the component that can reach overflow - in your case the component for a single row

andrzejewsky (approved):

src/hooks/useOverflowDetection/useOverflowDetection.ts
andrzejewsky: additionally: it can even lead to memory leaks, when I use it on a paginated table where each row has expandable text, this hook will remember every single entry

src/orders/components/OrderTransactionRefundPage/components/OrderTransactionRefundTable/RefundTableProductCell.tsx
Droniu: Setting open to `true` makes the tooltip always visible, even when not hovering. `undefined` displays it only on hover and `false` disables it

witoszekdev (approved):

PR #4879, opened 2024-05-23
Fix validation on manual refund form

Droniu:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [x] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes MERX-367

This PR:
- fixes validation on amount field - it wasn't possible to put positive values lower than 1
- adds maximum amount - form is now blocked if value exceeds selected transaction maxium amount
- fixes a typo in multiple filenames (`Transation` -> `Transaction`)

image

## Have you written tests?
- [x] Yes!


Cloud11PL (approved):

Cloud11PL (dismissed):

src/orders/components/OrderManualTransactionRefundPage/components/OrderManualTransactionRefundForm/OrderManualTransactionRefundForm.tsx
```diff
@@ -18,27 +19,42 @@ import { useIntl } from "react-intl";
...
+  const selectedTransactionId = methods.watch("transationId");
+  const selectedTransaction = transactions.find(
+    transaction => transaction.id === selectedTransactionId,
+  );
+
+  const chargedAmount = selectedTransaction?.chargedAmount?.amount || 0;
+
+  React.useEffect(() => {
+    setChargedAmountForValidation(chargedAmount);
+  }, [chargedAmount]);
```
poulch, line 56: suggestion: there is no reason too keep this in state as it's computed value

src/orders/components/OrderManualTransactionRefundPage/components/OrderManualTransactionRefundTiles/OrderManualTransactionRefundTiles.tsx
```diff
@@ -52,15 +53,27 @@ export const OrderManualTransationRefundTiles = ({
...
                           >
-                            {transaction?.name || "Transaction"}
-                          </Text>
+                            <Text
+                              size={5}
+                              fontWeight="medium"
+                              padding={4}
+                              color={isRefundable ? "default1" : "defaultDisabled"}
+                            >
+                              {transaction?.name || "Transaction"}
```
poulch, line 73: question: `"Transaction"` should't be translated?

poulch (dismissed):

src/orders/components/OrderManualTransactionRefundPage/components/OrderManualTransactionRefundForm/OrderManualTransactionRefundForm.tsx
Droniu: Yes, but it is computed based on `watch(transactionId)`, which comes from `useForm`. And `useForm` needs this value because there is no method to dynamically set the resolver.

src/orders/components/OrderManualTransactionRefundPage/components/OrderManualTransactionRefundTiles/OrderManualTransactionRefundTiles.tsx
Droniu: It was like that before but you're right I will add it to translations

src/orders/components/OrderManualTransactionRefundPage/components/OrderManualTransactionRefundForm/OrderManualTransactionRefundForm.tsx
andrzejewsky: +1 this is a computed value why do you need useState for it? why not just use chargeAmount?

andrzejewsky (changes requested):

src/orders/components/OrderManualTransactionRefundPage/components/OrderManualTransactionRefundForm/OrderManualTransactionRefundForm.tsx
andrzejewsky: I think you can solve this by doing refine on the entire schema as you actually have dependent field: amount depends on transactionId

andrzejewsky (approved):

Cloud11PL (approved):

PR #4903, opened 2024-05-28
Fix improper display of errors on create/edit refund views

Droniu:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [x] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes MERX-505, MERX-504

## Have you written tests?
- [ ] Yes!


src/orders/views/OrderTransactionRefundEdit/handlers.ts
```diff
@@ -0,0 +1,61 @@
...
+      errorLines.push({
+        code: err.code,
+        field: err.field,
+        lines: err.addLines,
+        message: err.message,
+      } as OrderTransactionRefundError);
+
+      setLinesErrors(errorLines);
+    });
+  }
```
poulch, line 60: question: seems like total duplication  of what we have here: `src/orders/views/OrderTransactionRefundCreate/handlers.ts`. Could we wrap this into a hook and share this code?

Droniu: There are slight differences, e.g.:
`submitData.orderGrantRefundCreate?.errors` vs
`submitData.orderGrantRefundUpdate?.errors`

or
`OrderGrantRefundUpdateErrorCode` vs `OrderGrantRefundCreateErrorCode`

behavior for completed mutation is different (order navigation vs clearing errors). I believe wrapping this into single abstracted hook will not make this any more readable and it's going to be much easier to modify mutation handling on specific view if we keep them separate.

src/orders/views/OrderTransactionRefundCreate/handlers.ts
```diff
@@ -49,9 +49,14 @@ export const handleRefundCreateComplete = ({
...
 
-  if (errors.length > 1) {
+  if (errors.length > 0) {
     errors.forEach((err: OrderGrantRefundCreateErrorFragment) => {
-      if (err.code !== OrderGrantRefundCreateErrorCode.REQUIRED) {
+      if (
+        ![
+          OrderGrantRefundCreateErrorCode.REQUIRED,
+          OrderGrantRefundCreateErrorCode.AMOUNT_GREATER_THAN_AVAILABLE,
+        ].includes(err.code)
```
andrzejewsky, line 58: suggestion: function or at least variable

src/orders/views/OrderTransactionRefundEdit/handlers.ts
```diff
@@ -0,0 +1,61 @@
...
+    return;
+  }
+
+  if (errors.length > 0) {
+    errors.forEach((err: OrderGrantRefundUpdateErrorFragment) => {
+      if (
+        ![
+          OrderGrantRefundUpdateErrorCode.REQUIRED,
+          OrderGrantRefundUpdateErrorCode.AMOUNT_GREATER_THAN_AVAILABLE,
+        ].includes(err.code)
```
andrzejewsky, line 43: issue: you have repeated the same piece of code, why don't re-use it? it's rather simple and these errors will be handled mostly in the same way so there is no reason to break DRY

andrzejewsky (approved):

poulch (approved):

PR #4982, opened 2024-06-21
Close GraphiQL Playground on escape button

Cloud11PL:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!

### DIFF
--- a/src/components/DevModePanel/useDevModeKeyTrigger.ts
+++ b/src/components/DevModePanel/useDevModeKeyTrigger.ts
@@ -18,6 +18,7 @@ export const useDevModeKeyTrigger = (handleOpen: HandleOpen, handleClose: Handle
     };
 
     document.addEventListener("keydown", keyDownHandler);
+    // GraphiQL Playground stops propagation of keydown event for Escape key
     document.addEventListener("keyup", keyUpHandler);
 
     return () => {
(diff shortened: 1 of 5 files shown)


src/components/DevModePanel/useDevModeKeyTrigger.ts
```diff
@@ -1,17 +1,28 @@
...
+    };
 
-    return () => document.removeEventListener("keydown", handler);
-  }, [callbackHandler]);
+    document.addEventListener("keydown", keyDownHandler);
+    document.addEventListener("keyup", keyUpHandler);
+
+    return () => {
+      document.removeEventListener("keydown", keyDownHandler);
+      document.removeEventListener("keyup", keyUpHandler);
```
poulch, line 25: question: what is the reason to use separate triggers to open and close, IMO it could be the same. I know that in most cases no one notice that close fires when you release the esc key but for some people it might be confusing.

poulch (dismissed):

src/components/DevModePanel/useDevModeKeyTrigger.ts
Cloud11PL: @poulch surprisingly `keydown` event is not fired for Escape key because the playground stops propagation of this event. I'll add a comment

Droniu (approved):

andrzejewsky (approved):

Droniu (approved):

poulch (approved):

PR #4983, opened 2024-06-21
Fix unhandler error when storing password credentials 

poulch:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!

### DIFF
--- a/.changeset/tricky-lamps-flash.md
+++ b/.changeset/tricky-lamps-flash.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+You should not see credential error when running playwright tests
--- a/.changeset/tricky-lamps-flash.md
+++ b/.changeset/tricky-lamps-flash.md
@@ -2,4 +2,4 @@
 "saleor-dashboard": patch
 ---
 
-You should not see credential error when running playwright tests
+An error is no longer raised and visible to the user when saving credential is not supported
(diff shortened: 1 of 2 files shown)


Cloud11PL (approved):

src/utils/credentialsManagement.ts
```diff
@@ -31,21 +31,21 @@ export async function checkIfCredentialsExist() {
...
 
   if (isSupported) {
-    const cred = new PasswordCredential({
-      id: user.email,
-      name: user.firstName ? `${user.firstName} ${user.lastName}` : undefined,
-      password,
-    });
-
     try {
-      result = navigator.credentials.store(cred);
```
andrzejewsky, line 48: Why not cleaning this a bit:

```
if (!isSupported) return null

try {
  const cred = new PasswordCredential({
    id: user.email,
    name: user.firstName ? `${user.firstName} ${user.lastName}` : undefined,
    password,
  });
  
  return navigator.credentials.store(cred);
} catch (e) {
  return null
}

```

simpler and you eliminate the ifs

.changeset/tricky-lamps-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+You should not see credential error when running playwright tests
```
andrzejewsky, line 5: it's rather: _An error is no longer raised and visible to the user when saving credential is not supported_

andrzejewsky (dismissed):

Cloud11PL (approved):

andrzejewsky (approved):

PR #5009, opened 2024-06-28
Improve local search + improve Navigator Search UI

Cloud11PL:
* Replaces `fuzzaldrin` with `fuse.js`
* Ports Navigator to new macaw-ui

## What type of PR is this?
- [x] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [x] 🔩 Maintenance

- closes #

Search:
CleanShot 2024-07-01 at 12 15 25
CleanShot 2024-07-01 at 12 15 58

Navigator Search:
CleanShot 2024-07-01 at 12 14 26

## Have you written tests?
- [ ] Yes!


src/components/NavigatorSearch/modes/commands/actions.test.ts
```diff
@@ -0,0 +1,36 @@
...
+jest.mock("react-intl", () => {
+  const reactIntl = jest.requireActual("react-intl");
+  const intl = reactIntl.createIntl({
+    locale: "en",
+  });
+
+  return {
+    ...reactIntl,
+    useIntl: () => intl,
+  };
```
poulch, line 14: suggestion: using actual react-intl does not have sense because it doesn't work without context

src/components/NavigatorSearch/modes/utils.ts
```diff
@@ -50,3 +48,54 @@ export function getMode(command: string): QuickSearchMode {
...
+                  id: "BooQvo",
+                  defaultMessage: "Type {key} to see available actions",
+                  description: "navigator placeholder",
+                },
+                {
+                  key: "'?'",
+                },
+              )
+            : null;
+};
```
poulch, line 89: suggestion: please refactor this not to use nested ternaries

src/components/NavigatorSearch/modes/utils.ts
```diff
@@ -50,3 +48,54 @@ export function getMode(command: string): QuickSearchMode {
...
+  return mode === "orders"
+    ? "#"
+    : mode === "customers"
+      ? "@"
+      : mode === "catalog"
+        ? "$"
+        : mode === "help"
+          ? "?"
+          : ">";
+};
```
poulch, line 101: suggestion: same here

src/components/NavigatorSearch/useQuickSearch.ts
```diff
@@ -28,7 +28,7 @@ function useQuickSearch(open: boolean, input: RefObject<HTMLInputElement>): UseQ
   const { result: customers, search: searchCustomers } = useCustomerSearch({
     variables: {
       ...DEFAULT_INITIAL_SEARCH_DATA,
-      first: 5,
+      first: 10,
```
poulch, line 31: question: is there any reason to change this to 10?

src/misc.ts
```diff
@@ -633,3 +634,20 @@ const getAllRemovedRowsBeforeRowIndex = (rowIndex: number, removedRowsIndexs: nu
 
 export const getDatagridRowDataIndex = (rowIndex: number, removedRowsIndexs: number[]) =>
   rowIndex + getAllRemovedRowsBeforeRowIndex(rowIndex, removedRowsIndexs).length;
+
+export const fuzzySearch = <T>(array: T[], query: string | undefined, keys: string[]) => {
```
poulch, line 638:
```suggestion
export const fuzzySearch = <T>({array, query, keys}: { array: T[], query: string | undefined, keys: string[] }) => {
```

src/misc.ts
```diff
@@ -633,3 +634,20 @@ const getAllRemovedRowsBeforeRowIndex = (rowIndex: number, removedRowsIndexs: nu
...
+
+  const fuse = new Fuse(array, {
+    keys,
+    includeScore: true,
+    threshold: 0.3,
+  });
+
+  return fuse
+    .search(query.toLocaleLowerCase())
+    .filter(({ score }) => (query ? score !== 1 : 0))
```
poulch, line 651: question: do you have to check `query` if you have a condition checking this above?
Previous solution use a threshold, maybe we should also consider using it.

poulch (dismissed):

src/components/NavigatorSearch/modes/utils.ts
```diff
@@ -50,3 +48,54 @@ export function getMode(command: string): QuickSearchMode {
...
+                },
+                {
+                  key: "'?'",
+                },
+              )
+            : null;
+};
+
+export const getModeSymbol = (mode: QuickSearchMode) => {
+  return mode === "orders"
```
andrzejewsky, line 92: suggestion: please refactor this not to use nested ternaries

src/components/NavigatorSearch/modes/utils.ts
```diff
@@ -50,3 +48,54 @@ export function getMode(command: string): QuickSearchMode {
       return null;
   }
 }
+
+export const getModePlaceholder = (mode: QuickSearchMode, intl: IntlShape) => {
+  return mode === "orders"
```
andrzejewsky, line 53: suggestion: please refactor this not to use nested ternaries

.changeset/perfect-crews-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Searching for countries and other items is now more efficient, making it easier to find what you need. Additionally, the Dashboard Navigator UI has been improved to match the rest of the application, providing a more consistent experience.
```
andrzejewsky, line 5: what does mean "improved", "consistent", "easier to find" - how does that impact the user? If this is just migration, meaning technical stuff / maintenance, you can prepare more developer-related descriptions.

_The "fuzzaldrin" is no longer used within the codebase, this means we have migrated to "Fuse.js"_

andrzejewsky (changes requested):

.changeset/perfect-crews-speak.md
Cloud11PL: > what does mean "improved", "consistent", "easier to find"
I'm not sure I understand the confusion. From the user perspective search now works better and gives relevant results. I mentioned country search specifically as the link PR from community was fixing that particular search. 

> If this is just migration, meaning technical stuff / maintenance, you can prepare more developer-related descriptions.
According to our changeset guide we have to be user-focused and focused on user workflow. Why does a user have to care about a migration of a library? I wrote what user can expect when this change is merged and released.

src/components/NavigatorSearch/useQuickSearch.ts
Cloud11PL: I thought that 5 results were too little for a search like that.

src/components/NavigatorSearch/modes/commands/actions.test.ts
poulch: Nope it doesn't, it works in your case because you mocked useIntl, but you could do this like that:
```suggestion
jest.mock("react-intl", () => ({
  useIntl: jest.fn(() => ({
    formatMessage: jest.fn(x => x.defaultMessage),
  })),
}));
```

src/misc.ts
```diff
@@ -633,3 +634,20 @@ const getAllRemovedRowsBeforeRowIndex = (rowIndex: number, removedRowsIndexs: nu
...
+
+  const fuse = new Fuse(array, {
+    keys,
+    includeScore: true,
+    threshold: 0.3,
+  });
+
+  return fuse
+    .search(query.toLocaleLowerCase())
+    .filter(({ score }) => score !== 1)
```
poulch, line 651: question: Previuse solution uses a threshold, don't we need to support it too?

src/components/NavigatorSearch/modes/commands/actions.test.ts
Cloud11PL: I meant that it does work 😄

src/misc.ts
Cloud11PL: This solution also uses threshold of 0.3 - the results get rounded down so they can be up to 0.4 (0.39..) so in theory we don't need this filter here. I'll remove it

poulch (approved):

src/components/NavigatorSearch/modes/commands/actions.test.ts
andrzejewsky: @Cloud11PL test should mock dependencies, not mock and provide their actual implementation. You are not testing react-intl here, you test only your code and the only you care about is that your code calls intl with proper arguments.

andrzejewsky (approved):

PR #5064, opened 2024-07-24
Fix required permissions in subscription query editor

Cloud11PL:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

CleanShot 2024-07-25 at 12 29 35
## Have you written tests?
- [ ] Yes!


krzysztofzuraw (approved):

src/custom-apps/components/PermissionAlert/utils.ts
```diff
@@ -1,47 +1,32 @@
 // @ts-strict-ignore
```
poulch, line 1: suggestion: If you rearrange a whole file would be good to remove it and fix types accordingly to rules

poulch (dismissed):

src/custom-apps/components/PermissionAlert/utils.ts
```diff
@@ -1,47 +1,53 @@
...
-      types {
-        kind
-        name
-        description
-        fields(includeDeprecated: false) {
-          name
-          description
-        }
-      }
+const tokenTree = (token: Token | undefined, tokens: string[] = []): string[] => {
```
poulch, line 32: question: What does this function do? IMO it is not a token tree it is more like flattened token tree

src/custom-apps/components/PermissionAlert/utils.ts
```diff
@@ -1,47 +1,53 @@
...
+  if (token && token.next) {
+    if (typeof token.value === "string") {
+      tokens.push(token.value);
     }
```
poulch, line 36: question: shouldn't you push token value no matter if token has next?

src/custom-apps/components/PermissionAlert/utils.ts
```diff
@@ -50,92 +56,119 @@ const uniq = <T>(value: T, index: number, self: T[]) => self.indexOf(value) ===
...
+
+    const _permissions = Object.keys(descriptions).reduce((acc, key) => {
+      const { isOneOfRequired, permissions } = extractPermissions(descriptions[key]);
+
+      if (permissions.length > 0) {
+        acc[key] = { isOneOfRequired, permissions };
+      }
+
+      return acc;
+    }, {} as SubscriptionQueryPermission);
```
poulch, line 166: suggestion: That could be also be wrapped into separate function and `extractPermissionsFromQuery ` would look like this:

```suggestion
const extractPermissionsFromQuery = (
  query: string,
  introspectionQuery: IntrospectionQuery,
): SubscriptionQueryPermission => {
    try {
      const schema = buildClientSchema(introspectionQuery);
      const descriptions = getDescriptionsFromQuery(query, schema);
      
      return getAllPermission(descriptions) // You name it better 
    } catch (error) {
      // explicit silent
    }
}
```

poulch (approved):

src/custom-apps/components/PermissionAlert/utils.ts
Cloud11PL: Doesn't matter here that much, I did that to filter out empty values

poulch (approved):

krzysztofzuraw (approved):

PR #5102, opened 2024-08-08
Fix errors when user has insufficient permissions

Cloud11PL:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!


andrzejewsky (changes requested):
It does not work: 
Screenshot 2024-08-08 at 10 27 59

When I click for the first item from the gift card list.

+ Please add some tests, especially for the new code you created.

src/giftCards/GiftCardUpdate/GiftCardHistory/hooks/useGiftCardHistoryEvents.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+      },
+    },
+  };
+});
+
+describe("useGiftCardHistoryEvents", () => {
+  it("should return gift card events", () => {
+    // Arrange
+    const wrapper = ({ children }: { children: React.ReactNode }) => (
+      <MockedProvider>
```
andrzejewsky, line 53: You mock "entire" app here just for testing your hook and the logic inside of it. You can simply mock "useContext" and control what returns, then this test will be a way faster. Same with the other hooks used there

src/giftCards/GiftCardUpdate/GiftCardHistory/hooks/useGiftCardHistoryEvents.ts
```diff
@@ -1,13 +1,40 @@
...
+  events: NonNullable<GiftCardEventsQuery["giftCard"]>["events"] | null | undefined;
+}
+
+const useGiftCardHistoryEvents = (): GiftCardHistoryEvents => {
+  const { user } = useUser();
+
+  const canSeeApp =
+    user?.userPermissions?.some(permission => APP_PERMISSIONS.includes(permission.code)) || false;
+  const canSeeUser =
+    user?.userPermissions?.some(permission => USER_PERMISSIONS.includes(permission.code)) || false;
```
andrzejewsky, line 21: it can be a single function, that checks permission agains given array of permissions

src/giftCards/GiftCardUpdate/GiftCardHistory/hooks/useGiftCardHistoryEvents.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    };
+  }
+
+  return {
+    data: {
+      giftCard: {
+        events: [{ id: "event1" }, { id: "event2" }],
+      },
+    },
+  };
```
andrzejewsky, line 46: this code actually should be tested as well - you provide here **your own implementation** for the hook that comes from library (read the comments below regarding using wrappers)

andrzejewsky (changes requested):

src/giftCards/GiftCardUpdate/GiftCardHistory/hooks/useGiftCardHistoryEvents.test.tsx
Cloud11PL: Turns out this provider is not needed here.

.changeset/nice-humans-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+You can now view gift card details page if you have only gift card permissions and customer details page if you have only customer permissions.
```
witoszekdev, line 5: question: What was previous behavior? Could we also add that to the changeset?

witoszekdev (dismissed):

.changeset/nice-humans-add.md
Cloud11PL: Done ✅

poulch (approved):

andrzejewsky (approved):

PR #5121, opened 2024-08-20
Improve Stability of Authentication Setup Tests

michalina-graczyk:
## What type of PR is this?
- [x] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!
- [x] No... here is why: _Writing tests are mandatory, please replace this text with why test are not included in this PR_

## [Optional] Description

This PR addresses issues with flaky Playwright authentication tests, especially related to undefined cookie values.

#### Key Changes

1. **Safe Property Access**
   - Added optional chaining to prevent crashes when cookies are missing.

2. **Retry Mechanism**
   - Implemented retries (up to 3 times) for the authentication process to handle intermittent failures.

3. **Improved Logging**
   - Enhanced logging for better insight into failures, especially when cookies are missing or API errors occur.

### DIFF
--- a/playwright/tests/auth.setup.ts
+++ b/playwright/tests/auth.setup.ts
@@ -13,22 +13,50 @@ const authenticateAndSaveState = async (
   filePath: string,
 ) => {
   const basicApiService = new BasicApiService(request);
+  let retries = 0;
+  const maxRetries = 3;
 
-  await basicApiService.logInUserViaApi({ email, password });
+  while (retries < maxRetries) {
+    try {
+      const loginResponse = await basicApiService.logInUserViaApi({ email, password });
 
-  const loginJsonInfo = await request.storageState();
+      if (loginResponse.data.tokenCreate?.errors?.length > 0) {
+        loginResponse.data.tokenCreate.errors.forEach(error => {
+          console.error(`Login error (code: ${error.code}): ${error.message}`);
+        });
+      } else {
+        const loginJsonInfo = await request.storageState();
 
-  loginJsonInfo.origins.push({
-    origin: process.env.BASE_URL!,
-    localStorage: [
-      {
-        name: "_saleorRefreshToken",
-        value: loginJsonInfo.cookies[0].value,
-      },
-    ],
-  });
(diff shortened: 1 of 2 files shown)


michalina-graczyk:
After applying the changes the test provided the following output:

```
Login error (code: LOGIN_ATTEMPT_DELAYED): Logging has been suspended till 2024-08-20 20:04:37.030798+00:00 due to too many logging attempts originating from the same IP address.
Authentication failed, retrying 1/3...
Login error (code: LOGIN_ATTEMPT_DELAYED): Logging has been suspended till 2024-08-20 20:04:37.929046+00:00 due to too many logging attempts originating from the same IP address.
Authentication failed, retrying 2/3...
Login error (code: LOGIN_ATTEMPT_DELAYED): Logging has been suspended till 2024-08-20 20:04:38.807281+00:00 due to too many logging attempts originating from the same IP address.
Authentication failed, retrying 3/3...
```

Can we whitelist the servers from which these auth attempts are performed during the CI phase?

playwright/tests/auth.setup.ts
```diff
@@ -6,29 +6,64 @@ import path from "path";
...
   password: string,
   filePath: string,
 ) => {
   const basicApiService = new BasicApiService(request);
+  let retries = 0;
+  const maxRetries = 3;
+  const delayBetweenRetries = 3000;
 
-  await basicApiService.logInUserViaApi({ email, password });
+  while (retries < maxRetries) {
```
andrzejewsky, line 22: we can't login in a loop until API will pass - I don't think that fixes anything:

PW have nice guides how to do login stuff, in our case we need to login and share sessions across the workers: https://playwright.dev/docs/auth#moderate-one-account-per-parallel-worker

andrzejewsky: We should login accordingly to the API limits, not fighting with them. When we reach limitations that means either doing something wrong or API has a bug

andrzejewsky (changes requested):

PR #5141, opened 2024-09-02
✨ Add label to run e2e on demand ✨

Cloud11PL:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance
- [x] 🛠 Workflow CI/CD changes

- closes #
https://linear.app/saleor/issue/MERX-766/add-label-to-run-e2e-to-be-not-run-on-every-commit

This PR introduces a label-based mechanism to trigger Playwright end-to-end (E2E) tests. Instead of running E2E tests on every commit, they will now only be executed when the pull request is labeled with run pw-e2e.

This approach aims to optimize our CI pipeline by reducing unnecessary test runs and allowing more control over when E2E tests are executed, ensuring they are triggered only when needed.

## How to use?
Apply the label run pw-e2e to the pull request when you want the E2E tests to run.
Without the label, E2E tests will be skipped.

## Have you written tests?
- [ ] Yes!

## [Optional] Description
Key Changes:
- Updated the GitHub Actions workflow to conditionally run Playwright E2E tests based on the presence of the run pw-e2e label.
- Added conditional logic to skip E2E tests unless the label is present.
- Ensured that the merge of E2E test reports only occurs if the label is applied and tests have run.


NyanKiyoshi (approved):

.changeset/kind-weeks-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Add e2e label to be not run on every commit
```
andrzejewsky, line 5: _E2e tests are no longer executed on every commit, this means performing e2e tests happens only when you add label "run pw-e2e"_

.github/workflows/pr-automation.yml
```diff
@@ -174,7 +176,8 @@ jobs:
           PW_RETRIES: ${{ vars.PW_RETRIES }}
 
   merge-reports:
-    if: "!cancelled()"
+    if: "!cancelled() && contains(github.event.pull_request.labels.*.name, 'run pw-e2e')"
```
andrzejewsky, line 179: is it needed? there is "needs" that points to "run-tests", so the constraint to run this job if "run-tests" is finished is already here

andrzejewsky (changes requested):

.github/workflows/pr-automation.yml
michalina-graczyk: Fixed, you're right.

.github/workflows/pr-automation.yml
```diff
@@ -174,7 +176,8 @@ jobs:
           PW_RETRIES: ${{ vars.PW_RETRIES }}
 
   merge-reports:
-    if: "!cancelled()"
+    if: !cancelled()
```
Cloud11PL, line 179:
```suggestion
    if: "!cancelled()"
```

michalina-graczyk (approved):

andrzejewsky (approved):

NyanKiyoshi (approved):

PR #5176, opened 2024-09-23
[Draft] 🎀 Add login on demand depending on user permission 🎀 

michalina-graczyk:
## What type of PR is this?
- [x] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!

### DIFF
--- a/playwright/utils/auth.ts
+++ b/playwright/utils/auth.ts
@@ -16,7 +16,7 @@ export const getStorageState = async (permission: UserPermission | "admin"): Pro
   const tempDir = path.join(__dirname, "../.auth");
   const storageStatePath = path.join(tempDir, `${permission}.json`);
 
-  process.stdout.write("getStorageState attempt");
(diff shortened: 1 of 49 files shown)


playwright/utils/auth.ts
```diff
@@ -0,0 +1,71 @@
...
+    await basicApiService.logInUserViaApi({ email, password });
+
+    const loginJsonInfo = await apiRequestContext.storageState();
+
+    loginJsonInfo.origins.push({
+      origin: process.env.BASE_URL!,
+      localStorage: [
+        {
+          name: "_saleorRefreshToken",
+          value: loginJsonInfo.cookies[0].value,
```
Cloud11PL, line 44: Since this line of code fails very often I would check if there's an error message in `loginJsonInfo` and throw it before so that it's easier to debug later

PR #5236, opened 2024-11-05
🎀  Increase global single test timeout 🎀 

michalina-graczyk:
This test passed locally and actually on CI too, the only error message was Test timeout of 30000ms exceeded. It may mean that 30s per test might be too short, so I increase them

## What type of PR is this?
- [x] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #
- https://linear.app/saleor/issue/MERX-1218/flaky-saleor-217

## Have you written tests?
- [ ] Yes!

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -36,7 +36,7 @@ export default defineConfig({
     : [["html"], ["list"]],
   expect: { timeout: 10 * 1000 },
   maxFailures: 10,
-  timeout: 30 * 1000,
+  timeout: 35 * 1000,
   use: {
     baseURL: env.BASE_URL || "",
     trace: env.CI ? "retain-on-failure" : "on",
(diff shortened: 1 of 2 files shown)


playwright.config.ts
```diff
@@ -36,7 +36,7 @@ export default defineConfig({
     : [["html"], ["list"]],
   expect: { timeout: 10 * 1000 },
   maxFailures: 10,
-  timeout: 30 * 1000,
+  timeout: 35 * 1000,
```
Cloud11PL, line 39: Suggestion: I think we should break down "long" tests into smaller tests if that's possible.

Cloud11PL (approved):

playwright.config.ts
michalina-graczyk: Totally agree, let's first stabilize them and then refactor the 'long' one

andrzejewsky (approved):

szczecha (approved):

PR #5246, opened 2024-11-13
Fix anchor in datagrid-based lists

Cloud11PL:
* Makes anchor render in correct place allowing user o middle click or right click on the anchor 
* Since `onClick` of the `<a />` anchor is now executed instead of Datagrid's `onRowClick` then it adds missing `navigatorOpts` to Datagrid lists allowing them to preserve state of the previous location so that the backlink redirects to list with pagination and filters 
* Cleans up backlink logic

## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [x] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!
- [x] No... here is why: Test here would have to be directly on Datagrid checking behaviour on hover (if anchor is rendered or not) - as datagrid is a canvas it is not possible.


poulch (approved):
LGTM, one think that I've noticed is that anchor right click does not work on select channel column in product datagrid

src/attributes/components/AttributeListDatagrid/AttributeListDatagrid.tsx
```diff
@@ -120,6 +119,7 @@ export const AttributeListDatagrid = ({
             onToggle={handlers.onToggle}
           />
         )}
+        navigatorOpts={{ state: getPrevLocationState(location) }}
```
andrzejewsky, line 122: if this is repeated across each view with the gridd, can just grid call `useLocation` inside of its component?

andrzejewsky (approved):

PR #5255, opened 2024-11-15
Improve test functionality - SALEOR_215

michalina-graczyk:
## What type of PR is this?
- [x] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [x] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!


playwright/tests/orders.spec.ts
```diff
@@ -342,46 +342,58 @@ for (const refund of orderRefunds) {
...
+
+test(`TC: SALEOR_215 Inline discount is applied in a draft order @draft @discounts @e2e`, async () => {
   test.slow();
 
+  const calculateDiscountedPrice = (
+    undiscountedPrice: number,
+    discountPercentage: number,
+  ): number => {
+    return undiscountedPrice - (undiscountedPrice * discountPercentage) / 100;
+  };
```
andrzejewsky, line 354: This is dangerous: you just create a code to test the code. It implements the way discount is being calculated, meaning the test now is aware of it while it should not. Everywhere when possible make assertions on real values, because the incoming data is always known (you know what prices given product has because it's test data)

Non, blocking for now, but please keep in mind that.

.changeset/violet-laws-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Improve inline discount test precision and reliability in draft orders
```
andrzejewsky, line 5: Please write changeset where you clearly explain what you have improved. Something like

_Now, test that checks discount calculation asserts against fixed precision (2 decimals)_

andrzejewsky (changes requested):
OK, fix only the changeset

andrzejewsky (dismissed):

szczecha (dismissed):

szczecha (approved):

andrzejewsky (dismissed):

andrzejewsky (approved):

PR #5261, opened 2024-11-19
Fix a login error when using a password manager

Cloud11PL:
It prevents 2 concurrent login attempts that sometimes happen when using a password manager (eg. 1Password). It caused not fully handled error regarding login attempt delay due to too many attempts from one IP. It also adds an error message for that case. 

## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [x] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [x] Yes!
- [ ] No... here is why:


src/auth/AuthProvider.test.tsx
```diff
@@ -211,4 +211,49 @@ describe("AuthProvider", () => {
     expect(hook.result.current.errors).toEqual(["noPermissionsError"]);
     expect(hook.result.current.authenticated).toBe(false);
   });
+
+  it("should handle concurrent login attempts correctly", async () => {
+    const intl = useIntl();
+    const notify = useNotifier();
+    const apolloClient = useApolloClient();
```
poulch, line 218: suggestion: IMO you should mock them

poulch (approved):

src/auth/AuthProvider.test.tsx
Cloud11PL: I just checked, they're already mocked in the file

andrzejewsky (approved):

PR #5271, opened 2024-11-20
Home page information tiles 

Cloud11PL:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

CleanShot 2024-11-20 at 15 34 38

## Have you written tests?
- [ ] Yes!


poulch (approved):

andrzejewsky (approved):

src/newHome/homeInfoTiles/tileData.tsx
```diff
@@ -0,0 +1,206 @@
...
+const StarIcon = () => <SVG src={star} className={noShrink} />;
+const ExtensionIcon = () => <SVG src={extension} className={noShrink} />;
+const BagIcon = () => <SVG src={bag} className={noShrink} />;
+
+// Bottom actions
+const GitHubIcon = () => <SVG src={github} className={noShrink} />;
+const DiscordIcon = () => <SVG src={discord} className={noShrink} />;
+const ExternalLinkIcon = () => <SVG src={externalLink} className={noShrink} />;
+
+export const getTilesData = ({ intl }: { intl: IntlShape }): InfoTile[] => [
```
andrzejewsky, line 36: just a comment here - this force on you a specific structure for each tile, while you don't have eg. hundreds of them where you would need pre-defined structure. You have only few items and for that case simple composition would work better

andrzejewsky: Imagine you need to change just a single tile now, in the way that not really fits this structure - it would need to probably opt-out from this

Cloud11PL: I definitely understand the concern but in this case we had only tiles that had the exact same structure. Changing it to a more composable solution shouldn't take long *if* we decide that we're going to need it.

src/newHome/homeInfoTiles/HomeInfoTile.tsx
```diff
@@ -0,0 +1,36 @@
...
+  content: React.ReactNode;
+  bottomActions: React.ReactNode;
+};
+
+export const HomeInfoTile: React.FC<InfoTile> = ({ id, bottomActions, content, header }) => (
+  <DashboardCard
+    backgroundColor="default1"
+    borderStyle="solid"
+    borderColor="default1"
+    borderRadius={4}
```
poulch, line 16: question: Shuldn't border-radius be `3`which is `6px;` which is ok with figma?

poulch (dismissed):

andrzejewsky (approved):

PR #5280, opened 2024-11-25
Store onboarding state in user metadata

poulch:
To restart onboarding state you have to remove metadata, using grapiQL:
1.  Go to you_env_url/graphql 
2. Put this mutation:
```
mutation clearMeMetadata($id: ID!, $keys: [String!]!) {
  deleteMetadata(id: $id, keys: $keys) {
    errors {
      field
      message
    }
  }
}
```
3. Add variables:
```
{
  "id": YOUR_USER_ID,
  "keys": ["onboarding"]
}
```
4. Add a headers:
```
{
   authorization-bearer: YOUR_ACCESS_TOKEN
}
```

You can take your id from this query:
```
query GetID {
  me {
    id
  }
}
```

## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!


src/newHome/homeOnboarding/onboardingContext/useOnboardingStorage.ts
```diff
@@ -0,0 +1,85 @@
...
+        return;
+      }
+
+      try {
+        const userMetadata: Array<MetadataInput> = [
+          ...(user?.metadata?.map(data => ({
+            key: data.key,
+            value: data.value,
+          })) ?? []),
+        ];
```
Cloud11PL, line 45:
```suggestion
const userMetadata: MetadataInput[] = user?.metadata 
  ? user.metadata.map(data => ({
      key: data.key,
      value: data.value
    }))
  : [];
```

src/newHome/homeOnboarding/onboardingContext/useOnboardingStorage.ts
```diff
@@ -0,0 +1,85 @@
...
+        } else {
+          userMetadata.push({
+            key: METADATA_KEY,
+            value: metadataValue,
+          });
+        }
+
+        await updateMetadata({
+          variables: {
+            id: user?.id ?? "",
```
Cloud11PL, line 63: Suggestion: We're already returning early if user is not present so `user.id` should be fine here I think

Cloud11PL (dismissed):

Cloud11PL (approved):

src/newHome/mutations.ts
```diff
@@ -0,0 +1,11 @@
+import { gql } from "@apollo/client";
+
+export const updateUserMetadata = gql`
+  mutation UpdateUserMetadata($id: ID!, $input: [MetadataInput!]!) {
```
andrzejewsky, line 4: this is used only for onboarding so you can call it eg "SaveOnBoardingState", update metadata sounds like generic solution that can be used anywhere

src/newHome/homeOnboarding/onboardingContext/useOnboardingStorage.ts
```diff
@@ -0,0 +1,83 @@
...
+          userMetadata[metadataIndex] = {
+            key: METADATA_KEY,
+            value: metadataValue,
+          };
+        } else {
+          userMetadata.push({
+            key: METADATA_KEY,
+            value: metadataValue,
+          });
+        }
```
andrzejewsky, line 58: preparing metadata to be saved can be a separated function

src/newHome/homeOnboarding/onboardingContext/useOnboardingStorage.ts
```diff
@@ -0,0 +1,83 @@
...
+      if (!metadata) {
+        return undefined;
+      }
+
+      return JSON.parse(metadata.value);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.warn("Could not get onboarding state from metadata", { error });
+
+      return undefined;
```
andrzejewsky, line 29: what happens if this occur? do you have some default state?

andrzejewsky (approved):

src/newHome/homeOnboarding/onboardingContext/useOnboardingStorage.ts
poulch: nothing happened as state is kept in react state hook so the user refresh the page state will be restarted to last one saved in metadata or initial one when never was saved in metadata

PR #5324, opened 2024-12-17
Add explicit waits for draft order shipping carrier button interaction

michalina-graczyk:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!


playwright/pages/draftOrdersPage.ts
```diff
@@ -60,6 +60,14 @@ export class DraftOrdersPage extends BasePage {
   }
 
   async clickAddShippingCarrierButton() {
+    await this.page.waitForLoadState("networkidle");
```
andrzejewsky, line 63: https://playwright.dev/docs/api/class-page#page-wait-for-load-state

Especially:
> Most of the time, this method is not needed because Playwright [auto-waits before every action](https://playwright.dev/docs/actionability).

and:
> 'networkidle' - DISCOURAGED wait until there are no network connections for at least 500 ms. Don't use this method for testing, rely on web assertions to assess readiness instead.

playwright/pages/draftOrdersPage.ts
```diff
@@ -60,6 +60,14 @@ export class DraftOrdersPage extends BasePage {
...
+
+    await this.addShippingCarrierLink.waitFor({
+      state: "visible",
+      timeout: 10000,
```
andrzejewsky, line 67: Do not add hardcoded timeout, rather we should rethink the test if it stucks somwhere.

andrzejewsky (changes requested):

szczecha (approved):

andrzejewsky (approved):

PR #5335, opened 2025-01-07
Fix identify users in analytics

poulch:
## What type of PR is this?
- [ ] 💅 Refactor
- [ ] 🌟 Feature
- [ ] 🔥 Bug Fix
- [ ] 🔩 Maintenance

- closes #

## Have you written tests?
- [ ] Yes!


src/auth/hooks/useAuthProvider.ts
```diff
@@ -85,8 +83,6 @@ export function useAuthProvider({ intl, notify, apolloClient }: UseAuthProviderO
     }
   };
   const handleLogout = async () => {
-    analytics.reset();
```
krzysztofzuraw, line 88: **Question:** how it will work now? We won't reset user if they are logging out?

poulch: We don't use our user_id to identify users so we rely on an auto-generated ID by Posthog, so when we run reset on logout and user login again, a new ID was generated.
For now, we will have one ID per device and only change the domain and email domain by setting user properties.

Cloud11PL (dismissed):

PR #5342, opened 2025-01-09
Exclude posthog traffic

andrzejewsky:
## Scope of the change

- Currently we push events to the posthog from various of our environments even if we don't need that (traffic is excluded in posthog). 
- This PR disables that and allows to specify excluded domains on releases


Cloud11PL (approved):

.github/workflows/deploy-staging-and-prepare-release.yaml
```diff
@@ -61,6 +61,7 @@ jobs:
       IS_CLOUD_INSTANCE: true
       POSTHOG_KEY: ${{ secrets.POSTHOG_KEY }}
       POSTHOG_HOST: ${{ secrets.POSTHOG_HOST }}
```
NyanKiyoshi, line 63: I'm not sure why we removed the secret from other places but not this one? Is it only needed for this one or it's a leftover?

(I'm assuming it's a frontend/client-side secret and not backend secret as well?)

poulch (approved):

NyanKiyoshi (approved):

PR #5356, opened 2025-01-16
PR tests reporting

szczecha:
## Scope of the change

I want to have test reports available in PRs and measure flaky tests
This PR adds a new PW reporter https://github.com/ctrf-io/playwright-ctrf-json-reporter
and publish test summary https://github.com/ctrf-io/github-test-reporter to PRs

Next, we will collect data form PRs and enable option for historical data: 
fail-rate-report
flaky-rate-report
previous-results-report
and add reporting for other workflows

### DIFF
--- a/.github/actions/flaky-tests-report/action.yml
+++ b/.github/actions/flaky-tests-report/action.yml
@@ -20,7 +20,7 @@ runs:
 
     - name: Merge CTRF reports
       shell: bash
-      run: npx ctrf merge ./ctrf --output ctrf/merged-report.json
+      run: npx ctrf merge ./ctrf --output merged-report.json
 
     - name: Upload CTRF report
       uses: actions/upload-artifact@v4
@@ -33,6 +33,9 @@ runs:
       uses: ctrf-io/github-test-reporter@v1
       with:
         report-path: ctrf/*.json
+        summary-report: true
+        failed-folded-report: true
+        flaky-report: true
       env:
         GITHUB_TOKEN: ${{ inputs.github_token }}
       if: always()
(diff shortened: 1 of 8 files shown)


.github/actions/flaky-tests-report/action.yml
```diff
@@ -0,0 +1,47 @@
...
+    - name: Download CTRF reports
+      uses: actions/download-artifact@v4
+      with:
+        pattern: ctrf-report-*
+        path: ctrf-reports
+        merge-multiple: true
+        
+    - name: Merge CTRF reports
+      shell: bash
+      run: npx ctrf merge ./ctrf-reports --output merged-report.json
```
michalina-graczyk, line 23: Shouldn't we consider adding error handling or conditional checks in case of failure?
If this command fails (e.g., due to a missing ctrf-reports directory or invalid output location), the workflow could break without proper feedback 🤔

szczecha: @michalina-graczyk I think no need for extra error handling, the job is already failing if there are no files to merge, or merge report wasn't created here is an example. So as developer you will see that not all checks passed.
image

karola312:
What if test if flaky? Now we have info only about test case status in merge reports, before I saw a "flaky" field

szczecha:
> What if test if flaky? Now we have info only about test case status in merge reports, before I saw a "flaky" field

@karola312 I disabled it for now because when we post failed/flaky or all results we have tags in the test name with @ so each time we notify someone for our PRs which isn't a great vibe. We need to change that and then we can enable more options 
Here is an example test report where we called additional people who has GitHub nick same as our test tags:
https://github.com/saleor/saleor-dashboard/actions/runs/12868849980

witoszekdev (approved):

karola312 (approved):

PR #5359, opened 2025-01-17
Bump react to 18

poulch:


.eslintrc.json
```diff
@@ -91,6 +91,7 @@
     ],
     // Decided to turn off:
     "react/prop-types": "off",
+    "react/react-in-jsx-scope": "off",
```
witoszekdev, line 94: suggestion: according to [react eslint plugin docs](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md), we should instead extend `react/jsx-runtime` config:

> If you are using the new JSX transform from React 17, you should disable this rule by extending react/jsx-runtime in your eslint config (add "plugin:react/jsx-runtime" to "extends").

package-lock.json
```diff
@@ -30,7 +30,7 @@
         "@material-ui/styles": "^4.11.4",
         "@reach/auto-id": "^0.16.0",
         "@saleor/macaw-ui": "npm:@saleor/macaw-ui@0.7.4",
-        "@saleor/macaw-ui-next": "npm:@saleor/macaw-ui@1.1.15",
+        "@saleor/macaw-ui-next": "npm:@saleor/macaw-ui@0.0.0-pr-20250127234423",
```
witoszekdev, line 33: issue: is this a mistake? if we have unreleased changes we should probably release new macaw version

src/apps/components/InstalledAppListRow/InstalledAppListRow.test.tsx
```diff
@@ -4,13 +4,14 @@ import { InstalledApp } from "@dashboard/apps/types";
...
 import { render, screen } from "@testing-library/react";
-import React from "react";
 import { MemoryRouter as Router } from "react-router-dom";
 
 import InstalledAppListRow from "./InstalledAppListRow";
 
 const Component = ({ data, context }: { data: InstalledApp; context: AppListContextValues }) => (
   <Wrapper>
+    {/* eslint-disable-next-line @typescript-eslint/ban-ts-comment */}
+    {/* @ts-ignore */}
```
witoszekdev, line 14: suggestion: we should use `@ts-expect-error` instead of `@ts-ignore`, because if for some reason we don't have TypeScript error anymore it will be shown as error so that we can remove this comment. You also don't need to disable ESLint rule for that (probably)

src/channels/components/AssignmentList/AssignmentList.tsx
```diff
@@ -52,6 +51,8 @@ const AssignmentList: React.FC<AssignmentListProps> = props => {
             <Skeleton className={classes.skeleton} />
           ) : (
             <>
+              {/* eslint-disable-next-line @typescript-eslint/ban-ts-comment */}
+              {/* @ts-ignore */}
```
witoszekdev, line 55: suggestion: same here let's use `@ts-expect-error`

src/channels/components/AssignmentList/Item.tsx
```diff
@@ -22,6 +21,8 @@ const Item = SortableElement(({ item, sortable = false, onDelete }: ItemProps) =
       <div className={classes.container}>
         <div className={classes.containerContent}>
           {sortable && (
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
```
witoszekdev, line 25: `@ts-expect-error`

src/components/AddressEdit/AddressEdit.tsx
```diff
@@ -6,7 +6,7 @@ import { getFormErrors } from "@dashboard/utils/errors";
 import { TextField } from "@material-ui/core";
 import { makeStyles } from "@saleor/macaw-ui";
 import { Option } from "@saleor/macaw-ui-next";
-import React from "react";
+import * as React from "react";
```
witoszekdev, line 9: question: why do we need this import, can we get rid of it?

witoszekdev (changes requested):
I think we should:
- get rid of `ts-ignore` in favor of `ts-expect-error` + add description why it's needed?
- don't use `import * as React from "react";` since we're importing only selected functions in rest of our codebase
- replace `ReactChildren` type with `ReactNode` if it makes sense - I don't think we should limit what children should be passed if we're not explicitly requiring them to be React components
- replace macaw with stable version
- change that ESLint config as they suggested in docs

package-lock.json
poulch: Now it's now, when all test passed I release macaw and bump version to correct one

src/components/VisibilityCard/VisibilityCard.tsx
```diff
@@ -212,7 +212,7 @@ export const VisibilityCard: React.FC<VisibilityCardProps> = props => {
                     },
                   })
                 }
-                error={getFieldError(errors, "isPublishedAt")}
+                error={getFieldError(errors, "isPublishedAt") as any}
```
Cloud11PL, line 215: Suggestion: I would add a `TODO` or try to type it but it's not necessary 😸

Cloud11PL (commented):
Overall LGTM 👍, I made some comments. I would check if `Icons` need to have `undefined` passed in props. Unusually I'll keep this review as a comment because then I want to test this manually :)

src/components/VisibilityCard/VisibilityCard.tsx
witoszekdev: +1 IMHO using `@ts-expect-error` is better because it forces you to remove the label once the error is fixed ;)

src/apps/components/InstalledAppListRow/InstalledAppListRow.test.tsx
```diff
@@ -15,6 +14,7 @@ jest.mock("@dashboard/featureFlags");
 
 const Component = ({ data, context }: { data: InstalledApp; context: AppListContextValues }) => (
   <Wrapper>
+    {/* @ts-expect-error  Router types does not have explicit children props  */}
```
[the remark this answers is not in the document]
poulch: I hope so :D

Cloud11PL (approved):

PR #5381, opened 2025-01-30
Fix open with cmd key

poulch:


src/components/Datagrid/Datagrid.tsx
```diff
@@ -530,16 +530,16 @@ export const Datagrid: React.FC<DatagridProps> = ({
           aria-hidden={true}
           onWheelCapture={hideLinkAndShowAfterDelay}
           onClick={e => {
```
andrzejewsky, line 532: move this handler to the separate function and a write test where you pass event and you make assertion what url is being passed to the "navigate" + add comment in the test that navitate uses browser history which handles /dashbaord/ vs /

andrzejewsky (approved):

Cloud11PL (approved):

PR #5389, opened 2025-02-04
New collection filters

Cloud11PL:
## Scope of the change

image


witoszekdev:
I haven't made a code review, but just testing it out I found some issues:

## 1. Only one filter type can be used at a time

I cannot search in between values, for example: Price greater than 1, but lower than 5 even though this is possible via API:

```graphql
query {
  products(channel: "default-channel", where: {price: {range: {gte: 1, lte: 5}}}, first: 5) {
    edges {
      node {
        id
      }
    }
  }
}
```

Input:

CleanShot 2025-02-06 at 15 17 40@2x

Query made:

CleanShot 2025-02-06 at 15 17 36@2x

## 2. Multi channel search

API doesn't allow providing more than one channel, but this is possible via UI:

CleanShot 2025-02-06 at 15 13 48@2x

## 3. UI improvement suggestions

- I think using labels `lower` or `greater` is incorrect, since API uses `lte` (lower than or equal) and `gte` (greater than or equal) - maybe we should just re-use the same description from Saleor API or use mathematical signs? E.g.: `≤`

- There's no tooltip when I didn't input my query correctly, in this example I need to provide channel for search and button is disabled, but I got confused why:

CleanShot 2025-02-06 at 15 11 21@2x

- Greyed out input - I think this is visually confusing when you need to enter required input for channel and it's greyed-out, it suggests that the input is disabled not that it's required :/

witoszekdev (dismissed):

src/components/ConditionalFilter/API/CollectionFilterAPIProvider.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const index = parseInt(position, 10);
+    const filterElement = getFilterElement(value, index);
+
+    const handler = createAPIHandler(filterElement, client, inputValue, intl);
+
+    return handler.fetch();
+  };
+  const fetchLeftOptions = async () => {
+    return [];
+  };
```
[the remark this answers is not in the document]
Cloud11PL: Fetch left in this context means that we can use 'dynamic' options from Saleor API. We use it in Products where you can filter by custom attributes: 
CleanShot 2025-02-06 at 18 41 54

Fetch right gets the values either from API (for example channels) or from e.g. an enum.

locale/defaultMessages.json
```diff
@@ -4938,6 +4938,9 @@
     "context": "tax classes card header",
     "string": "General information"
   },
+  "ThUvIL": {
+    "string": "Hidden"
```
witoszekdev, line 4942: suggestion: Let's add context for these strings, e.g.:

```suggestion
    "context": "Collection filter condition",
    "string": "Hidden"
```

locale/defaultMessages.json
```diff
@@ -9262,6 +9265,9 @@
   "w7jT4W": {
     "string": "No channels selected"
   },
+  "w7tf2z": {
+    "string": "Published"
```
witoszekdev, line 9269: suggestion: maybe we should use `yes/no` label values for this? When you look at the condition as a filter, then it looks incorrect:

CleanShot 2025-02-12 at 16 43 49@2x

`Published is Published` 😅

witoszekdev (approved):

locale/defaultMessages.json
witoszekdev: or maybe we can change the field name to `Visibility`, so that it becomes:

`Visibility is Published`
`Visibility is Hidden`

andrzejewsky (approved):

PR #5398, opened 2025-02-07
Retry clicking on dropdown and wait for first item 

szczecha:
## Scope of the change

fix flaky SALEOR_198
I use https://playwright.dev/docs/test-assertions#expecttopass to make sure that clicking at dropdown shows list of options


playwright/pages/dialogs/addNavigationMenuItemDialog.ts
```diff
@@ -15,8 +15,11 @@ export class AddNavigationMenuItemDialog extends BasePage {
   }
 
   async selectLinkTypeOption(linkType: string) {
-    await this.menuLinkType.click();
-    await this.waitForDOMToFullyLoad();
+    await expect(async () => {
+      await this.menuLinkType.click();
+      await this.waitForDOMToFullyLoad();
```
poulch, line 20: suggestion: Do you need it, it only works on first-page loading. Someone who does not know how it works puts it everywhere, and no one knows why.

szczecha: @poulch  I removed all `waitForDOMToFullyLoad` from this file and used `expect.toPass` also for selecting the value

andrzejewsky (approved):

PR #5410, opened 2025-02-14
Add reporting for workflow "Run test manually"

szczecha:
## Scope of the change

I duplicated the `Run test manually` workflow as it's hard to test changes from PR. 
The new `Run test manually` contains additional fields for the test report:
```

DASHBOARD_VERSION: ${{ steps.set_variables_mode.outputs.MODE }}
BRANCH_NAME: ${{ github.ref}}
SALEOR_CLOUD_SERVICE: ${{ needs.initialize-cloud.outputs.SALEOR_CLOUD_SERVICE }}

```
 and running flaky test report after it
```
      - name: Create flaky report
        uses: ./.github/actions/flaky-tests-report
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}

```

I added extra output fields from `prepare-api-variables/action.yml` and `prepare-tests-variables/action.yml` to be able to use it later in new workflow.


karola312 (approved):

.github/workflows/run-test-manual-new.yml
```diff
@@ -0,0 +1,186 @@
+name: NEW Run test manually 
+# This workflow is for testing only and will be disabled and removed after testing
```
witoszekdev, line 2: question: so we are not going to keep this and remove it in next PR?

witoszekdev (approved):

.github/workflows/run-test-manual-new.yml
szczecha: > question: so we are not going to keep this and remove it in next PR?

@witoszekdev If it works as I expect I will replace the old workflow for `Run test manually` with this new one in the next PR.

PR #5416, opened 2025-02-17
Alerts for apps with issues or disabled apps in app list

Cloud11PL:
## Scope of the change

CleanShot 2025-02-21 at 12 25 22

CleanShot 2025-02-18 at 09 44 04


witoszekdev (dismissed):

andrzejewsky (changes requested):

src/apps/components/AppAlerts/utils.test.ts
```diff
@@ -0,0 +1,143 @@
+import { EventDeliveryStatusEnum } from "@dashboard/graphql";
+
+import { getLatestFailedAttemptFromWebhooks } from "./utils";
+
+describe("getLatestFailedAttemptFromWebhook", () => {
+  it("should return the latest failed attempt from failedDelivers", () => {
+    // Arrange
+    const webhook: any = {
```
poulch, line 8: suggestion: avoid using `any` when you have types generated for this, using `as Somthing` is better

src/apps/components/AppAlerts/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+    return moment(fromFailedDelivers?.createdAt).isAfter(moment(fromPendingDelivers?.createdAt))
+      ? fromFailedDelivers
+      : fromPendingDelivers;
+  } else if (fromFailedDelivers) {
+    return fromFailedDelivers;
+  } else if (fromPendingDelivers) {
+    return fromPendingDelivers;
+  } else {
+    return null;
+  }
```
poulch, line 37: suggestion: this is unreadable
try something like this

```suggestion
 if (fromFailedDelivers && fromPendingDelivers) {
    return someCheck()
  } 
  
    if (fromFailedDelivers) {
    return fromFailedDelivers;
  } 
  
    if (fromPendingDelivers) {
    return fromPendingDelivers;
  } 
 
   return null;
```

andrzejewsky, line 37: Return early, you don't need such chain of `else if`

```
 if (fromFailedDelivers && fromPendingDelivers) {
    return moment(fromFailedDelivers?.createdAt).isAfter(moment(fromPendingDelivers?.createdAt))
      ? fromFailedDelivers
      : fromPendingDelivers;
 }

 if (fromFailedDelivers) {
    return fromFailedDelivers;
 }

if (fromPendingDelivers) {
    return fromPendingDelivers;
 }

return null
```

src/apps/components/AppAlerts/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export const getLatestFailedAttemptFromWebhooks = (webhooks: Webhook[]) =>
+  webhooks
+    .map(getLatestFailedAttemptFromWebhook)
+    .filter(Boolean)
+    .sort((a, b) => moment(b?.createdAt).diff(a?.createdAt))[0] ?? null;
```
andrzejewsky, line 44: if you use another `moment` call why `getLatestFailedAttemptFromWebhook` cannot return moment instances?

src/graphql/hooks.generated.ts
```diff
@@ -97,6 +97,44 @@ export const AppPermissionFragmentDoc = gql`
   code
 }
     `;
+export const AppEventDeliveriesFragmentDoc = gql`
+    fragment AppEventDeliveries on App {
+  webhooks @include(if: $PERMISSION_MANAGE_APPS) {
```
andrzejewsky, line 102: these delivers can be fetched ONLY when i have feature flag enabled, not when i have permission to load it. Obviously permission check is necessary, but you need to load it only when it's needed, meaning when feature is active.

andrzejewsky (changes requested):

src/apps/components/AppAlerts/AppRowDisabledAlert.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return (
+    <Tooltip>
+      <Tooltip.Trigger>
+        <Box color="default2" __transform="scale(0.8)">
+          <DisabledIcon />
+        </Box>
+      </Tooltip.Trigger>
+
+      <Tooltip.Content align="start" side="bottom">
+        <StopPropagation>
```
poulch, line 30: question: why do you use `StopPropagation` ?

src/apps/components/AppAlerts/utils.ts
```diff
@@ -0,0 +1,72 @@
+import { AppEventDeliveriesFragment, EventDeliveryStatusEnum } from "@dashboard/graphql";
+import moment from "moment";
+
+export type Webhook = NonNullable<AppEventDeliveriesFragment["webhooks"]>[0];
+
+const hasFailedAttemptsCheck = (webhook: Webhook) =>
+  webhook.failedDelivers && webhook.failedDelivers?.edges?.length > 0;
```
poulch, line 7:
```suggestion
const hasFailedAttemptsCheck = (webhook: Webhook) => webhook?.failedDelivers?.edges?.length > 0;
```

src/apps/components/AppAlerts/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+const hasFailedAttemptsInPendingCheck = (webhook: Webhook) => {
+  const preliminaryCheck = webhook.pendingDelivers && webhook.pendingDelivers?.edges?.length > 0;
```
poulch, line 9:
```suggestion
  const preliminaryCheck = webhook?.pendingDelivers?.edges?.length > 0;
```

poulch (approved):

src/apps/components/AppAlerts/AppRowDisabledAlert.tsx
Cloud11PL: Because the event is transffered to the app list row and when you click on the link inside the tooltip you would get navigated to the app instead of the link.

src/apps/components/AppAlerts/utils.ts
Cloud11PL: Yes, I know it's makes more sense to do what you have proposed, but it seems like it doesn't work with Dashboard's TS version - it produces a TS error (it works with TS 5.7.2)

andrzejewsky (approved):

PR #5423, opened 2025-02-18
Migrate from tabs to filter presets in page types list

poulch:
## Scope of the change

Before:
Screenshot 2025-02-18 at 17 06 44

After:
Screenshot 2025-02-18 at 17 06 33


src/pageTypes/components/PageTypeListPage/PageTypeListPage.tsx
```diff
@@ -36,38 +41,71 @@ const PageTypeListPage: React.FC<PageTypeListPageProps> = ({
...
-          onTabSave={onTabSave}
-        />
+      <DashboardCard gap={0}>
+        <Box paddingX={6} marginBottom={2}>
+          <Box __width="320px">
+            <SearchInput
+              initialSearch={initialSearch}
+              placeholder={intl.formatMessage({
+                id: "umsU70",
+                defaultMessage: "Search Page Type",
```
Cloud11PL, line 102: Suggestion: A teeny tiny nit pick but in line 92 and 80 we don't use capital letters for "page type" but here we do. Maybe we should just keep "Search" here or "Type to search"? Do we have any pattern for this?

Cloud11PL (dismissed):

src/pageTypes/components/PageTypeListPage/PageTypeListPage.tsx
andrzejewsky: We have common pattern across the other views "Search <entity name>" eg. "Search attributes" etc. so we should follow same here "Search page types"

andrzejewsky (changes requested):

Cloud11PL (approved):

andrzejewsky (approved):

PR #5424, opened 2025-02-19
Gift cards: replace legacy alerts with contextual link

poulch:
## Scope of the change
Screenshot 2025-02-19 at 10 09 49


Cloud11PL (approved):

src/components/AppLayout/ContextualLinks/useContextualLink.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+  <IntlProvider locale="EN">{children}</IntlProvider>
+);
+
+describe("useContextualLink", () => {
+  it("should return correct link for staff_members", () => {
+    // Act
+    const { result } = renderHook(() => useContextualLink("staff_members"), { wrapper });
+
+    // Assert
+    expect(result).toMatchSnapshot();
```
andrzejewsky, line 21: Please don't use snapszots, there are several reasons why it is a not good idea, for instance:
- you don't see the assertion, just "match snapshot", so you don't know what you expect
- the generated markup is prone to many changes, often not even related to your test, eg changing class or link, most likely you accept it right away
- quite hard to maintain due to vast markup after some time you won't even remember what you want to "test/match" here 

and more,

There are plenty other ways of how you can test it. Clearly in your test you expect some link to be properly set, this seem to be your main assertion. Just mock the ContextualLineLink and assert if the props were passed properly, that's it

witoszekdev (changes requested):
Please replace snapshot tests 🙇🏻 

other comments are non-blocking suggestions / questions.

andrzejewsky (approved):

Cloud11PL (approved):

witoszekdev (approved):

PR #5425, opened 2025-02-19
Store sidebar dot state in user metadata

Cloud11PL:


src/apps/components/AppAlerts/useAppsFailedDeliveries.ts
```diff
@@ -1,23 +1,39 @@
...
+
+          if (!acc) {
+            return latestFailedAttempt.createdAt;
+          }
+
+          return latestFailedAttempt.createdAt > acc ? latestFailedAttempt : acc;
+        },
+        null as Moment | null,
+      ) ?? null,
     [data],
```
witoszekdev, line 37: suggestion: Let's use nested data for dependencies array, so that we don't re-calculate on unrelated changes

```suggestion
    [data?.apps?.edges],
```

src/apps/components/AppAlerts/useSidebarWebhookAlertMetadata.ts
```diff
@@ -0,0 +1,57 @@
+import { useUser } from "@dashboard/auth";
+import { MetadataInput, useUserAccountUpdateMutation } from "@dashboard/graphql";
+import { useMemo } from "react";
+
+const deliveryAttemptKey = "sidebar_app_webhook_alert_state";
```
witoszekdev, line 5: suggestion: we can export it, or put into separate file so that this string can be re-used for tests

```suggestion
export const DELIVERY_ATTEMPT_KEY = "sidebar_app_webhook_alert_state";
```

witoszekdev (approved):

src/apps/components/AppAlerts/SidebarAppAlert.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+            ],
+          },
+        },
+      },
+    ]);
+
+    // Act
+    await act(async () => {
+      render(<SidebarAppAlert />);
+    });
```
poulch, line 81: why do you wrap render into act? 
```suggestion
      render(<SidebarAppAlert />);
```

src/apps/components/AppAlerts/SidebarAppAlert.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+    const trigger = screen.getByTestId("sidebar-app-alert-trigger");
+
+    fireEvent.focus(trigger);
```
poulch, line 85: Try to use userEvent from `import userEvent from '@testing-library/user-event'
`

src/apps/components/AppAlerts/SidebarAppAlert.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+    await screen.findByRole(/tooltip/);
+
+    // Assert
+    expect(screen.queryAllByText("Issues found.{break}Review app alerts.")[0]).toBeInTheDocument();
```
poulch, line 90: Why not like this
```suggestion
    expect(screen.queryByText("Issues found.{break}Review app alerts.")).toBeInTheDocument();
```

src/apps/components/AppAlerts/useSidebarDotState.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  persist: jest.fn().mockResolvedValue(undefined),
+  webhookAlertState: { value: null },
+};
+
+describe("useSidebarDotState", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("should initialize with dot not visible", () => {
```
poulch, line 18:
```suggestion
  it("should initialize with dot not visible when no value", () => {
```

src/apps/components/AppAlerts/useSidebarDotState.ts
```diff
@@ -0,0 +1,77 @@
...
+  }, [sidebarDotRemoteState]);
+
+  const handleClick = async (clickDate: string) => {
+    try {
+      await persist({
+        lastClickDate: clickDate,
+        lastFailedAttemptDate: lastFailedAttemptDateRef.current ?? "",
+      });
+    } catch (error) {
+      // Silently ignore errors
```
poulch, line 51: suggestion: I'd think about catch this error and send it to sentry, WDYT?

Cloud11PL: Sure, we can give it a try although I'm afraid we would pollute Sentry with junk errors like 401s

src/apps/components/AppAlerts/SidebarAppAlert.test.tsx
Cloud11PL: Because rendering and state setting takes a bit longer and assertions below fail without it.

src/apps/components/AppAlerts/SidebarAppAlert.test.tsx
Cloud11PL: Didn't work for me :/

src/apps/components/AppAlerts/SidebarAppAlert.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+      },
+    ]);
+
+    // Act
+    await act(async () => {
+      render(<SidebarAppAlert />);
+    });
+
+    // Assert
+    expect(screen.queryByTestId("sidebar-app-alert-trigger")).not.toBeInTheDocument();
```
andrzejewsky, line 133: Next time please don't add test-id for component/unit tests, You should match element by something more semantic such as role, element name, text, some attribute etc.

src/apps/components/AppAlerts/useAppsFailedDeliveries.ts
```diff
@@ -1,24 +1,40 @@
...
+            return acc;
+          }
+
+          if (!acc) {
+            return latestFailedAttempt.createdAt;
+          }
+
+          return latestFailedAttempt.createdAt > acc ? latestFailedAttempt : acc;
+        },
+        null as Moment | null,
```
andrzejewsky, line 35: how Moment can be null? this is either be `null` OR `Moment`

andrzejewsky (approved):

src/apps/components/AppAlerts/useAppsFailedDeliveries.ts
Cloud11PL: Right 😅 I'll fix it in the next PR

PR #5427, opened 2025-02-19
[BCK-1457] Add OrderLine metadata update form and ProductVariant metadata

witoszekdev:
## Scope of the change

This PR adds `OrderLine` to "View metadata" modal that's opened on Order details page. This metadata form is editable and was rewritten to use react-hook-form

## Screenshots

CleanShot 2025-02-19 at 16 07 05@2x

CleanShot 2025-02-19 at 16 07 25@2x


src/orders/components/OrderMetadataDialog/test-ids.ts
```diff
@@ -0,0 +1,2 @@
+export const TEST_ID_ORDER_LINE_METADATA = "order-line-metadata";
+export const TEST_ID_PRODUCT_VARIANT_METADATA = "product-variant-metadata";
```
poulch, line 2: question: Do we need to store data-test-id? Is it good practice? I'm asking because first time see something like this?

witoszekdev: I think storing them separately from app's code is a good idea:
- you can import these test-id's in tests, if you need to update them in app, then they don't break your tests
- you can re-use them for different set of tests if you want

I could have exported them from the same file (rather than creating `test-ids.ts`) but that would break hot-module reloading.

src/components/MetadataHookForm/MetadataHookForm.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+jest.mock("./useMetadataFormControls", () => ({
+  useMetadataFormControls: () => ({
+    metadataFields: mockData.metadata,
+    privateMetadataFields: mockData.privateMetadata,
+    handleMetadataChange: jest.fn(),
+    handlePrivateMetadataChange: jest.fn(),
+  }),
+}));
+
+describe("MetadataHookForm", () => {
```
Cloud11PL, line 45: Tests in Dashboard should follow Arrange / Act / Assert pattern

src/components/MetadataHookForm/useMetadataFormControls.ts
```diff
@@ -0,0 +1,89 @@
...
+
+import { EventDataAction, EventDataField } from "../Metadata/types";
+import { getDataKey, parseEventData } from "../Metadata/utils";
+
+const validateMetadata = (metadata: MetadataInput[]): true | string => {
+  const keys = metadata.map(entry => entry.key);
+  const uniqueKeys = new Set(keys);
+
+  if (uniqueKeys.size !== keys.length) {
+    return "Metadata keys must be unique, remove duplicate key";
```
Cloud11PL, line 13: Issue: If these messages are shown to the user then they should be translated

src/fragments/orders.ts
```diff
@@ -154,6 +154,12 @@ export const fragmentOrderLine = gql`
 export const fragmentOrderLineWithMetadata = gql`
   fragment OrderLineWithMetadata on OrderLine {
     ...OrderLine
+    metadata {
+      ...MetadataItem
+    }
+    privateMetadata {
+      ...MetadataItem
+    }
     variant {
```
Cloud11PL, line 163: Suggestion: We could just use `Metadata` fragment from `src/fragments/metadata.ts`

src/orders/components/OrderMetadataDialog/OrderMetadataDialog.tsx
```diff
@@ -1,45 +1,157 @@
...
+  data: OrderMetadataDialogData;
   loading?: boolean;
 }
 
 export const OrderMetadataDialog = ({ onClose, open, data, loading }: OrderMetadataDialogProps) => {
+  const { onSubmit, lastSubmittedData } = useHandleOrderLineMetadataSubmit({ initialData: data });
   const hasManageProducts = useHasManageProductsPermission();
 
+  const formMethods = useForm<MetadataFormData>({
+    // Display last submitted data while re-fetching to avoid flicker on UI
```
Cloud11PL, line 42: ❤️

Cloud11PL (changes requested):

src/fragments/orders.ts
```diff
@@ -154,6 +154,7 @@ export const fragmentOrderLine = gql`
 export const fragmentOrderLineWithMetadata = gql`
   fragment OrderLineWithMetadata on OrderLine {
     ...OrderLine
+    ...Metadata
```
andrzejewsky, line 157: How about loading line-related metadata when i click "view metadata"? - not for now, but it's always a bit of optimization

andrzejewsky (approved):

Cloud11PL (approved):

src/components/MetadataHookForm/useMetadataFormControls.test.ts
```diff
@@ -0,0 +1,330 @@
...
+describe("useMetadataFormControls", () => {
+  it("returns form values from hook form to be used in form", () => {
+    const formMethods = renderHook(() =>
+      useForm<MetadataFormData>({
+        values: {
+          metadata: [{ key: "key", value: "value" }],
+          privateMetadata: [{ key: "private-key", value: "private-value" }],
+        },
+      }),
+    );
```
poulch, line 23: question: Is it necessary to render an external library in the test? Why not mock them?

src/orders/components/OrderMetadataDialog/OrderMetadataDialog.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+describe("OrderMetadataDialog", () => {
+  const onCloseMock = jest.fn();
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("closes when user hits close icon button", () => {
+    render(
+      <MockedProvider>
```
poulch, line 51: suggestion: You could create wrapper component and pass it to each render

poulch (approved):
Tests still missing AAA comments

poulch (approved):
Tests still missing AAA comments

src/components/MetadataHookForm/useMetadataFormControls.test.ts
witoszekdev: I think it's worth leaving it for *this specific test* - we can at least test if we're passing data correctly do react-hook-form and if the metadata values are in fact rendering or not

src/orders/components/OrderMetadataDialog/OrderMetadataDialog.test.tsx
witoszekdev: Actually I just got rid of it altogether, I was supposed to remove it 😅

src/fragments/orders.ts
witoszekdev: Right, it makes sense I'll put it into separate PR 👍🏻

poulch (approved):

andrzejewsky (approved):

PR #5428, opened 2025-02-20
Fix gift card balance refetch

poulch:


src/giftCards/GiftCardUpdate/GiftCardUpdateBalanceDialog/GiftCardUpdateBalanceDialog.tsx
```diff
@@ -36,6 +36,8 @@ const GiftCardUpdateBalanceDialog: React.FC<DialogProps> = ({ open, onClose }) =
     balanceAmount: amount,
   };
   const [updateGiftCardBalance, updateGiftCardBalanceOpts] = useGiftCardUpdateMutation({
+    // TODO: remove when apollo client update to newest version
+    refetchQueries: ["GiftCardDetails"],
```
witoszekdev, line 40: non blocking: I don't think we're going to do that soon though... our first prio would be to upgrade React, I think, so maybe we can figure out to do this better?

Maybe @andrzejewsky can you look at this, and see why Apollo cache does't work properly? 🙏🏻 
The issue here is that `GiftCardDetails` has nested objects that have `__typename: Money` and for some reason first mutation updates values in `useQuery` but subsequent mutations don't... my guess was that Apollo does shallow comparison between cache changes and that's why it doesn't work? Hard to say why it works first time ¯\(°_o)/¯

witoszekdev (dismissed):

Cloud11PL (approved):

PR #5437, opened 2025-02-25
Fetch app alert data at intervals

Cloud11PL:
## Scope of the change
- Fetches the webhook delivery info at regular intervals so that this heavy query is not overused


src/apps/components/AppAlerts/useAppsFailedDeliveries.ts
```diff
@@ -12,36 +11,32 @@ interface AppsFailedDeliveries {
...
+      }, null) ?? null,
     [data?.apps?.edges],
   );
 
   const handleFetchAppsWebhooks = () => {
-    // TODO: checking if webhooks should be fetched will be extracted out of this hook in a separate ticket
     fetchAppsWebhooks({
       variables: {
-        canFetchAppEvents: hasManagedAppsPermission,
+        canFetchAppEvents: true,
```
poulch, line 39: question: Don't you need permission check?

andrzejewsky (dismissed):

src/hooks/useIntervalActionWithState.ts
```diff
@@ -0,0 +1,54 @@
...
+      savedAction.current();
+      setLastInvocation(new Date().getTime());
+    };
+
+    if (hasPassedInterval) {
+      action();
+      timeout.current = setTimeout(() => action(), interval);
+    } else {
+      timeout.current = setTimeout(() => action(), nextDelay);
+    }
```
poulch, line 50:
```suggestion
    if (hasPassedInterval) {
      action();
    }
    
   timeout.current = setTimeout(() => action(), interval);
```

poulch (dismissed):

src/apps/components/AppAlerts/useAppsFailedDeliveries.ts
Cloud11PL: It's checked outside of this hook

src/hooks/useIntervalActionWithState.ts
Cloud11PL: This is not the same, notice one timeout is with `interval` and the other with `nextDelay`

src/hooks/useIntervalActionWithState.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    return cleanup;
+  }, [lastInvocation, action, interval, key, setLastInvocation, skip]);
```
witoszekdev, line 53: issue(?): if we put `action` in `useRef` (I assumed it was to avoid re-renders) why do we put it into dependencies array? This will cause to start new timeout

src/apps/components/AppAlerts/useAppsFailedDeliveries.ts
```diff
@@ -12,36 +11,32 @@ interface AppsFailedDeliveries {
...
       },
```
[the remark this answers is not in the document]
Cloud11PL: a fragment in this query is used in multiple places, since user without `MANAGE_APPS` can fetch app list we need to use graphql's `@include` method that's controlled by this variable

src/hooks/useIntervalActionWithState.ts
Cloud11PL: It seems you're right and I'll fix it soon. It wouldn't be a big deal since new timeout will have calculated delay since last execution of the action

src/apps/components/AppAlerts/useAppsFailedDeliveries.ts
poulch: Maybe would be good to add short comment about this

src/hooks/useIntervalActionWithState.ts
poulch: You are right

poulch (approved):

PR #5448, opened 2025-03-03
Handle re-login in sidebar apps alerts

Cloud11PL:
## Scope of the change

- shows alert again after user logs in again to Dashboard


Cloud11PL:
Turns out this is not needed 😄

PR #5462, opened 2025-03-07
Set constant height for rows

andrzejewsky:
## Scope of the change

Setting constant height for table rows as in some browser it may collapse.


poulch (approved):

Cloud11PL (approved):
Confirmed it works correctly on firefox 👍 
CleanShot 2025-03-07 at 11 24 19

PR #5481, opened 2025-03-14
Extensions explore page

poulch:
## Scope of the change

Create Exolore extensions page with mocked data and pin it to nav 

Screenshot 2025-03-17 at 15 17 59

### DIFF
--- a/.changeset/great-apes-type.md
+++ b/.changeset/great-apes-type.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-You can now view explore extensions page with mocked data
--- a/.changeset/great-apes-type.md
+++ b/.changeset/great-apes-type.md
@@ -2,4 +2,4 @@
 "saleor-dashboard": patch
 ---
 
-You can now view explore extensions page with mocked daa
+You can now view explore extensions page with mocked data
--- a/.changeset/great-apes-type.md
+++ b/.changeset/great-apes-type.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+You can now view explore extensions page with mocked daa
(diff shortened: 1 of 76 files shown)


.changeset/great-apes-type.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+You can now view explore extensions page with mocked daa
```
witoszekdev, line 5:
```suggestion
You can now view explore extensions page with mocked data
```

.changeset/great-apes-type.md
witoszekdev, line 1: I dunno if we want to have this in changelog, since it's not released? 🤔 

what's our approach for stuff like this?

src/extensions/components/ExtenionItem/components/InstalledBadge.tsx
```diff
@@ -0,0 +1,19 @@
+import { SUCCESS_ICON_COLOR } from "@dashboard/colors";
+import { Text } from "@saleor/macaw-ui-next";
+import React from "react";
+import SVG from "react-inlinesvg";
+import { useIntl } from "react-intl";
+
+import CheckIcon from "../../../../../assets/images/check.svg";
+import { messages } from "../../../messages";
```
witoszekdev, line 8: suggestion: these imports are super long, do we have maybe an alias for that?

witoszekdev (approved):

src/extensions/components/ExtenionItem/components/InstalledBadge.tsx
poulch: I prefer to keep all import within module as relative

assets/images/check.svg
witoszekdev, line 1: We'll expose this icon in appstore, so I dunno if we should have it in dashboard - I guess it's fine for now, so that we have mock data

witoszekdev (approved):

PR #5520, opened 2025-04-04
Extensions: installed extensions page with data

poulch:

### DIFF
--- a/src/extensions/queries.ts
+++ b/src/extensions/queries.ts
@@ -1,8 +1,14 @@
 import { gql } from "@apollo/client";
 
 export const installedApps = gql`
-  query InstalledApps($before: String, $after: String, $first: Int, $last: Int) {
-    apps(before: $before, after: $after, first: $first, last: $last) {
+  query InstalledApps(
+    $before: String
+    $after: String
+    $first: Int
+    $last: Int
+    $filter: AppFilterInput
+  ) {
+    apps(before: $before, after: $after, first: $first, last: $last, filter: $filter) {
       pageInfo {
         hasNextPage
         hasPreviousPage
@@ -15,6 +21,39 @@ export const installedApps = gql`
           ...InstalledApp
         }
       }
+    }
+  }
+`;
+
+export const installedAppsList = gql`
+  query InstalledAppsList(
+    $before: String
+    $after: String
+    $first: Int
+    $last: Int
+    $filter: AppFilterInput
+  ) {
+    apps(before: $before, after: $after, first: $first, last: $last, filter: $filter) {
+      pageInfo {
+        hasNextPage
+        hasPreviousPage
+        startCursor
+        endCursor
+      }
+      totalCount
+      edges {
+        node {
+          ...InstalledAppDetails
+        }
+      }
+    }
+  }
+`;
+
+export const appsInstallations = gql`
+  query AppsInstallations {
+    appsInstallations {
+      ...AppInstallation
     }
   }
 `;
--- a/src/extensions/components/InstalledExtensionsList/InstalledExtensionsList.tsx
+++ b/src/extensions/components/InstalledExtensionsList/InstalledExtensionsList.tsx
@@ -2,15 +2,31 @@ import { GridTable } from "@dashboard/components/GridTable";
 import { ExtensionAvatar } from "@dashboard/extensions/components/ExtensionAvatar";
 import { messages } from "@dashboard/extensions/messages";
 import { InstalledExtension } from "@dashboard/extensions/types";
-import { Box, Text } from "@saleor/macaw-ui-next";
+import { Box, Skeleton, Text } from "@saleor/macaw-ui-next";
 import React from "react";
 import { FormattedMessage } from "react-intl";
 
 interface InstalledExtensionsListProps {
   installedExtensions: InstalledExtension[];
+  loading: boolean;
 }
 
-export const InstalledExtensionsList = ({ installedExtensions }: InstalledExtensionsListProps) => {
+export const InstalledExtensionsList = ({
+  installedExtensions,
+  loading,
+}: InstalledExtensionsListProps) => {
+  if (loading) {
+    return (
+      <Box display="grid" gap={3}>
+        <Skeleton __width="50%" />
+        <Skeleton __width="100%" />
+        <Skeleton __width="50%" />
+        <Skeleton __width="100%" />
+        <Skeleton __width="50%" />
+      </Box>
+    );
+  }
+
   return (
     <Box __marginLeft="-24px" __marginRight="-24px">
       <GridTable>
--- a/src/extensions/hooks/useInstalledExtensionsData.ts
+++ b/src/extensions/hooks/useInstalledExtensionsData.ts
@@ -1,0 +1,39 @@
+import {
+  AppTypeEnum,
+  useAppsInstallationsQuery,
+  useInstalledAppsListQuery,
+} from "@dashboard/graphql";
+import { mapEdgesToItems } from "@dashboard/utils/maps";
+
+export const useInstalledExtensionsData = () => {
+  const {
+    data: appsData,
+    loading: appsLoading,
+    error: appsError,
+  } = useInstalledAppsListQuery({
+    variables: {
+      first: 100,
+      filter: {
+        type: AppTypeEnum.THIRDPARTY,
+      },
+      after: null,
+    },
+  });
+
+  const apps = mapEdgesToItems(appsData?.apps) || [];
+
+  const {
+    data: appInstallationData,
+    loading: appsInstallationLoading,
+    error: appsInstallationError,
+  } = useAppsInstallationsQuery();
+
+  return {
+    data: {
+      apps: apps,
+      appsInstallation: appInstallationData?.appsInstallations ?? [],
+    },
+    loading: appsLoading || appsInstallationLoading,
+    error: appsError || appsInstallationError,
+  };
+};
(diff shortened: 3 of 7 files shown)


Cloud11PL (dismissed):

src/extensions/components/AppDisabledInfo.tsx
```diff
@@ -8,7 +8,15 @@ export const AppDisabledInfo = () => {
...
     <Box display="flex" alignItems="center" gap={1} color="default2">
       <DisabledIcon height="10px" width="10px" />
-      <Text size={2} color="default2">
+      <Text
+        size={2}
+        color="default2"
+        display={{
+          desktop: "block",
+          tablet: "block",
+          mobile: "none",
```
witoszekdev, line 17: question: do we have some other way to show that app is disabled on mobile or we just hide that?

src/extensions/components/FailedWebhookInfo.tsx
```diff
@@ -3,33 +3,42 @@ import Link from "@dashboard/components/Link";
...
 
+const alignVeritically = { verticalAlign: "baseline" };
+
 export const FailedWebhookInfo = ({ date, link }: FailedWebhookInfoProps) => {
   return (
     <Box display="flex" alignItems="center" gap={1} color="default2">
       <Box __color={WARNING_ICON_COLOR} display="flex" alignItems="center" justifyContent="center">
         <ExclamationIcon width="11px" height="11px" />
       </Box>
-      <Box display="flex" alignItems="center" gap={1}>
```
witoszekdev, line 20: suggestion: fyi, I think we can actually still use flex here, you could instead use `alignItems: baseline`

src/extensions/hooks/useActiveAppsInstallations.ts
witoszekdev, line 1: question: can't we just use the same hook from `src/apps/hooks/useActiveAppsInstallations`?

src/extensions/hooks/useInstalledExtensionsData.tsx
```diff
@@ -0,0 +1,116 @@
...
+      },
+    });
+  const eventDeliveries = mapEdgesToItems(eventDeliveriesData?.apps) ?? [];
+
+  useEffect(() => {
+    if (initialLoading && data) {
+      setInitialLoading(false);
+      queryEventDeliveries();
+    }
+  }, [data]);
```
witoszekdev, line 57: question: is there a reason why we can't fetch both event deliveries and apps at the same time? 🤔

src/extensions/hooks/useInstalledExtensionsData.tsx
```diff
@@ -0,0 +1,116 @@
...
+      return (
+        <FailedWebhookInfo
+          link={AppPaths.resolveAppDetailsPath(id)}
+          date={lastFailedAttempt.createdAt}
+        />
+      );
+    }
+
+    return null;
+  };
```
witoszekdev, line 88: suggestion: I think we can safely put this function outside of the scope of this hook (in the same file)

src/extensions/hooks/useInstalledExtensionsData.tsx
```diff
@@ -0,0 +1,116 @@
...
+      logo: brand?.logo?.default ?? "",
+      info: getExtensionInfo({
+        id,
+        isActive,
+        loading: eventDeliveriesLoading,
+        lastFailedAttempt,
+      }),
+      actions: <ViewDetailsActionButton id={id} isDisabled={!isActive} />,
+    };
+  });
```
witoszekdev, line 107: suggestion: I think we can place it in `useMemo` - to re-calculate only when list of filtered apps actually changes, not on every re-render

src/extensions/hooks/useInstalledExtensionsData.tsx
```diff
@@ -0,0 +1,116 @@
...
+    useEventDeliveryLazyQuery({
+      variables: {
+        first: 100,
+        filter: {
+          type: AppTypeEnum.THIRDPARTY,
+        },
+        canFetchAppEvents: hasManagedAppsPermission && appAlertsEnabled,
+      },
+    });
+  const eventDeliveries = mapEdgesToItems(eventDeliveriesData?.apps) ?? [];
```
witoszekdev, line 50: suggestion: I think after we fetch event deliveries we could create `Map` in order to make querying app problems later in this hook faster, so in this case we would have `key: <app id`, `value: event deliveries`. This way when creating `const installedApps` we won't have to iterate over entire list

src/extensions/messages.ts
```diff
@@ -69,6 +69,29 @@ export const infoMessages = defineMessages({
   },
 });
 
+export const notifyMessages = defineMessages({
+  extensionReadyToUse: {
+    defaultMessage: "{name} is ready to be used",
+    id: "p42saN",
+    description: "extensions ready to be use",
```
witoszekdev, line 76: suggestion: typo

```suggestion
    description: "extensions list ready to be used",
```

src/extensions/messages.ts
```diff
@@ -69,6 +69,29 @@ export const infoMessages = defineMessages({
...
+  extensionRemoved: {
+    defaultMessage: "Extension successfully removed",
+    description: "extension has been removed",
+    id: "JbKGHC",
+  },
+  extensionInstallError: {
+    id: "VokWlV",
+    defaultMessage: "Couldn’t Install {name}",
+    description: "extension has not been installed",
+  },
```
witoszekdev, line 92: suggestion: we can add note that this is about extensions _list_

```suggestion
  extensionInstalled: {
    defaultMessage: "Extension installed",
    description: "extensions list has been installed",
    id: "SItkhR",
  },
  extensionRemoved: {
    defaultMessage: "Extension successfully removed",
    description: "extensions list has been removed",
    id: "JbKGHC",
  },
  extensionInstallError: {
    id: "VokWlV",
    defaultMessage: "Couldn’t Install {name}",
    description: "extensions list has not been installed",
  },
```

additional one:
we can rename `{name}` -> `{extensionName}` to make it clear it's about extension

src/extensions/messages.ts
```diff
@@ -158,4 +181,20 @@ export const messages = defineMessages({
...
+    id: "6hLZNA",
+    defaultMessage: "Are you sure you want to delete this app?",
+    description: "delete app",
+  },
+  deleteNamedApp: {
+    id: "EWD/wU",
+    defaultMessage:
+      "Deleting {name}, you will remove installation of the app. If you are paying for app subscription, remember to unsubscribe from the app in Saleor Marketplace. Are you sure you want to delete the app?",
+    description: "delete app",
+  },
```
witoszekdev, line 199: issue: Is this copy correct? We're using **App** instead of **Extension**

additional suggestion: in description we should probably use something like `delete extension page`

src/extensions/queries.ts
```diff
@@ -18,3 +24,62 @@ export const installedApps = gql`
...
+      }
+      edges {
+        node {
+          id
+          ...AppEventDeliveries
+        }
+      }
+    }
+  }
+`;
```
witoszekdev, line 77: question: don't we already have a query for fetching event deliveries? (e.g. when displaying "!" icon on navigation)

witoszekdev (commented):
Left some suggestion about performance + some minor ones and one important issue with copy (Apps delete, not Extension delete - to verify). Fyi, they are not blocking, we can also fix them in next PR.

I'm going to manually test this now

src/extensions/components/FailedWebhookInfo.tsx
poulch: When youu use flex, o event flex-inline text wrapping looks weird, I'd like to have natural text wrapping
Screenshot 2025-04-10 at 11 56 58

witoszekdev:
I found some visual issues while clicking through this:

## 1. Duplicated apps on installation

Steps:
1. Install Klavyio app that `/register` endpoint doesn't work

What happens: After a moment there appears 3 apps: 1) with Klaviyo app icon, 2) without app icon, 3) pending app installation
Expected: there's only one Klaviyo app

Screenshot:

CleanShot 2025-04-10 at 11 53 14@2x

I was able to do this only once, when I tried to replicate this I didn't work ¯\(°_o)/¯ - maybe there's some issue with what Saleor returns?

## 2. Jitter on app installation

Steps:
1. Go to explore page
2. Install Adyen app
4. Wait for a moment

Expected: pending app instalation cleanly moves to installed state
What happens: pending app dissapears for a moment, then newly installed app appears causing layout shift

Video:

https://github.com/user-attachments/assets/b7f8eb5d-cbd1-4b67-a7eb-2474e90f682c

## 3. Loading spinner

Steps:
1. Delete installed app

Expected: there's no loading state
What happens: there's a loading spinner at the top of the page, even though we don't have any pending query / mutation in the background

Screenshot:

CleanShot 2025-04-10 at 11 56 54@2x

witoszekdev:
Code wise, looks good to me ;)

I've found some other small UI issues:

## 1. Missalignment with "x" icon when installation failed

It visually looks wrong to me, like the icon is not aligned with the text:

CleanShot 2025-04-11 at 11 59 52@2x

## 2. Search doesn't work on app installations

I'm not sure if this is by design, but search looks like is not working with app installations, which might be confusing - we can perhaps include installations in the search or add some copy to explain that you cannot search for app installations, only installed apps.

My list:
CleanShot 2025-04-11 at 12 00 54@2x

Searching for app installation:
CleanShot 2025-04-11 at 12 00 49@2x

witoszekdev (approved):

witoszekdev (approved):

PR #5523, opened 2025-04-11
Add product variants selection in Vouchers 

Cloud11PL:
## Scope of the change

This PR adds an ability to assign individual variants to vouchers, both in edit and create forms. This allows users to select specific vouchers instead of entire products that the voucher applies to.

CleanShot 2025-04-14 at 08 54 53


andrzejewsky (changes requested):
Assigning new items sometimes wipes-up previous choices, seems like race-condition

andrzejewsky (approved):

PR #5527, opened 2025-04-14
Add customer metadata filtering in Customer list

Cloud11PL:
## Scope of the change

This PR adds filtering by metadata to customer list. It uses the same pattern as in Orders. 

CleanShot 2025-04-14 at 10 36 19


andrzejewsky (changes requested):
Having two inputs with values (key & value), refreshing page does not reflecting the state in the url

andrzejewsky (approved):

PR #5529, opened 2025-04-15
Installed extensions remove dev ff

poulch:
## Scope of the change
Remove dev feature flag from installed extensions list, which means that now we will show new installed extensions list instead of old apps installd list

### DIFF
--- a/.featureFlags/extensions.md
+++ b/.featureFlags/extensions.md
@@ -1,6 +1,6 @@
 ---
 name: extensions
-displayName: New Explore view
+displayName: New Extensions
 enabled: true
 payload: "default"
 visible: true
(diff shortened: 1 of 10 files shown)


andrzejewsky (approved):

.featureFlags/extensions.md
```diff
@@ -1,12 +1,13 @@
...
 enabled: true
 payload: "default"
 visible: true
 ---
 
 ![Extensions](./images/extensions.png)
 
-Experience the new interface for browsing and installing extensions in the ‘Explore’ view
+Experience the new "Explore" interface for browsing and installing extensions,
+and new ‘Installed’ view for managing installed extensions
```
witoszekdev, line 12: suggestion: let's use the same quotation marks

```suggestion
and new "Installed" view for managing installed extensions
```

witoszekdev (approved):

witoszekdev (approved):

andrzejewsky (approved):

PR #5540, opened 2025-04-17
Unify pagination across all views and lists

Cloud11PL:
## Scope of the change

This PR removes old material-ui based pagination in favor for updated one Dashboard uses in collection products list. This change will further unify Dashboard UI and UX.

Note: The Select component has to be updated after Macaw's Select position issue is fixed.


andrzejewsky (changes requested):
These paginations are unaligned; some views presenting this properly some does not as if they drop margins/paddings.

Additionally, having smaller screen then the table almost touches the edge of the screen, dropdown for page exceeds the screen (you cannot click it), simple fix is adding additional margin on the botom or dropdown should appear from the top (if easy, and possible)

Cloud11PL:
Funny story - MacOS has OS-level settings for showing scrollbars, they look like padding in Dashboard so depending on MacOS Appearance settings pagination may seem misaligned :) Fixing it now

PR #5541, opened 2025-04-17
Show custom apps and plugins on installed extensions

poulch:
## Scope of the change
This PR introduce plugins and a custom app for the installed extensions page. This feature is under the dev feature flag, you can turn it on by running this code:

```bash
localStorage.setItem('FF_extensions_dev', '{"enabled":true,"payload":"default"}')
```

https://github.com/user-attachments/assets/1634efd4-ae0a-477b-96c9-5b2e4a5d0d35


andrzejewsky (approved):

src/extensions/views/InstalledExtensions/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (a.name.toLowerCase() > b.name.toLowerCase()) {
+    return 1;
+  }
+
+  if (a.name.toLowerCase() < b.name.toLowerCase()) {
+    return -1;
+  }
+
+  return 0;
+};
```
witoszekdev, line 21: suggestion: we could replace it with `String.localeCompare()` with options to ignore case: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#ignorepunctuation

witoszekdev (approved):

PR #5543, opened 2025-04-18
Add caching policies for extensions list

andrzejewsky:
Now extensions list is loaded only once; when navigating across the pages, going back to the list shows it instantly while in the background everything is being fetched.

Changes
- disabling normalisation for a single app, now each app is cached individual, this prevents affecting by the other queries that are using app fetching from wiping up each other
- when fetching event delivers, there is no need to fetch all of them, we only are interested in active apps

### DIFF
--- a/src/graphql/client.ts
+++ b/src/graphql/client.ts
@@ -77,6 +77,9 @@ export const apolloClient = new ApolloClient({
           },
         },
       },
+      App: {
+        keyFields: false,
+      },
     } as TypedTypePolicies,
   }),
   link,
--- a/.changeset/light-fireants-wave.md
+++ b/.changeset/light-fireants-wave.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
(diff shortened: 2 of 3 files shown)


witoszekdev (approved):

src/graphql/client.ts
```diff
@@ -77,6 +77,9 @@ export const apolloClient = new ApolloClient({
           },
         },
       },
+      App: {
+        keyFields: false,
+      },
```
witoszekdev, line 82: ok, so this way we will be disabling normalization of apps (by default they would be stored as `App:<id>` in Apollo Cache.

I'm wondering if this change is actually hiding other underlying issue in dashboard, where some mutation changes app and breaks the cache instead? I don't quite understand how this fixes a loading issue when fetching apps list 🤔

andrzejewsky: it's jus change the way of normalisation, tl;dr now query results are stored separately, without impacting each other, sometimes it make sense.

Generally speaking, updating apollo would help in the future since as i recall there are more settings related to the cache and you can change something from the query level, without touching global config. eg when in certain case you want to update something

PR #5565, opened 2025-04-29
Copy views for managing plugins to `/extensions`

witoszekdev:
## Scope of the change

Added view for managing plugin to `/extensions` route in `/extensions/plugin/<id>` which is now used in extension list page for "View details" button

> [!NOTE]
> We have codecov and macawui migration not passing because I copy-pasted existing views which do not meet the requirements, we can improve them later on, for now I just want them moved to extensions with updated URLs

### DIFF
--- a/src/extensions/views/EditPluginExtension/EditPluginExtension.test.ts
+++ b/src/extensions/views/EditPluginExtension/EditPluginExtension.test.ts
@@ -5,7 +5,7 @@ import {
   ConfigurationTypeFieldEnum,
 } from "@dashboard/graphql";
 
-import { getConfigurationInput } from "./PluginsDetails";
+import { getConfigurationInput } from "./EditPluginExtension";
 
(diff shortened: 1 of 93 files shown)


src/extensions/views/EditPluginExtension/EditPluginExtension.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+test("Ensure that no secret is sent in input", () => {
+  const output: ConfigurationItemInput[] = getConfigurationInput(config, input);
+
+  expect(output).toHaveLength(2);
+  expect(
+    output.find(
+      field =>
+        config.find(configField => configField.name === field.name).type ===
+        ConfigurationTypeFieldEnum.PASSWORD,
```
andrzejewsky, line 68: the test cases should not contain any specific logic, i think you should make just toEqual call

andrzejewsky (dismissed):

andrzejewsky (approved):

PR #5566, opened 2025-04-29
Move app views to `/extensions`

witoszekdev:
## Scope of the change

This PR copies views from `/app` to `/extensions` with updated URLs e.g. `/extensions/app/<app_id>`

This also includes changes to:
- app routing -> when app will post an action to redirect to `/apps/<id>/app/subroute` we will actually redirect to `/extensions/app/<id>/subroute` to keep backwards compatibility
- updated other places in dashboard that point to apps, now will redirect to extensions when feature flag is enabled


andrzejewsky (approved):
I'm giving approve here, however one takeaway here: smaller pull requests.

it's fine than when we need to do massive change, but important is the fact the PR includes ONLY this massive change as it's just easier for the reviewer. In your case it could be:

one PR for changing folder naming (this would be big one)
one PR for changes routing
one PR for changing translations
one PR for any other "apps" references, if remaining 

Now, eg. when reviewer seeing the description PR that mentiones "it's just folder renaming" speeds up significantly the reviewing process, probably that would be accepting right away.

PR #5567, opened 2025-04-30
Update onboarding for extensions

witoszekdev:
## Scope of the change

This PR adds new onboarding step for browsing extensions page. It also adds autocomplete when user views installed extensions page. Old "View webhooks" onboarding step is also shown if user has "Extensions" feature flag disabled. When user enables this feature flag this new step will be shown as uncompleted.

This PR also adds missing tests for `InstalledExtensions` page.

CleanShot 2025-04-30 at 14 30 59@2x.


src/welcomePage/WelcomePageOnboarding/onboardingContext/utils.ts
```diff
@@ -54,29 +53,41 @@ export const getFirstExpanderStepId = (onboardingState: OnboardingState) => {
...
+
+  const stepIndex = stepsWithState.findIndex(step => step.id === lastCompletedStepId);
+
+  if (stepIndex === -1 || stepIndex === stepsWithState.length - 1) {
+    return "";
+  }
 
   return (
-    steps.slice(stepIndex + 1).find(step => !step.completed && step.expanded !== false)?.id ?? ""
+    stepsWithState.slice(stepIndex + 1).find(step => !step.completed && step.expanded !== false)
```
andrzejewsky, line 89: Tip: if you see something like that or do it yourself, it's a way better and more readable to delegate callback logic to the separate function and name it properly, example:

```
const arr = [1, 2, null, 3, 4, 5];

const toMultiplied = (a: number, b: number) => {
  return a * b;
};

const byNotNull = (a: number) => {
  return a !== null;
};

const byIsOdd = (a: number) => {
  return a % 2 !== 0;
};


const result = arr
  .filter(byNotNull)
  .filter(byIsOdd)
  .map(toMultiplied);
```

Now i can basically read this almost as sentence in english + worth keep this in a separate file for keeping separation

andrzejewsky (dismissed):

andrzejewsky (approved):

PR #5578, opened 2025-05-08
Change "Orders" to "Fulfulment"

andrzejewsky:
## Scope of the change

- Change copy "Orders" to "Fulfulment"
- Adjust references to "apps" in app view
- Duplicate alert on "installed" position

### DIFF
--- a/src/apps/components/AppAlerts/AlertExclamationIcon.tsx
+++ b/src/apps/components/AppAlerts/AlertExclamationIcon.tsx
@@ -4,18 +4,28 @@ import { ExclamationIconFilled } from "@dashboard/icons/ExclamationIconFilled";
 import { Box } from "@saleor/macaw-ui-next";
 import React, { useState } from "react";
 
-export const AlertExclamationIcon = () => {
+export const AlertExclamationIcon = ({
+  width = 17,
+  height = 17,
+}: {
+  width?: number;
+  height?: number;
+}) => {
   const [isHovered, setIsHovered] = useState(false);
 
   return (
     <Box
       __color={isHovered ? WARNING_ICON_COLOR_LIGHTER : WARNING_ICON_COLOR}
-      __width={17}
-      __height={17}
+      __width={width}
+      __height={height}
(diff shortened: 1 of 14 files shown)


src/apps/components/AppAlerts/AlertExclamationIcon.tsx
```diff
@@ -4,18 +4,28 @@ import { ExclamationIconFilled } from "@dashboard/icons/ExclamationIconFilled";
...
+      __height={height}
       onMouseEnter={() => setIsHovered(true)}
       onMouseLeave={() => setIsHovered(false)}
     >
-      {isHovered ? <ExclamationIconFilled /> : <ExclamationIcon width="17px" height="17px" />}
+      {isHovered ? (
+        <ExclamationIconFilled />
+      ) : (
+        <ExclamationIcon width={`${width}px`} height={`${height}px`} />
+      )}
```
Cloud11PL, line 28: Question: Does it "jump" on hover? Maybe we could pass width/height to `ExclamationIconFilled` too

witoszekdev (dismissed):

Cloud11PL (dismissed):

src/apps/components/AppAlerts/AlertExclamationIcon.tsx
andrzejewsky: turns out it's no, they are different icons

witoszekdev:
@andrzejewsky i see one issue with installed extension / app

1. Click extensions
2. Click "View app details"
3. Sidebar no longer has focus on extensions

CleanShot 2025-05-08 at 13 10 30@2x

After refresh the section is also collapsed:

CleanShot 2025-05-08 at 13 11 10@2x


I think this might be because we haven't migrated apps to the new URL, just letting you know

witoszekdev:
@andrzejewsky I also have an issue when app has broken webhook the state is not reflected in sidebar:

CleanShot 2025-05-08 at 13 20 56@2x

I'm not sure why this is happening, I though it might be cache, but I've cleared all site data and there is still no icon :/

PR #5579, opened 2025-05-08
Unify components in transaction refund view

Cloud11PL:
## Scope of the change

Before:
CleanShot 2025-05-08 at 14 03 58
CleanShot 2025-05-08 at 15 21 37
CleanShot 2025-05-08 at 15 24 25

After:
CleanShot 2025-05-08 at 14 04 08
CleanShot 2025-05-13 at 14 41 52
CleanShot 2025-05-13 at 14 40 45
CleanShot 2025-05-08 at 15 34 59


witoszekdev:
suggestion: I know that it's what we had previously but maybe we should add a tiny bit of margin here:

CleanShot 2025-05-13 at 12 59 44@2x

witoszekdev:
We don't have any loading state when form is being submitted, we should probably disable all form elements including this submit button:

https://github.com/user-attachments/assets/83755654-7ffe-439a-8b6f-2440136165fa

@andrzejewsky - maybe we could add spinner on this button when submitting? I'm not sure what's our current approach for this

src/components/QuantityInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+      isError && (
+        <Text
+          color="critical1"
+          __marginTop="-2px"
+          display="block"
+          position="absolute"
+          bottom={0}
+          fontSize={1}
+        >
+          {labels?.error ?? (
```
witoszekdev, line 64: question: shouldn't we show label.error value that has error description instead of a generic error message?

src/orders/components/OrderRefund/OrderRefund.tsx
```diff
@@ -62,8 +54,9 @@ const OrderRefund: React.FC<OrderRefundProps> = props => {
           disabled={disabled}
           name={"type" as keyof FormData}
           value={data.type}
-          onChange={onChange}
-          variant="inline"
+          onChange={change => {
+            onChange(change);
+          }}
```
witoszekdev, line 59: question: why did we need to make this change? is it due to a TypeScript error?

src/orders/components/OrderRefundFulfilledProducts/OrderRefundFulfilledProducts.tsx
```diff
@@ -165,35 +167,18 @@ const OrderRefundFulfilledProducts: React.FC<OrderRefundFulfilledProductsProps>
...
-                          "data-test-id": "quantityInput" + line?.id,
-                          max: (line?.quantity ?? 0).toString(),
-                          min: 0,
-                          style: { textAlign: "right" },
-                        }}
-                        fullWidth
-                        value={selectedLineQuantity?.value}
+                        className={classes.quantityInnerInputNoRemaining}
+                        data-test-id={"quantityInput" + line?.id}
+                        value={Number(selectedLineQuantity?.value)}
```
witoszekdev, line 174: issue: This could result in a NaNNaNNaNBatman 😄 if the value is null or undefined, we should probably add a fallback:

```suggestion
                        value={Number(selectedLineQuantity?.value || 0)}
```

src/orders/components/OrderReturnPage/OrderReturnRefundItemsCard/MaximalButton.tsx
```diff
@@ -1,34 +1,21 @@
...
-    <Button
-      className={classes.button}
-      onClick={onClick}
-      data-test-id="set-maximal-quantity-unfulfilled-button"
-    >
-      <FormattedMessage id="2W4EBM" defaultMessage="Set maximal quantities" description="button" />
-    </Button>
-  );
-};
+const MaximalButton: React.FC<MaximalButtonProps> = ({ onClick }) => (
```
witoszekdev, line 9: suggestion: if we already remove old code, we can also get rid of React.FC, like so:

```suggestion
const MaximalButton = ({ onClick }: MaximalButtonProps) => (
```

witoszekdev (changes requested):
one issue about possible NaN value, also please see my other commends above

src/components/QuantityInput.tsx
Cloud11PL: Not sure I understand, I used `??` so if there's an error label then it will show, otherwise it's going to be a generic error message

witoszekdev: oh, ok sorry you're right 👍🏻

witoszekdev (approved):

witoszekdev (approved):

PR #5584, opened 2025-05-09
Attributes properties form UI update

Cloud11PL:
## Scope of the change

Before:
CleanShot 2025-05-09 at 11 58 06

After: 
CleanShot 2025-05-13 at 15 56 06


andrzejewsky (approved):

witoszekdev:
Create attribute still uses outdated components:

CleanShot 2025-05-21 at 15 23 10@2x

witoszekdev (approved):

PR #5595, opened 2025-05-12
Fix: custom app details page not loading when user doesn't have manage_apps permissions

witoszekdev:
## Description

Fixed custom app views when user doesn't have `MANAGE_APPS` permissions:
- details view now loads
  - tokens and webhook & events are hidden with message
  - activate / deactivate button is hidden
  - save button is disabled, with tooltip
  - permission checkboxes are disabled
- view and edit app webhooks are hidden from users that don't have permission and return 404

## Screenshots
CleanShot 2025-05-13 at 11 33 45@2x
CleanShot 2025-05-13 at 11 33 52@2x

### DIFF
--- a/src/extensions/views/EditCustomExtension/EditCustomApp.tsx
+++ b/src/extensions/views/EditCustomExtension/EditCustomApp.tsx
@@ -19,6 +19,7 @@ import {
   useWebhookDeleteMutation,
   WebhookDeleteMutation,
 } from "@dashboard/graphql";
+import { useHasManagedAppsPermission } from "@dashboard/hooks/useHasManagedAppsPermission";
 import useNavigator from "@dashboard/hooks/useNavigator";
 import useNotifier from "@dashboard/hooks/useNotifier";
 import useShop from "@dashboard/hooks/useShop";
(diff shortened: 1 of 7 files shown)


src/extensions/views/EditCustomExtension/EditCustomApp.tsx
```diff
@@ -205,6 +208,11 @@ export const EditCustomExtension: React.FC<OrderListProps> = ({
     return <NotFoundPage backHref={ExtensionsUrls.resolveInstalledExtensionsUrl()} />;
   }
 
+  // Show loading state if data isn't available yet
+  if (loading && !data) {
+    return null; // Or some loading indicator
```
andrzejewsky, line 213: can we have one? I assume it's previous implementation, but worth fixing in next PR

andrzejewsky (approved):

src/extensions/views/EditCustomExtension/EditCustomApp.tsx
witoszekdev: i forgot to remove this when debugging 🙃

andrzejewsky (approved):

PR #5602, opened 2025-05-16
Change all references to page,content,page types to modeling-related

andrzejewsky:
## Scope of the change

Changing all references to old Content and Pages and replace them to the ones related to Modeling: Models, Structures, Model types.


Cloud11PL (dismissed):

witoszekdev:
one missing piece:
1. Go to models
2. Click "Create model"

There's a modal with "page" in title and input:
CleanShot 2025-05-16 at 10 58 01@2x

witoszekdev:
structures has "Model" in table heading:

CleanShot 2025-05-16 at 11 01 50@2x

src/pageTypes/components/PageTypeBulkDeleteDialog/PageTypeBulkDeleteDialog.tsx
```diff
@@ -29,27 +29,27 @@ const PageTypeBulkDeleteDialog: React.FC<PageTypeBulkDeleteDialogProps> = ({
...
       })}
       variant="delete"
     >
       {hasPages ? (
         <FormattedMessage
-          id="2Zyit2"
-          defaultMessage="{counter,plural,one{Page Type you want to delete is used by some pages. Deleting this page type will also delete those pages. Are you sure you want to delete this page type? After doing so you won’t be able to revert changes.} other{Page Types you want to delete are used by some pages. Deleting these page types will also delete those pages. Are you sure you want to delete {displayQuantity} page types? After doing so you won’t be able to revert changes.}}"
-          description="delete page types with its pages"
+          id="45W13E"
+          defaultMessage="{counter,plural,one{Model type you want to delete is used by some models. Deleting this model type will also delete those models. Are you sure you want to delete this model type? After doing so you won’t be able to revert changes.} other{Model types you want to delete are used by some models. Deleting these model types will also delete those models. Are you sure you want to delete {displayQuantity} model types? After doing so you won’t be able to revert changes.}}"
```
witoszekdev, line 41: question: do we want to use `’` as an apostrophe instead of `'`? I think we are using `'`' everywhere, so maybe we should stick with that one?

witoszekdev (dismissed):
lgtm, please look at my previous comments - but these can be done separately

krzysztofwolski (changes requested):
Go to "defaultMessages.json" and search for `"string".*page`. There are still multiple non updated strings. Beware - not all mentions should be updated, eg "home_page_", so it needs verification

andrzejewsky:
@krzysztofwolski this is not only one pattern - but yeah defaultMessages is the center here

src/pageTypes/components/PageTypeBulkDeleteDialog/PageTypeBulkDeleteDialog.tsx
andrzejewsky: hmm good question i think we can standardise it later on + add some spell checking

krzysztofwolski (approved):

PR #5612, opened 2025-05-23
Rows on extension lists are clickable

andrzejewsky:
## Scope of the change

- each app row is now clickable
- warnings have its own link to the manage view
- align buttons (always use icon buttons) + title attribute


src/extensions/types.ts
```diff
@@ -45,10 +45,16 @@ export type APIExtensionsResponse = Array<{
...
+  We should create dedicated types for each case (PendingInstallation, FailedInstallation and so on)
+*/
 export type InstalledExtension = {
   id: string;
   name: string;
   logo: ReactNode;
   info: ReactNode;
-  actions: ReactNode;
+  href?: string;
+  actions?: ReactNode;
```
witoszekdev, line 59: suggestion: maybe this should be named `action` not actions since it accepts single ReactNode, or maybe we can change it to be an array: `ReactNode[]`

witoszekdev (dismissed):
comment for future refactor

PR #5613, opened 2025-05-23
Fix search extensions input

witoszekdev:
This PR fixes search input globally to be wrapped with `<label>`, this fixes issue where user clicking outside of narrow `<input>` element (e.g. on search icon, near border) would not trigger focus.

It also unifies search look across extensions views, to remove border and fix margins to make it look more like other views:

**Explore extensions**

CleanShot 2025-05-26 at 11 44 25@2x

**Installed extensions**

CleanShot 2025-05-26 at 11 44 20@2x

### DIFF
--- a/.changeset/tiny-trees-hope.md
+++ b/.changeset/tiny-trees-hope.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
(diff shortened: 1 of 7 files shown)


andrzejewsky (changes requested):
please change also the margins around - they are too wide right now, use other pages as reference. eg product listing and its search

PR #5627, opened 2025-05-29
Replace extensions_dev -> extensions flag

witoszekdev:


witoszekdev:
Will be done in #5587

PR #5634, opened 2025-06-03
Add ButtonGroupWithDropdown component and use it when app is adding extension

krzysztofzuraw:
## Scope of the change

New button: `ButtonGroupWithDropdown` that is used to show app extensions

https://github.com/user-attachments/assets/5f45e6f3-a61c-4e74-845d-27d2d5beb5e7

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13231,28 +13231,6 @@
       "integrity": "sha512-jmYNElW7yvO7TV33CjSmvSiE2yco3bV2czu/OzDKdMNVZQWfxCblURLhf+47syQRBntjfLdd/H0egrzIG+oaFQ==",
       "dev": true
     },
-    "node_modules/fast-xml-parser": {
-      "version": "4.5.0",
-      "resolved": "https://registry.npmjs.org/fast-xml-parser/-/fast-xml-parser-4.5.0.tgz",
-      "integrity": "sha512-/PlTQCI96+fZMAOLMZK4CWG1ItCbfZ/0jx7UIJFChPNrx7tcEgerUgWbeieCM9MfHInUDyK8DWYZ+YrywDJuTg==",
-      "dev": true,
-      "funding": [
-        {
-          "type": "github",
-          "url": "https://github.com/sponsors/NaturalIntelligence"
-        },
-        {
-          "type": "paypal",
-          "url": "https://paypal.me/naturalintelligence"
-        }
-      ],
-      "dependencies": {
-        "strnum": "^1.0.5"
-      },
-      "bin": {
-        "fxparser": "src/cli/cli.js"
-      }
-    },
     "node_modules/fastq": {
       "version": "1.11.0",
       "dev": true,
@@ -20544,12 +20522,6 @@
       "funding": {
         "url": "https://github.com/sponsors/sindresorhus"
       }
-    },
-    "node_modules/strnum": {
-      "version": "1.0.5",
-      "resolved": "https://registry.npmjs.org/strnum/-/strnum-1.0.5.tgz",
-      "integrity": "sha512-J8bbNyKKXl5qYcR36TIO8W3mVGVHrmmxsd5PAItGkmyzwJvybiw2IVq5nqd0i4LSNSkB/sx9VHllbfFdr9k1JA==",
-      "dev": true
     },
     "node_modules/stylis": {
       "version": "4.1.3",
@@ -31185,15 +31157,6 @@
         }
       }
     },
-    "fast-xml-parser": {
-      "version": "4.5.0",
-      "resolved": "https://registry.npmjs.org/fast-xml-parser/-/fast-xml-parser-4.5.0.tgz",
-      "integrity": "sha512-/PlTQCI96+fZMAOLMZK4CWG1ItCbfZ/0jx7UIJFChPNrx7tcEgerUgWbeieCM9MfHInUDyK8DWYZ+YrywDJuTg==",
-      "dev": true,
-      "requires": {
-        "strnum": "^1.0.5"
-      }
-    },
     "fastq": {
       "version": "1.11.0",
       "dev": true,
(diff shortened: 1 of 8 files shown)


package-lock.json
kzuraw, line 1: TBH I don't know why this was changed - did we changed something recently?

src/components/ButtonGroupWithDropdown/ButtonGroupWithDropdown.tsx
```diff
@@ -0,0 +1,75 @@
...
+    label: string;
+    testId?: string;
+    onSelect: <T>(params?: T) => void;
+  }>;
+  testId?: string;
+  disabled?: boolean;
+}
+
+// TODO: consider moving this to Macaw UI
+export const ButtonGroupWithDropdown: React.FC<ButtonGroupWithDropdownProps> = ({
```
lkostrowski: I'm not sure about the name

I think it's either
- ButtonGroup, where you put Button & Button as children
- ButtonWithDropdown

WDYT?

src/components/ButtonGroupWithDropdown/ButtonGroupWithDropdown.tsx
```diff
@@ -0,0 +1,75 @@
+import { Box, Button, ChervonDownIcon, Dropdown, List, Text } from "@saleor/macaw-ui-next";
+import React from "react";
+
+interface ButtonGroupWithDropdownProps {
```
lkostrowski: shouldn't it inherit Dropdown props? So I can e.g. set margin on it?

src/components/ButtonGroupWithDropdown/ButtonGroupWithDropdown.tsx
```diff
@@ -0,0 +1,75 @@
...
+          }}
+        >
+          {children}
+        </Button>
+
+        <Dropdown.Trigger>
+          <Button
+            icon={<ChervonDownIcon />}
+            disabled={disabled}
+            borderColor="default1"
```
lkostrowski, line 50: Maybe it should be transparent, so it's like the background?

src/components/ButtonGroupWithDropdown/ButtonGroupWithDropdown.tsx
kzuraw: I'm wondering if by having `ButtonGroup` with two children is the best abstraction here - sure it is making this more usable in the future but currently there is only one usage which is `ButtonGroupWithDropdown` 🤔
 
About the name: it is called button group in other systems:
* https://tailwindcss.com/plus/ui-blocks/application-ui/elements/button-groups#component-bfc7e9cc9d7b5762cb139096ac3266c1
* https://getbootstrap.com/docs/4.0/components/button-group/#nesting

We also already have `ButtonWithDropdown` https://github.com/saleor/saleor-dashboard/blob/948b90b7fe52e495b15441b9a65f2cdc13bf25ab/src/components/ButtonWithDropdown/ButtonWithDropdown.tsx#L12

src/components/ButtonGroupWithDropdown/ButtonGroupWithDropdown.tsx
kzuraw: I don't need that yet - I made it narrow on purpose. When we will be adding new functionality we can then decide: either extend with Dropdown (or maybe with Button?) props

src/components/ButtonGroupWithDropdown/ButtonGroupWithDropdown.tsx
kzuraw: I've checked and it didn't work
2025-06-03 at 11 24 41 - CleanShot@2x
afaik `default1` is the background color we use in Dashboard anyway 😄

package-lock.json
witoszekdev: did you maybe use newer version of `npm`? 🤔

src/components/ButtonGroupWithDropdown/ButtonGroupWithDropdown.tsx
witoszekdev: +1 we already have `ButtonWithDropdown` in my opinion explicit name is fine since `ButtonGroup` will be probably different? (e.g. two buttons groped together)

src/components/ButtonGroupWithDropdown/ButtonGroupWithDropdown.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// TODO: consider moving this to Macaw UI
+export const ButtonGroupWithDropdown: React.FC<ButtonGroupWithDropdownProps> = ({
+  children,
+  options,
+  onClick,
+  disabled = false,
+  testId,
+}) => {
```
witoszekdev: We want to get rid of React.FC in Dashboard, since it has changes in props between React versions (e.g. children included vs not included):

```suggestion
// TODO: consider moving this to Macaw UI
export const ButtonGroupWithDropdown = ({
  children,
  options,
  onClick,
  disabled = false,
  testId,
}: ButtonGroupWithDropdownProps) => {
```

src/components/ButtonGroupWithDropdown/ButtonGroupWithDropdown.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <Box display="flex">
+        <Button
+          onClick={onClick}
+          data-test-id={testId}
+          disabled={disabled}
+          // TODO: fix this in Macaw UI - allow overriding border radius
+          style={{
+            borderBottomRightRadius: 0,
+            borderTopRightRadius: 0,
+          }}
```
witoszekdev: question: can't we use __borderTopRightRadius ?

kzuraw: Fixed in https://github.com/saleor/saleor-dashboard/pull/5634/commits/55307f035733f1e61a4b057ed99d74e99261781d

src/components/ButtonGroupWithDropdown/ButtonGroupWithDropdown.tsx
kzuraw: Fixed in https://github.com/saleor/saleor-dashboard/pull/5634/commits/55307f035733f1e61a4b057ed99d74e99261781d

src/components/ButtonGroupWithDropdown/ButtonGroupWithDropdown.tsx
kzuraw: Added `boxProps` in https://github.com/saleor/saleor-dashboard/pull/5634/commits/55307f035733f1e61a4b057ed99d74e99261781d

PR #5638, opened 2025-06-05
Addtional mounting points

krzysztofzuraw:
- **Add ButtonGroupWithDropdown component and use it when app is adding extension**
- **Category section**
- **Collection**
- **Gift cards**
- **Draft orders**
- **Discounts**
- **Vouchers**
- **Models**
- **PageTypes**
- **Structures**
- **Added additional mount points**
- **fix**


introspection.json
lkostrowski, line 1: is it expected?

schema.graphql
lkostrowski, line 1: also this, should we commit some specific schema?

introspection.json
kzuraw: We discussed it internally and in the future we will commit only schema.graphql instead of introspection.json - but it is out of the scope of this PR

PR #5646, opened 2025-06-09
Use GraphQL schema to generate types with fetching it every day

kzuraw:
## Scope of the change

PR opened by bot: https://github.com/saleor/saleor-dashboard/pull/5655


package.json
```diff
@@ -291,5 +291,10 @@
     "knip:fix": "knip --fix"
   },
   "description": "![Saleor Dashboard](https://user-images.githubusercontent.com/44495184/185379472-2a204c0b-9b7a-4a3e-93c0-2cb85205ed5e.png)",
-  "packageManager": "npm@10.8.2"
+  "packageManager": "npm@10.8.2",
+  "config": {
+    "saleor": {
+      "schemaVersion": "main"
```
andrzejewsky, line 297: hmm it cannot be main, even on main dashboard branch it should be always latest stable - think this way - when you merge something to main it can be released, right now

package.json
```diff
@@ -103,7 +103,6 @@
     "@graphql-codegen/cli": "^2.1.1",
     "@graphql-codegen/fragment-matcher": "^3.1.0",
     "@graphql-codegen/import-types-preset": "^2.1.12",
-    "@graphql-codegen/introspection": "^2.1.1",
```
lkostrowski, line 106: I'm not sure if we should throw away current behavior. It's good DX for someone who want's to develop with their own Saleor, don't thinking about source of schema

IMO we should add this behavior on top:
- If I want to use introspection, let me do it
- Give me a way to fetch schema and overwrite types from it

lkostrowski: After discussion proposal is:
- Generate ALWAYS from schema that is commited
- Additionally allow some script like "introspect-schema" that will use API_URL to print into schema.graphql (interop with current flow)
- always commit schema and types
- run CRON that will check with schema is up to date (on main should be core main, on branch 3.21 should be branch 3.21 etc)

package.json
lkostrowski: we need to change this behavior

package.json
kzuraw: Fixed - now we have `fetch-local-schema` script that will use `API_URL` & cron that checks if schema is up to date

lkostrowski (dismissed):

package.json
```diff
@@ -251,25 +250,26 @@
     "dev": "vite --host",
     "build": "cross-env NODE_OPTIONS=--max-old-space-size=8192 vite build",
     "preview": "vite preview",
-    "build-types": "node scripts/build-types.js",
+    "generate": "graphql-codegen --config ./graphql.config.ts",
```
witoszekdev: suggestion: by default GrpahQL Codegen uses `codegen.ts` file, maybe we can use it instead of overriding config?

https://the-guild.dev/graphql/codegen/docs/config-reference/codegen-config

src/graphql/types.generated.ts
```diff
@@ -2301,17 +2269,9 @@ export type GiftCardCreateInput = {
   balance: PriceInput;
   /** Slug of a channel from which the email should be sent. */
   channel?: InputMaybe<Scalars['String']>;
-  /**
-   * Code to use the gift card.
-   *
-   * DEPRECATED: this field will be removed. The code is now auto generated.
-   */
```
witoszekdev, line 2308: issue, non blocking: I think we might be missing some previous config, because some fields are no longer marked as deprecated, even if they are marked as such in schema:

CleanShot 2025-06-11 at 16 31 34@2x

I think ideally we should use JSDoc syntax to easily find fields which Dashboard should no longer use:

```
/** 
* ...
* @deprecated <insert note from Saleor>
*/
```

witoszekdev (dismissed):

witoszekdev:
@krzysztofzuraw we can also add changeset for such changes (I dunno if we should), they are listed in changelog as maintanance

lkostrowski (approved):

package.json
kzuraw: Done in https://github.com/saleor/saleor-dashboard/pull/5646/commits/3739755dae9acaee8ac366e736a0affb72325d36

PR #5649, opened 2025-06-10
Fix release report for slack

szczecha:
## Scope of the change
Test runs on releases are failing because the required Slack webhook URL is missing from the workflow parameters.

In [this PR](https://github.com/saleor/saleor-dashboard/pull/5560), the Slack webhook was made a required parameter for reporting test results. However, only one workflow currently passes this parameter (for tests on PRs), while others do not. As a result, any test runs triggered by those workflows are failing when they try to send results to Slack.

image

image

Needed to update `STAGING_TOKEN` as it looks like it's tided to staff user that created it so when I removed some Old employee from Cloud workflow stopped working. So I change it to take value from 1Password.


.github/workflows/run-test-manual.yml
```diff
@@ -125,6 +125,16 @@ jobs:
...
 
+      - name: Load secrets
+        uses: 1password/load-secrets-action@v2
+        id: load-secrets
+        with:
+          # Export loaded secrets as environment variables
+          export-env: true
+        env:
+          OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
+          SLACK_WEBHOOK_URL: "op://Continuous Integration/o2giutzdyhfdylvdw4kwg7lugu/password"
```
kzuraw: **Question:** shoudn't we use name of item in 1pass here? like we are doing in apps https://github.com/saleor/apps/blob/main/.github/workflows/e2e.yml#L84

szczecha: You're probably right, I just copy paste from mentioned PR :D

krzysztofzuraw (dismissed):

.github/workflows/run-tests-on-release.yml
```diff
@@ -270,10 +279,21 @@ jobs:
         if: needs.add-check-and-prepare-instance.outputs.FRAMEWORK == 'Cypress-tests'
         uses: ./.github/actions/combineReportsFromE2E
 
+      - name: Load secrets
+        uses: 1password/load-secrets-action@v2
```
kzuraw:
```suggestion
        uses: 1password/load-secrets-action@581a835fb51b8e7ec56b71cf2ffddd7e68bb25e0 # v2.0.0
```
Pin deps (security) as in https://github.com/saleor/apps/blob/main/.github/workflows/e2e.yml#L74

PR #5652, opened 2025-06-11
[3.20] Load variables from 1pass vault

kzuraw:
## Scope of the change

The same as https://github.com/saleor/saleor-dashboard/pull/5649 but for 3.20

### DIFF
--- a/.github/workflows/pr-automation.yml
+++ b/.github/workflows/pr-automation.yml
@@ -211,10 +211,10 @@ jobs:
           SHARD: ${{ matrix.shard }}
           BASE_URL: ${{ needs.initialize-cloud.outputs.BASE_URL }}
           API_URL: ${{ needs.initialize-cloud.outputs.API_URL }}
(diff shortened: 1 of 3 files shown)


NyanKiyoshi:
@krzysztofzuraw I don't see any Slack reporting step in 3.20[^1] (I can find it in `main` & 3.21 though) - are you sure there is anything to do in this PR?

[^1]: https://github.com/saleor/saleor-dashboard/blob/0af1d09db9b38e1fa67a39d8465b60059479c484/.github/actions/flaky-tests-report/action.yml

kzuraw:
@NyanKiyoshi this is weird as I see there is https://github.com/saleor/saleor-dashboard/blob/3.20/.github/actions/flaky-tests-report/action.yml on 3.20 branch 🤔

This PR was mainly added to get token from 1password instead of GHA secrets so we can more easily rotate them (this don't require admin for repo)

PR #5663, opened 2025-06-17
accept relative urls for NEW_TAB

lkostrowski:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,10 +11,9 @@ COPY locale/ locale/
 COPY scripts/ scripts/
 COPY vite.config.js ./
 COPY tsconfig.json ./
+COPY codegen.ts ./
 COPY *.d.ts ./
 COPY schema.graphql ./
-COPY introspection.json ./
-COPY introspection*.json ./
 COPY .featureFlags/ .featureFlags/
 
 COPY src/ src/


src/extensions/hooks/useExtensions.test.ts
```diff
@@ -1,11 +1,20 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
+import { ExtensionWithParams } from "@dashboard/extensions/types";
 import { AppExtensionMountEnum, PermissionEnum } from "@dashboard/graphql";
 import { renderHook } from "@testing-library/react-hooks";
 
 import { useExtensions } from "./useExtensions";
 
 const mockOpenApp = jest.fn();
 
+// Dodaj mockowanie newTabActions
```
witoszekdev: suggestion: we don't need that comment ;)

```suggestion
```

src/extensions/hooks/useExtensions.test.ts
```diff
@@ -14,6 +23,8 @@ jest.mock("../components/ExternalAppContext", () => ({
   }),
 }));
 
+const { newTabActions } = require("../new-tab-actions");
```
witoszekdev: suggestion: I think we can use ESModules imports normally here, instead of CommonJS

```suggestion
import { newTabActions } from "../new-tab-actions";
```

src/apps/isUrlAbsolute.ts
witoszekdev, line 1: issue: I'm not sure if this is going to be fully safe - I think we should also check for potential invalid URLs that enable XSS: for example `url = "data:text/html,<script>alert("XSS")</script>"` or `url = `blob:malicious-data` in this function and return `false` if they don't pass validation

For example we could do this:

```js
return ['http:', 'https:'].includes(parsedUrl.protocol);
```

witoszekdev (changes requested):

src/apps/components/AppWidgets/AppWidgets.tsx
```diff
@@ -74,7 +93,12 @@ const IframePost = ({
             ))}
         </>
       </form>
+      <Box ref={loadingRef} width={"100%"} __height={defaultIframeSize}>
+        <Skeleton __height={defaultIframeSize} />
+      </Box>
       <Box
+        style={hiddenStyle}
```
witoszekdev, line 100: tip: We can use this instead of inline styles

```suggestion
        __visibility="hidden"
```

krzysztofzuraw (approved):

lkostrowski:
changeset was a part of the previous changes, from the user perspective this is a single release with single changeset entry

PR #5664, opened 2025-06-17
ENG-294: Show attributes as separate entry on product search

witoszekdev:
## Scope of the change

This PR changes UI for selecting attribute search in Product list page. Now instead of typing attribute name in the first search element, users will see explicit "Attribute" option and a second input with attribute search will appear.

Existing search functionality is being kept the same.

https://github.com/user-attachments/assets/c342b141-e706-4f49-ab09-a0a775e41ce4


src/components/ConditionalFilter/API/providers/ProductFilterAPIProvider.tsx
```diff
@@ -104,14 +109,14 @@ export const useProductFilterAPIProvider = (): FilterAPIProvider => {
...
   };
-  const fetchLeftOptions = async (inputValue: string) => {
+  const fetchAttributeOptions = async (inputValue: string) => {
     const handler = new AttributesHandler(client, inputValue);
 
     return handler.fetch();
   };
 
   return {
     fetchRightOptions,
```
kzuraw, line 118: **Question:** what is `right` in this context? shouldn't we change that as well?

src/components/ConditionalFilter/API/providers/ProductFilterAPIProvider.tsx
```diff
@@ -51,7 +51,12 @@ const createAPIHandler = (
   }
 
   if (rowType === "attribute") {
-    return new AttributeChoicesHandler(client, selectedRow.value.value, inputValue);
+    if (selectedRow.selectedAttribute?.value) {
+      return new AttributeChoicesHandler(client, selectedRow.selectedAttribute.value, inputValue);
+    }
+
+    // Is this correct?
+    return new BooleanValuesHandler([]);
```
kzuraw: **Question:** shouldn't we throw here?

src/components/ConditionalFilter/FilterElement/FilterElement.ts
```diff
@@ -63,11 +63,25 @@ export class FilterElement {
...
     public constraint?: Constraint,
+    public selectedAttribute: ExpressionValue | null = null,
+    public availableAttributesList: LeftOperand[] = [],
+    public attributeLoading = false,
   ) {
-    const newConstraint = Constraint.fromSlug(this.value.value);
-
-    if (newConstraint) {
-      this.constraint = newConstraint;
+    this.updateConstraint();
```
kzuraw, line 70: **Suggestion:** what about `initConstraint`?

src/components/ConditionalFilter/API/providers/ProductFilterAPIProvider.tsx
witoszekdev: `right` is this input:

CleanShot 2025-06-18 at 13 56 04@2x

> shouldn't we change that as well?

nope, we don't want to change any logic for choosing attribute value / other product search options

I just got rid of `fetchLeftOptions` since it was a `return Promise.resolve([])` in every place I looked (previously it was used to fetch attributes in first input, now we have separate input for that)

src/components/ConditionalFilter/API/providers/ProductFilterAPIProvider.tsx
witoszekdev: whoops my bad I forgot to change it, yes you're right 😄

src/components/ConditionalFilter/FilterElement/FilterElement.ts
witoszekdev: I'm not sure I get this: I've updated recalculation of constraints into separate method, it's also calculated initially in constructor and recalculated when attribute is selected / unselected 🤔

kzuraw: makes sense - I was thinking if this `updateConstraint` is done only in constructor it can be named `initializeConstraint` but if it is used in many places `update` is making more sense 😄

krzysztofzuraw (approved):

andrzejewsky (approved):

PR #5681, opened 2025-06-26
Product materials

PitiMonster:


witoszekdev:
Hi @PitiMonster, this looks to me as a fork you made of Saleor Dashboard - are you sure you pushed to the correct remote? 😄 

We cannot accept this as a PR to Saleor Dashboard since this introduced a changes you made specifically for your deployment (e.g. `IS_ARTISO_ADMIN` env variable).

It also looks like you are hardcoding values for products, which is modeled in Saleor in [Product Attributes feature](https://docs.saleor.io/developer/attributes/overview). So in this case you would create "Material" attribute with all your values (wool, nylon, etc.).

If you have more questions about how to use Product Attributes feel free to ask around on [Saleor Community's Discord](https://saleor.io/discord). For contributions that add new features (not bug fixes) we have a guideline on approach this in our docs: https://docs.saleor.io/developer/community/contributing#feature-requests - we ask our community members to first start a thread on our Discord where we can discuss new features before we start any implementation work 🙌🏻

PR #5686, opened 2025-06-30
New search page

andrzejewsky:
## Scope of the change

- Added new page for global searching
- Search presents the results across major resources: product, variants, orders, models, model types, collections and categories
- You can trigger search by using `/` key

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -196,6 +196,7 @@ const Routes: React.FC = () => {
                     PermissionEnum.MANAGE_PAGES,
                     PermissionEnum.MANAGE_PRODUCT_TYPES_AND_ATTRIBUTES,
                   ]}
+                  matchPermission="any"
                   path="/search"
                   component={SearchSection}
                 />


src/index.tsx
```diff
@@ -188,6 +189,11 @@ const Routes: React.FC = () => {
               <Switch>
                 {legacyRedirects}
                 <SectionRoute exact path="/" component={WelcomePage} />
+                <SectionRoute
+                  // permissions={[PermissionEnum.MANAGE_PRODUCTS]}
```
witoszekdev: suggestion: I think we can get rid of this? I think search page should handle permissions for any entity on it's own (shown on the select along with search)

```suggestion
```

or maybe we can add a check if user has any of the listed permissions

```
anyPermissions={[MANAGE_PRODUCTS, ...]}
```

src/search/SearchShortcut.tsx
```diff
@@ -0,0 +1,43 @@
...
+      if (event.key === "/") {
+        event.preventDefault();
+
+        if (!location.pathname.includes("/search")) {
+          navigate("/search");
+        }
+      }
+    };
+
+    document.addEventListener("keydown", handleKeyDown);
```
witoszekdev, line 21: suggestion: I think this is fine, but maybe we could create global hook for handling any events on "document"? 🤔  or just separate hook to be used here, so we can keep track of such events

src/search/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+        window.history.back();
+      }
+    };
+
+    document.addEventListener("keydown", handleKeyDown);
+
+    return () => {
+      document.removeEventListener("keydown", handleKeyDown);
+    };
+  }, []);
```
witoszekdev, line 51: same thing as prev. comment we could add it as separate hook

I think we should also handle a case when user doesn't have history - maybe we could just navigate to home page?

src/search/resultsTable/ResultItem.tsx
```diff
@@ -0,0 +1,334 @@
...
+          </LinkCell>
+        </GridTable.Cell>
+        <GridTable.Cell __height="inherit" padding={0}>
+          <LinkCell href={orderUrl(node.id)}>
+            <DisplayDate date={node?.updatedAt} />
+          </LinkCell>
+        </GridTable.Cell>
+      </Row>
+    );
+  }
```
witoszekdev: suggestion: I think these could be each separate component in `ResultItem/OrderResultItem`

witoszekdev (dismissed):

src/search/SearchShortcut.tsx
andrzejewsky: it's always mounted (sidebar) so there is no need to move it somewhere globally

src/search/index.tsx
andrzejewsky: in case there is no history - browser will do nothing

witoszekdev (approved):

witoszekdev (approved):

PR #5702, opened 2025-07-03
ENG-294 Add support for filtering REFERENCE attributes on Product page

witoszekdev:
## Scope of the change

This PR adds support for filtering by REFERENCE attributes. It uses new `valueNames` filter for `query products`.

> [!CAUTION]
> This PR doesn't load correct diff in GitHub's beta Files changed browser

## Video

**Product variants**

https://github.com/user-attachments/assets/dcb777ff-050e-4f36-8f34-92ad3e2a2c7f

**Models** (prev. Pages)

https://github.com/user-attachments/assets/ed4400db-ba0c-4595-b60f-d95fc4e516e9

**Products**

https://github.com/user-attachments/assets/952d0640-4e42-4e2b-982d-86d121bf910e

### DIFF
--- a/src/components/Form/ExitFormDialog.tsx
+++ b/src/components/Form/ExitFormDialog.tsx
@@ -12,6 +12,7 @@ interface ExitFormDialogProps {
   isOpen: boolean;
 }
 
+/** @deprecated Use react-hook-form instead */
 const ExitFormDialog: React.FC<ExitFormDialogProps> = ({ onLeave, onClose, isOpen }) => {
   const intl = useIntl();
 
--- a/src/components/Form/Form.tsx
+++ b/src/components/Form/Form.tsx
@@ -18,6 +18,7 @@ export interface FormProps<TData, TErrors>
   mergeData?: boolean;
 }
 
+/** @deprecated Use react-hook-form instead */
 function Form<TData, Terrors>({
   children,
   initial,
--- a/src/components/Form/useExitFormDialog.ts
+++ b/src/components/Form/useExitFormDialog.ts
@@ -15,6 +15,7 @@ export interface UseExitFormDialogProps {
   isDisabled?: boolean;
 }
 
+/** @deprecated Use react-hook-form instead */
 export const useExitFormDialog = (
   { formId, isDisabled }: UseExitFormDialogProps = { formId: undefined },
 ): UseExitFormDialogResult => {
--- a/src/components/Form/useExitFormDialogProvider.tsx
+++ b/src/components/Form/useExitFormDialogProvider.tsx
@@ -16,6 +16,7 @@ const defaultValues = {
   formsData: {},
 };
 
+/** @deprecated Use react-hook-form instead */
 export function useExitFormDialogProvider() {
   const history = useHistory();
   const { history: routerHistory } = useRouter();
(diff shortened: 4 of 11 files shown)


src/components/ConditionalFilter/API/Handler.ts
```diff
@@ -81,6 +87,30 @@ export const createCustomerOptionsFromAPI = (
...
+        name: string;
+      };
+    };
+  }>,
+): ItemOption[] =>
+  data.map(
+    ({ node }) =>
+      ({
+        // AttributeValue.name for product variant reference attributes
+        label: node.product ? `${node.product.name}: ${node.name}` : node.name ?? "",
```
witoszekdev, line 108: This matches our search vector in Saleor which means that if user has product variant "M" on produdct "T-shirt", name will be "T-shirt: M".

Entering the same input into combobox will also return this option

src/components/ConditionalFilter/API/initialState/product/useProductInitialAPIState.tsx
```diff
@@ -87,21 +104,100 @@ export const useProductInitialAPIState = (): InitialProductAPIState => {
...
           },
         }),
       );
     }
 
     const data = await Promise.all(queriesToRun);
-    const initialState = createInitialProductStateFromData(data, channel);
+    let initialState = createInitialProductStateFromData(data, channel);
+
+    // Fetch Choices for Reference Attributes
```
witoszekdev, line 131: fyi: this needs to be done **after** we fetch data about attributes, because we need to know `entityType` from attribute's definition

src/components/ConditionalFilter/API/queries.ts
```diff
@@ -178,6 +182,19 @@ export const dynamicOperandsQueries = gql`
...
+        node {
+          id
+          name: title
+          slug
+          originalSlug: slug
+        }
+      }
+    }
+  }
+
```
witoszekdev, line 197: these are used to fetch options when user enteres data in combobox

src/components/ConditionalFilter/API/queries.ts
```diff
@@ -226,13 +259,44 @@ export const dynamicOperandsQueries = gql`
...
```
witoszekdev, line 275: these are used to rehydrate page data after user refreshes, in URL we store `slug`'s selected by user

witoszekdev:
We have e2e's failing due to changes in data in our env:

CleanShot 2025-07-10 at 16 27 09@2x
CleanShot 2025-07-10 at 16 27 12@2x

lkostrowski:
Dont know if it's only me, but I find copy/UX confusing when I see `attribute X in model`.

So suggestion: we can improve copy and if this is reference attribute type, full query for the user can look like:

```
Attribute X referenced in PageY
```

this "referenced" makes it more readable IMO

src/components/ConditionalFilter/constants.ts
```diff
@@ -608,6 +608,7 @@ export const STATIC_OPTIONS = [
 export const ATTRIBUTE_INPUT_TYPE_CONDITIONS = {
   DROPDOWN: [{ type: "multiselect", label: "in", value: "input-2" }],
   MULTISELECT: [{ type: "multiselect", label: "in", value: "input-2" }],
+  REFERENCE: [{ type: "multiselect", label: "in", value: "input-2" }],
```
lkostrowski, line 611: can we make `referenced in`?

witoszekdev: I'll ask for opinions, I think if we want to change it, this should be: **Attribute** .. **Product** .. **references** .. **<pick value>** (not `referenced in` because that it's the other way around: we have a reference on a product A to product B, but it's not referenced itself on product B)

PR #5704, opened 2025-07-04
[latest] Fix failing e2e tests

witoszekdev:
Fixed broken e2e tests:
- **Added missing extensions from environment to the list**
- **Fixed test using wrong path to navigate**
- **Fixed incorrect data-test-id for discounts button**
- **Fixed channel settings: transaction flow is now selected by default**
- **Fixed extensions tests**


playwright/tests/singlePermissions/plugins.spec.ts
```diff
@@ -8,7 +8,15 @@ test.use({ permissionName: "plugin" });
 let mainMenuPage: MainMenuPage;
 let extensionsPage: ExtensionsPage;
 
-const expectedExtensions = ["Admin emails", "OpenID Connect", "User emails", "Webhooks"];
+const expectedExtensions = [
+  "Admin emails",
+  "Adyen",
```
kzuraw: **Question:** we disabled `Adyen`, `Avalara` & `Stripe (deprecated)` plugins - does e2e env still have it enabled?

krzysztofzuraw (commented):
**Question:** should we port those changes to 3.21?

playwright/tests/singlePermissions/plugins.spec.ts
witoszekdev: Yes, e2e env still has these plugins enabled. They are no longer on the "Explore" page, but if you had them enabled previously they will be visible on "Installed" list.

witoszekdev:
@krzysztofzuraw sure thing, I've backported the changes: https://github.com/saleor/saleor-dashboard/pull/5709

lkostrowski (approved):

PR #5709, opened 2025-07-08
[3.21] Fix failing e2e tests

witoszekdev:
## Scope of the change

Port of changes for 3.21 branch from https://github.com/saleor/saleor-dashboard/pull/5704


lkostrowski:
merging, because looks like we can't release 3.21 due to failing tests

PR #5712, opened 2025-07-09
Fix generate schema version GitHub workflow

krzysztofzuraw:

### DIFF
--- a/.github/workflows/update-schema.yml
+++ b/.github/workflows/update-schema.yml
@@ -53,11 +53,20 @@ jobs:
         id: schema-version
         run: echo "version=$(node -p "require('./package.json').config.saleor.schemaVersion")" >> $GITHUB_OUTPUT
 
+      - name: Generate github access token
+        id: generate-token
+        if: steps.git-check.outputs.has_changes == 'true'
+        run: |
+          token=$(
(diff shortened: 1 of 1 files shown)


.github/workflows/update-schema.yml
```diff
@@ -53,11 +53,20 @@ jobs:
...
         run: echo "version=$(node -p "require('./package.json').config.saleor.schemaVersion")" >> $GITHUB_OUTPUT
 
+      - name: Generate github access token
+        id: generate-token
+        if: steps.git-check.outputs.has_changes == 'true'
+        run: |
+          token=$(
+            curl --request GET --url ${{ secrets.VAULT_URL }} --header "Authorization: JWT ${{ secrets.VAULT_JWT }}" | jq -r .token
+          )
+          echo "token=${token}" >> $GITHUB_OUTPUT
```
witoszekdev, line 63: question: Where is this getting tokens from? Is it a single token we have in a vault or is it generating new one on each run? 🤔 

If so then maybe this should be named "Fetch GitHub Access Token from Vault"?

kzuraw: We generate new one on each run - I will fix naming in the next PR (as this step is used in many places)

PR #5715, opened 2025-07-10
Mark useForm hook deprecated

lkostrowski:


witoszekdev:
@lkostrowski I've added more hooks related to that old hook solution: https://github.com/saleor/saleor-dashboard/pull/5716

PR #5719, opened 2025-07-11
ENG-175: Make reference attribute values clickable

witoszekdev:
### Feature changes

This PR makes attribute values clickable when pointing to a reference and rewrites existing components to use `@dnd-kit` library (already used in other places) and new macaw components which makes user interactions nicer.

It also changes UI to use horizontal space for attribute elements, previously they were always using only vertical space. It also removes "Assign reference" button which added additional vertical whitespace, instead replacing it with a Plus ("+") icon button:

https://github.com/user-attachments/assets/94216cc3-b93e-4560-af01-6b2f6cb26beb

### Maintenance changes

- `react-sortable-hoc` was marked as deprecated in ESLint and importing it will now show a warning.
- Product Variant details URL was changed from `/products/<productId>/variant/<variantId>` to `/products/variant/<variantId`, added a redirect to be removed in next minor release.


src/components/SortableChip/SortableChip.tsx
```diff
@@ -1,96 +1,97 @@
...
-            <CloseIcon
-              className={clsx(classes.closeIcon, {
-                [classes.disabled]: loading,
-              })}
+            <GripIcon
+              color="default2"
+              size="small"
+              data-test-id="button-drag-handle"
+              // @ts-expect-error - style is not accepted in types, but can be used
+              style={{ cursor: isDragged ? "grabbing" : "grab", outline: "none" }}
```
witoszekdev, line 77: This was fixed in macaw: https://github.com/saleor/macaw-ui/pull/907

.changeset/slimy-dragons-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Changed Product Variant details URL from `/products/<productId>/variant/<variantId>` to `/products/variant/<variantId`.
```
lkostrowski, line 5: will old URL still work and redirect?

src/channels/components/AssignmentList/Item.tsx
```diff
@@ -13,6 +14,7 @@ interface ItemProps extends SortableElementProps {
   onDelete: (id: string) => void;
 }
 
+/** @deprecatged This componnt should use @dnd-kit instead of react-sortable-hoc */
```
lkostrowski: typo in comment. I also wonder if this is correct, because you deprecate usage of the old library, not the `Item` right? So you can technically keep Item, but replace it's internal library?

Also *question*: How much work is to entirely drop this old library and is it planned? You introduced new one - but is there a plan to fully replace the old, so we don't stay with two solutions in the codebase?

src/channels/components/AssignmentList/SortableHandle.tsx
```diff
@@ -1,6 +1,7 @@
 import { DragIcon, makeStyles } from "@saleor/macaw-ui";
 import clsx from "clsx";
 import React from "react";
+// TODO: Remove and use @dnd-kit instead
```
lkostrowski: can you actually add a simple rule 

https://eslint.org/docs/latest/rules/no-restricted-imports

so no one will try to import this library if it's deprecated?

src/channels/components/AssignmentList/Item.tsx
kzuraw:
```suggestion
/** @deprecated This component should use @dnd-kit instead of react-sortable-hoc */
```

Double check if such JSDoc is correct as I always used multiline like:

```
/**
* @deprecated ...
*
**/
```
I see that the same typo is in other changed files

src/components/NavigatorSearch/modes/catalog.ts
```diff
@@ -73,7 +73,7 @@ export function searchInCatalog(
     label: getProductVariantLabel(variant),
     searchValue: `${variant.product.name} ${variant.name} ${variant.sku}`,
     onClick: () => {
-      navigate(productVariantEditUrl(variant.product.id, variant.id));
+      navigate(productVariantEditUrl(variant.id));
```
kzuraw, line 76: **Question:** we now will don't need product id for navigation? isn't url `products/<product-id>/variant/<variant-id>`?

krzysztofzuraw (commented):
Left a comments / questions. You can consider if you can remove some `@ts-strict` comments in changed files.

src/products/index.tsx
```diff
@@ -127,10 +127,7 @@ const Component = () => {
         <Route exact path={productListPath} component={ProductList} />
         <Route exact path={productAddPath} component={ProductCreate} />
         <Route exact path={productVariantAddPath(":id")} component={ProductVariantCreate} />
-        <Route
-          path={productVariantEditPath(":productId", ":variantId")}
-          component={ProductVariant}
-        />
```
lkostrowski: you should keep old path and add "redirect" to the new path so we don't face 404s

src/channels/components/AssignmentList/Item.tsx
lkostrowski: yea probably typo in `@deprecated` will not make TS and IDE to correctly understand it (which e.g. makes IDE to ~mark the usage properly~)

src/components/NavigatorSearch/modes/catalog.ts
lkostrowski: Variant ID must be unique in Saleor so technically it is good move and simplification of routing. However this should be a separate PR

witoszekdev: I've changed URL for Product Variants to `/dashboard/products/variant/<variantId>`. This had to be done along with this PR otherwise we wouldn't be able to easily navigate for attribute references that point to product variants. We would need to make nested query, find all product variants, their assosiated product.id. This would be needlessly expensive for Saleor.

Internal discussion about this change: https://saleorcommerce.slack.com/archives/C08UKL35R5G/p1752677861248299

src/channels/components/AssignmentList/Item.tsx
witoszekdev: I think we should mark these components as deprecated and simply create new versions if we change any view / component that uses it. This way you'll see it in editor that it uses the old library and we have motivation to get rid of it 😄 

About effort for replacing it... it wasn't trivial for this case (had to do some workarounds with sorting due to our chosen layout) but maybe other cases will be much easier to implement, hard to tell :/

I didn't introduce `@dnd-kit` in this PR, it was already used in Dashboard as replacement, only one dependency for constraints had to be added.

.eslintrc.json
```diff
@@ -32,10 +40,31 @@
...
+            "paths": [
+              {
+                "name": "react-sortable-hoc",
+                "message": "Use @dnd-kit instead of react-sortable-hoc."
+              }
+            ]
+          }
+        ]
+      }
     }
```
witoszekdev, line 68: Prettier had to format entire file 🙃 

Here's the only changed part

.changeset/slimy-dragons-whisper.md
witoszekdev: Nope, I've added a redirect and updated changeset 🫡

.changeset/slimy-dragons-whisper.md
```diff
@@ -0,0 +1,7 @@
...
+
+Old URL will still work and redirect to new one for backwards compatibility, it will be removed in next minor release.
```
witoszekdev, line 7: This should be removed in Saleor 3.23 release: https://linear.app/saleor/issue/f0fad883-81ae-4cd0-82dd-98b20effa577

witoszekdev:
@krzysztofzuraw @lkostrowski 

- I've added redirect for old product variants page
- Added ESLint rule to get rid of old react-sortable-hoc library

As for other comments:
- We had change URL with this PR, since it was blocking adding clickable references for us - we couldn't easily add fetching for product when referenced object was product variant (too expensive query); forgot to mention it in the description of PR 🙇🏻 
- I won't be removing `@ts-strict-ignore` from these components, since they are already deprecated and our end goal is to get rid of them entirely 😄

krzysztofzuraw (approved):

lkostrowski (approved):

PR #5722, opened 2025-07-14
Remove extensions dead code

witoszekdev:
This PR removes extensions dead code after we have already got rid of `extensions_dev` feature flag.

`false` state was no longer used since these views are only visible if you have `extensions` set to `true`


krzysztofzuraw (dismissed):

lkostrowski (approved):
not sure what this flag does but code looks fine if it builds

krzysztofzuraw (approved):

PR #5732, opened 2025-07-22
Command menu implementation

andrzejewsky:


src/components/NavigatorSearch/Actions.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+interface ActionsProps {
+  query: string;
+  onActionClick: (event: React.MouseEvent<HTMLAnchorElement>) => void;
+}
+
+export const Actions = ({ query, onActionClick }: ActionsProps) => {
+  const searchResults = fuzzySearch(allActions, query, ["name"]);
+
+  // @ts-expect-error - Old types for nodejs that does not have "groupBy"
```
lkostrowski: but it's not node.js running this code, but browser. Browser widely supports that. We must change typescript target

src/components/NavigatorSearch/useKeyboardNavigatioin.ts
lkostrowski, line 1: typo in file name

lkostrowski, line 1: should be unit tested

src/index.css
```diff
@@ -113,3 +113,11 @@ body {
...
   }
 }
+
+a.command-menu-item {
+  display: block;
+}
+
+.command-menu-item[data-focus="true"] {
+  background-color: var(--mu-colors-background-default1-hovered) !important;
+}
```
lkostrowski: if this is forced by the library, let's add comment that will tell that these styles come from this lib, etc

.changeset/nice-snakes-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Now you can trigger actions from CMD+K menu
```
witoszekdev, line 5: suggestion: Let's add examples of actions there are in CMD+K menu, e.g. Create new product

src/components/NavigatorSearch/Actions.tsx
```diff
@@ -0,0 +1,378 @@
...
+        </Text>
+      </Box>
+    </Link>
+  );
+};
+
+const allActions = [
+  {
+    section: "Orders",
+    name: "Create new order",
```
witoszekdev: issue: Shouldn't we use `react-intl` messages here as well? Without this our search won't work for translated list

src/components/NavigatorSearch/Actions.tsx
```diff
@@ -0,0 +1,378 @@
...
+      >
+        <Text size={2} fontWeight="medium" color="default1">
+          {children}
+        </Text>
+      </Box>
+    </Link>
+  );
+};
+
+const allActions = [
```
witoszekdev: suggestion: I think we could define type for action here (e.g. `const allActions: Action[]`) so that we don't have to define types for each component render method (long `onClick` type def) and you have completion when adding new items

src/components/NavigatorSearch/Actions.tsx
```diff
@@ -0,0 +1,378 @@
...
+  const groupedBySection = Object.groupBy(searchResults, result => result.section) as Record<
+    string,
+    {
+      Component: React.ComponentType<{
+        onClick?: (event: React.MouseEvent<HTMLAnchorElement>) => void;
+      }>;
+    }[]
+  >;
+
+  useEffect(onActionsLoaded, [query]);
```
witoszekdev, line 32: issue: Is there a reason why isn't this useEffect done in `CommandContent` component instead? Query is passed as prop and it triggers callback fn which is also a prop, overall everything is coming outside of this component 😄

src/components/NavigatorSearch/useActionItems.test.ts
```diff
@@ -0,0 +1,287 @@
+import { renderHook } from "@testing-library/react-hooks";
+
+import { useActionItems } from "./useActionItems";
+
+const mockNavigate = jest.fn();
+
+jest.mock("@dashboard/hooks/useNavigator", () => ({
+  __esModule: true,
```
witoszekdev, line 8: suggestion: I don't think we need this for mocks


```suggestion
```

src/components/NavigatorSearch/useInput.ts
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useRef } from "react";
+
+export const useInput = () => {
```
witoszekdev: suggestion: This name will be really hard to distinguish in import suggestions, imho we should use something more explicit here

```suggestion
export const useCommandMenuInput = () => {
```

src/components/NavigatorSearch/useKeyboardNavigation.ts
witoszekdev, line 1: question: This all looks fine especially around accessibility (aria-activedescendant 👏🏻), but it's a lot of code to maintain for us. I'm wondering why couldn't we use existing library to handle all of interaction within CMD+K, for example: https://github.com/pacocoursey/cmdk

src/components/NavigatorSearch/useKeyboardNavigation.test.ts
```diff
@@ -0,0 +1,250 @@
...
+
+    it("should not update aria active descendant when no active focused element", () => {
+      mockGetActiveFocusedElement.mockReturnValue(undefined);
+
+      renderHook(() => useKeyboardNavigation({ query: "test" }));
+
+      expect(mockFocusFirst).toHaveBeenCalled();
+      expect(mockGetActiveFocusedElement).toHaveBeenCalled();
+      expect(mockUpdateAriaActiveDescendant).not.toHaveBeenCalled();
+    });
```
witoszekdev: question: Shouldn't we set `aria-activedescendant` to empty string (or remove the property) when there is no active element? This should comply with WAI-ARIA spec: https://arc.net/l/quote/pqmvbkvk

> Sometimes states and properties are present in the DOM but have a zero-length string ("") as their value. Authors MAY specify a zero-length string ("") for any supported (but not required) state or property. User agents SHOULD treat state and property attributes with a value of "" the same as they treat an absent attribute.

issue: https://github.com/w3c/aria/issues/501

src/components/NavigatorSearch/useActionItems.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const focusFirst = () => {
+    currentFocusIndex.current = 0;
+    focusElement(currentFocusIndex.current, true);
+  };
+
+  const resetFocus = () => {
+    currentFocusIndex.current = undefined;
+    items.current = [];
+  };
```
witoszekdev, line 58: question: Will this actually reset focus to an element active before cmd+k menu was open?

src/components/NavigatorSearch/useKeyboardNavigation.test.ts
```diff
@@ -0,0 +1,250 @@
...
+      it("should prevent default and focus next element", () => {
+        const mockEvent = { preventDefault: jest.fn() } as any;
+
+        const result = hotkeyHandlers["tab"](mockEvent);
+
+        expect(mockEvent.preventDefault).toHaveBeenCalled();
+        expect(mockFocusNext).toHaveBeenCalled();
+        expect(result).toBe(false);
+      });
+    });
```
witoszekdev, line 184: question: What about shift+tab? Will it work the same way as up key?

src/components/NavigatorSearch/Actions.tsx
andrzejewsky: yes, because it's mounted later and only then it's accesible - i can catch all actions because they are just rendered (will leave a comment)

src/components/NavigatorSearch/useKeyboardNavigation.ts
andrzejewsky: fully agree, my bad, just didn't check properly if such library is available...

src/components/NavigatorSearch/useActionItems.ts
andrzejewsky: "focus" here is not the focus the "browser" focus. Browser focus can be one and it's on the input element always, while aria decendant "allows" to have more, in this case focus means the item that i have active in my state (kind of virtual focus)

PR #5733, opened 2025-07-22
More translation links and local storage caching

lkostrowski:


src/components/LanguageSwitch/LanguageSwitch.tsx
```diff
@@ -42,3 +42,26 @@ export const LanguageSwitch: React.FC<LanguageSwitchProps> = props => {
 };
 
 LanguageSwitch.displayName = "LanguageSwitch";
+
+export const LanguageSwitchWithCaching = (props: LanguageSwitchProps) => {
+  const { cachedValues, pushValue } = useCachedLocales();
+
+  const cachedWithLabel = cachedValues
+    .map(code => props.languages.find(l => l.code === code))
+    .filter(v => !!v) as LanguageFragment[];
```
witoszekdev, line 51: tip: We could use this to filter out missing values this should remove need for type casting.

```suggestion
    .filter(Boolean);
```

src/categories/components/CategoryUpdatePage/CategoryUpdatePage.tsx
```diff
@@ -96,8 +100,23 @@ export const CategoryUpdatePage: React.FC<CategoryUpdatePageProps> = ({
...
+                    TranslatableEntities.categories,
+                    categoryId,
+                  ),
+                )
+              }
+            />
             {extensionMenuItems.length > 0 && (
-              <TopNav.Menu items={[...extensionMenuItems]} dataTestId="menu" />
+              <Box marginLeft={3}>
+                <TopNav.Menu items={[...extensionMenuItems]} dataTestId="menu" />
```
witoszekdev, line 118: question: Why do we need to create new array on each render? 🤔 

```ts
[...extensionMenuItems]
```

witoszekdev (approved):

src/components/LanguageSwitch/LanguageSwitch.tsx
lkostrowski: actually it didn't thats why I casted. I think it's working in newer TS

src/categories/components/CategoryUpdatePage/CategoryUpdatePage.tsx
lkostrowski: I don't know, it's not part of my changes ;)

PR #5739, opened 2025-07-23
Update ESLint to v9 and TypeScript to latest

witoszekdev:
This PR migrates Dashboard to use ESLint v9 which uses flat config and fixes reported errors after that migration (that's the cause of large file count change). It also updates TypeScript to latest version - this had to be done to resolve issues with TypeScript ESLint plugin.

Most of changes in files are stylistic or to fix TypeScript errors and don't impact any logic. For logic changes there were comments from me (see below).

>[!NOTE]
> For some reason depcruise detected a change in:
> ```
> >   info Macaw UI - migration: testUtils/themeWrapper.tsx → node_modules/@saleor/macaw-ui/dist/cjs/index.js
> ```
> Which is incorrect, this file was not changed. That's why macaw migration check is failing

>[!NOTE]
> There are more warnings now due to recommended changes in Code Review. All of existing warnings still exists (checked using diff), due to plugin updates there were also new error messages added in the same files or some were changed.

It also fixes some other setup related issues that become apparent after this migration:
- TypeScript version is now pinned (previously we used `^`).
- Removed prettier ESLint plugin and config - we don't need it, since prettier is running separately and no rules are causing conflicts
- Removed unused dependencies
- Codegen files no longer add `/* eslint-disable */` comment, since they were removed from ESLint checks by using global config
- Fixed a bug in redirects that got discovered due to TypeScript error after update


src/components/RichTextEditor/ReactEditorJS.tsx
```diff
@@ -55,6 +55,10 @@ class ClientEditorCore implements EditorCore {
   public async render(data: OutputData) {
     await this._editorJS.render(data);
   }
+
+  public get dangerouslyLowLevelInstance() {
```
witoszekdev, line 59: Note: This has to be implemented to satisfy EditorCore interface

src/orders/components/OrderTransaction/components/CardTitle/CardTitle.tsx
```diff
@@ -140,18 +140,16 @@ export const OrderTransactionCardTitle: React.FC<CardTitleProps> = ({
 
         {showActionButtons && (
           <Box display="flex" flexDirection="row" gap={2}>
-            {actions
-              .filter(action => action !== TransactionActionEnum.REFUND)
```
witoszekdev, line 144: This is already filter above in the same file (actions don't have refund) - it was a duplicate of filter

src/search/useSearchCriteria.ts
```diff
@@ -34,7 +34,7 @@ export const useSearchCriteria = () => {
   }, []);
 
   useEffect(() => {
-    if (!triggerSearch) false;
```
witoszekdev, line 37: I'm not sure about this but just using a `false` didn't cause anything, so I guess this was the intent

src/utils/richText/useRichText.test.ts
```diff
@@ -57,6 +55,7 @@ describe("useRichText", () => {
       destroy: jest.fn(),
       clear: jest.fn(),
       render: jest.fn(),
+      dangerouslyLowLevelInstance: {},
```
witoszekdev, line 58: This is required by interfaces in EditorJS

eslint.config.mjs
```diff
@@ -0,0 +1,217 @@
...
+import reactHooks from "eslint-plugin-react-hooks";
+import react from "eslint-plugin-react";
+import importPlugin from "eslint-plugin-import";
+import formatjs from "eslint-plugin-formatjs";
+import reactRefresh from "eslint-plugin-react-refresh";
+import simpleImportSort from "eslint-plugin-simple-import-sort";
+import localRules from "./lint/rules/named-styles.js";
+
+export default defineConfig([
+  globalIgnores(["node_modules/", "build/", "dist/", "dev-dist/", "coverage/"]),
```
witoszekdev: Note: eslint no longer has native `.eslitignore` file it has be be defined here, or loaded from such file manually

eslint.config.mjs
```diff
@@ -0,0 +1,217 @@
...
+
+  // React configuration
+  {
+    files: ["src/**/*.{js,jsx,ts,tsx}"],
+    ...react.configs.flat.recommended,
+    languageOptions: {
+      ...react.configs.flat.recommended.languageOptions,
+      globals: {
+        ...globals.browser,
+        ...globals.es2021,
```
witoszekdev: question: what target we should have in dashboard? @krzysztofzuraw @lkostrowski

package.json
```diff
@@ -158,12 +153,11 @@
     "mochawesome-report-generator": "^6.2.0",
     "playwright-ctrf-json-reporter": "^0.0.18",
     "rollup-plugin-polyfill-node": "^0.11.0",
-    "typescript": "^5.0.4",
+    "storybook": "9.0.18",
+    "typescript": "5.8.3",
```
witoszekdev, line 201: Note: pinned version of TypeScript due to using `^` (probably a mistake) we were already using latest version 😄

package.json
```diff
@@ -241,9 +235,6 @@
       ]
     }
   },
-  "overrides": {
-    "resolve": "1.20.0"
-  },
```
witoszekdev: This is no longer needed, it broke installation with ESLint v9

package.json
```diff
@@ -269,7 +260,7 @@
     "test": "jest src/",
     "test:ci": "jest src/ --coverage",
     "test:watch": "jest --watch src/",
-    "lint": "eslint \"{src,playwright}/**/*.@(tsx|ts|jsx|js)\" --fix",
+    "lint": "cross-env NODE_OPTIONS=--max-old-space-size=8192 eslint \"{src,playwright}/**/*.@(tsx|ts|jsx|js)\" --fix",
```
witoszekdev: I had to add this because **sometimes** ESLint run out of memory on my machine. We already use more memory for other scripts in dashboard as well.

package.json
```diff
@@ -225,6 +218,9 @@
       "json",
       "node"
     ],
+    "transformIgnorePatterns": [
+      "/node_modules/(?!chroma-js/.*)"
```
witoszekdev: For some reason chroma-js didn't get parsed in DataGrid tests after updates

package.json
```diff
@@ -122,33 +120,29 @@
...
+    "@typescript-eslint/eslint-plugin": "8.38.0",
+    "@typescript-eslint/parser": "8.38.0",
     "@vitejs/plugin-react-swc": "^3.8.0",
     "code-inspector-plugin": "^0.6.1",
     "core-js": "^3.7.0",
     "cross-env": "^6.0.3",
     "dependency-cruiser": "^12.10.0",
     "dotenv": "^10.0.0",
-    "eslint": "^8.43.0",
-    "eslint-config-prettier": "^8.5.0",
```
witoszekdev, line 190: I've checked our eslint config according to their docs: https://github.com/prettier/eslint-config-prettier#cli-helper-tool - and we don't have any rules that conflict with prettier, so it's safe to get rid of this.

package.json
```diff
@@ -57,10 +57,8 @@
     "debug": "^4.3.4",
     "downshift": "^6.1.7",
     "editorjs-inline-tool": "^0.4.0",
-    "eslint-plugin-prettier": "^5.1.3",
```
witoszekdev, line 106: This plugin is not needed since we're running prettier separately anyway

package.json
```diff
@@ -57,10 +57,8 @@
...
     "faker": "^5.1.0",
     "fast-array-diff": "^0.2.0",
-    "front-matter": "^4.0.2",
```
witoszekdev, line 109: According to knip this wasn't used

eslint.config.mjs
```diff
@@ -0,0 +1,217 @@
...
+import typescriptParser from "@typescript-eslint/parser";
+import reactHooks from "eslint-plugin-react-hooks";
+import react from "eslint-plugin-react";
+import importPlugin from "eslint-plugin-import";
+import formatjs from "eslint-plugin-formatjs";
+import reactRefresh from "eslint-plugin-react-refresh";
+import simpleImportSort from "eslint-plugin-simple-import-sort";
+import localRules from "./lint/rules/named-styles.js";
+
+export default defineConfig([
```
lkostrowski: is typescript running here? Probably soon we will be able to do it with native node without jiti ( in v22 its behind the flag), but maybe jsdoc comment could work here? Like we have in jest config or next cofnig

eslint.config.mjs
```diff
@@ -0,0 +1,217 @@
...
+
+  // Base configuration for all files
+  {
+    files: ["{src,playwright}/**/*.{js,jsx,ts,tsx}"],
+    plugins: {
+      "simple-import-sort": simpleImportSort,
+    },
+    rules: {
+      "simple-import-sort/imports": "error",
+      "simple-import-sort/exports": "error",
```
lkostrowski: was it necessary to change config? It would be much easier to
1. update eslint and don't change the rules
2. then change the rules

now entire codebase can be conflicted due to pr size

eslint.config.mjs
```diff
@@ -0,0 +1,217 @@
...
+      "@typescript-eslint/prefer-nullish-coalescing": "off",
+      "@typescript-eslint/explicit-function-return-type": "off",
+      "@typescript-eslint/no-floating-promises": "off",
+      "@typescript-eslint/no-misused-promises": "off",
+      "@typescript-eslint/consistent-type-imports": "off",
+      "@typescript-eslint/no-confusing-void-expression": "off",
+      // Migration in progress:
+      // Tracked in https://github.com/saleor/saleor-dashboard/issues/3813
+      "@typescript-eslint/no-non-null-assertion": "off",
+      "@typescript-eslint/no-non-null-asserted-optional-chain": "off",
```
lkostrowski: let's change it to WARN. We don't want it to break the build but we DO WANT developer feedback that he does bad stuff

eslint.config.mjs
lkostrowski: I think we should safely go with 2024. Best if we somehow can track it between major versions, e.g. Saleor 3.22 drop some old browser. But, I think we don't have to overthink it. I would go with 2024 now and if we receive any error in Sentry, we can add polyfill

eslint.config.mjs
lkostrowski, line 1: I'm not checking every rule but I think you should port them 1:1. This PR should be about bumping ESLint, not changint it's rules. Would be good if you verify that

package.json
lkostrowski: We can discuss what's better. If we run eslint and prettier separately, probably it's slower because it must re-run files scanning

package.json
lkostrowski: ^ is not about last version but to allow `pnpm install` to bump it when 5.8.4 appears. Anyway IMO we should pin everything

src/extensions/hooks/useExtensions.test.ts
```diff
@@ -14,8 +14,14 @@ jest.mock("../new-tab-actions", () => ({
...
-jest.mock("@dashboard/auth/hooks/useUserPermissions");
-jest.mock("@dashboard/graphql");
+jest.mock("@dashboard/auth/hooks/useUserPermissions", () => ({
+  useUserPermissions: jest.fn(),
+}));
+
+jest.mock("@dashboard/graphql", () => ({
+  ...(jest.requireActual("@dashboard/graphql") as jest.Mocked<typeof import("@dashboard/graphql")>),
+  useExtensionListQuery: jest.fn(),
+}));
```
lkostrowski, line 24: i think spyOn is easier syntax here (but not in scope)

lkostrowski (approved):

eslint.config.mjs
kzuraw: Node 22.18 should have support for ts by default https://nodejs.org/en/blog/release/v22.18.0#type-stripping-is-enabled-by-default

package.json
witoszekdev: Yes, this comment was incorrect, I forgot to remove it 😅

package.json
witoszekdev: typescript-eslint maintainers recommend to actually get rid of the eslint plugin because it slows down ESLint: https://typescript-eslint.io/troubleshooting/typed-linting/performance/#eslint-plugin-prettier

https://www.joshuakgoldberg.com/blog/you-probably-dont-need-eslint-config-prettier-or-eslint-plugin-prettier/

I've seen some examples where it was a reason for 70% of linter runtime 🙈 https://duncanleung.com/profile-measure-find-remove-slow-eslint-rules/

eslint.config.mjs
witoszekdev: I've added `@ts-check`

eslint.config.mjs
witoszekdev: I've checked and it didn't cause actually any changes in this PR so I'll leave it as error

witoszekdev:
@lkostrowski I've checked rules again and there in fact were some of them missing either due to my mistake or because I used different set of recommended rules 😅

krzysztofzuraw (approved):

lkostrowski (approved):

krzysztofzuraw (approved):

PR #5742, opened 2025-07-24
Improve some ts-strict-ignore instances

lkostrowski:

### DIFF
--- a/src/hooks/useOnSetDefaultVariant.ts
+++ b/src/hooks/useOnSetDefaultVariant.ts
@@ -1,4 +1,3 @@
-// @ts-strict-ignore
 import { Node, useProductVariantSetDefaultMutation } from "@dashboard/graphql";
 import useNotifier from "@dashboard/hooks/useNotifier";
 import { getProductErrorMessage } from "@dashboard/utils/errors";
@@ -7,9 +6,10 @@ import { useIntl } from "react-intl";
 function useOnSetDefaultVariant(productId: string, variant: Node) {
   const notify = useNotifier();
   const intl = useIntl();
+
   const [productVariantSetDefault] = useProductVariantSetDefaultMutation({
     onCompleted: data => {
-      const errors = data.productVariantSetDefault.errors;
+      const errors = data.productVariantSetDefault?.errors ?? [];
 
       if (errors.length) {
         errors.map(error =>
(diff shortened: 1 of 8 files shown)


src/orders/views/OrderGrantRefund/OrderGrantRefund.tsx
```diff
@@ -82,7 +81,7 @@ const OrderGrantRefund: React.FC<OrderGrantRefundProps> = ({ orderId }) => {
       />
 
       <OrderGrantRefundPage
-        order={data?.order}
+        order={data?.order ?? undefined}
```
kzuraw, line 84: is this a different than `data?.order`?

src/orders/components/OrderGrantRefundPage/components/ShippingInluded.tsx
```diff
@@ -33,7 +33,7 @@ export const ShippingIncluded = ({
         data-test-id="refundShippingCheckbox"
         disabled={!currency || !canRefundShipping}
       >
-        {!currency ? (
+        {!currency || !amount ? (
```
kzuraw, line 36: what if amount will be 0?

src/hooks/useOnSetDefaultVariant.ts
```diff
@@ -38,11 +38,11 @@ function useOnSetDefaultVariant(productId: string, variant: Node) {
       }
     },
   });
-  const onSetDefaultVariant = (selectedVariant = null) => {
+  const onSetDefaultVariant = (selectedVariant: Node | null = null) => {
     productVariantSetDefault({
       variables: {
         productId,
-        variantId: variant ? variant.id : selectedVariant.id,
+        variantId: variant ? variant.id : (selectedVariant?.id as string),
```
kzuraw, line 45: I'm wondering if we should not throw here if there will be no `variant.id` or `selectedVariant?.id` instead - I'm wondering what will happen in runtime or it will be edge case? 🤔

src/orders/views/OrderGrantRefund/OrderGrantRefund.tsx
lkostrowski: I was thinking about this too, but clearly without TS prints error 🤔

src/orders/components/OrderGrantRefundPage/components/ShippingInluded.tsx
lkostrowski: good catch

src/hooks/useOnSetDefaultVariant.ts
lkostrowski: I'm trying to avoid runtime changes as much as possible, only fixing typescript. So in the worst scenario I preserved a bug, but I don't want to risk breaking things by changing behavior

src/orders/components/OrderGrantRefundPage/components/ShippingInluded.tsx
lkostrowski: I checked, amount (IMoney) is an object, so it should be fine

krzysztofzuraw (approved):

PR #5747, opened 2025-07-25
Improve ESLint import rules

lkostrowski:
- MUI and legacy macaw is now warned as deprecated
- Imports are now grouped (external, internal, etc)

### DIFF
--- a/src/attributes/components/AttributePage/AttributePage.tsx
+++ b/src/attributes/components/AttributePage/AttributePage.tsx
@@ -1,5 +1,11 @@
+import React from "react";
+import { useIntl } from "react-intl";
+import slugify from "slugify";
+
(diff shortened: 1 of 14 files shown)


.eslintrc.json
```diff
@@ -64,12 +75,49 @@
...
+        "alphabetize": {
+          "order": "asc",
+          "caseInsensitive": true
+        }
+      }
+    ],
+    "sort-imports": "off",
     "import/no-default-export": "warn",
-    "simple-import-sort/imports": "warn",
+    "simple-import-sort/imports": "off",
```
kzuraw, line 120: maybe we can remove `eslint-plugin-simple-import-sort` altogether if we disable the rule?

witoszekdev (approved):

PR #5755, opened 2025-07-31
Fix bulk gift card re-rendering

krzysztofzuraw:
## Scope of the change

This PR fixes gift card modal re-rendering issues. I also removed `@ts-ignore`, barrel file and default export.

Modal: 
2025-07-31 at 14 35 47 - CleanShot@2x

### DIFF
--- a/src/giftCards/GiftCardBulkCreateDialog/GiftCardBulkCreateDialog.tsx
+++ b/src/giftCards/GiftCardBulkCreateDialog/GiftCardBulkCreateDialog.tsx
@@ -1,11 +1,6 @@
-// @ts-strict-ignore
 import { IMessage } from "@dashboard/components/messages";
 import { DashboardModal } from "@dashboard/components/Modal";
-import {
-  GiftCardBulkCreateInput,
-  useChannelCurrenciesQuery,
-  useGiftCardBulkCreateMutation,
-} from "@dashboard/graphql";
+import { GiftCardBulkCreateInput, useGiftCardBulkCreateMutation } from "@dashboard/graphql";
 import useCurrentDate from "@dashboard/hooks/useCurrentDate";
 import useNotifier from "@dashboard/hooks/useNotifier";
 import { DialogProps } from "@dashboard/types";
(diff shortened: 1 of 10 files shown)


src/giftCards/GiftCardBulkCreateDialog/GiftCardBulkCreateDialog.tsx
```diff
@@ -111,24 +105,20 @@ const GiftCardBulkCreateDialog: React.FC<DialogProps> = ({ onClose, open }) => {
...
           <ContentWithProgress>
-            {!loadingChannelCurrencies && (
-              <GiftCardBulkCreateDialogForm
-                opts={bulkCreateGiftCardOpts}
-                onClose={onClose}
-                formErrors={formErrors}
-                onSubmit={handleSubmit}
-              />
-            )}
+            <GiftCardBulkCreateDialogForm
```
lkostrowski, line 108: shouldnt it incldue prop `loading={loadingChannelCurrencies}`?

src/giftCards/hooks/useChannelCurrenciesWithCache.ts
```diff
@@ -0,0 +1,46 @@
...
+  useChannelCurrenciesQuery,
+} from "@dashboard/graphql";
+import { useMemo } from "react";
+
+/**
+ * Custom hook that uses useChannelCurrenciesQuery only if the data isn't already in cache.
+ * This helps avoid unnecessary network requests when the channel currencies data is already available.
+ */
+export function useChannelCurrenciesWithCache() {
+  const client = useApolloClient();
```
lkostrowski, line 15: why this is not cached with apollo? I mean we can't create a wrapper for every object and appollo supports that

https://www.apollographql.com/docs/react/caching/overview

kzuraw: TBH I don't know but if I didn't used custom hook `useChannelQuery` gets triggered - maybe we should use `useQuery` instead of custom generated hooks?

src/giftCards/GiftCardBulkCreateDialog/GiftCardBulkCreateDialog.tsx
kzuraw: Nope - it turns out loading was only for `GiftCardCreateMoneyInput` - now this component has `loading` inside.

lkostrowski (approved):

PR #5758, opened 2025-08-01
Add base order v2 page struct

kzuraw:
## Scope of the change

Base new order page structure - storybook https://saleor-dashboard-storybook-git-eng-569-or-8d413f-saleorcommerce.vercel.app/?path=/story/order-detail-page--fulfilled

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -8361,11 +8361,28 @@ input ProductVariantWhereInput @doc(category: "Products") {
   """Filter by when was the most recent update."""
   updatedAt: DateTimeRangeInput
 
+  """
+  Filter by attributes associated with the variant.
+  
+  Added in Saleor 3.22.
+  """
+  attributes: [AssignedAttributeWhereInput!]
+
   """List of conditions that must be met."""
   AND: [ProductVariantWhereInput!]
 
   """A list of conditions of which at least one must be met."""
   OR: [ProductVariantWhereInput!]
+}
+
+input AssignedAttributeWhereInput {
+  """Filter by attribute slug."""
+  slug: String
+
+  """
+  Filter by value of the attribute. Only one value input field is allowed. If provided more than one, the error will be raised.
+  """
+  value: AssignedAttributeValueInput
 }
 
 input ProductVariantSortingInput @doc(category: "Products") {
@@ -13077,16 +13094,6 @@ input MetadataValueFilterInput {
 
   """The value included in."""
   oneOf: [String!]
-}
-
-input AssignedAttributeWhereInput {
-  """Filter by attribute slug."""
-  slug: String
-
-  """
-  Filter by value of the attribute. Only one value input field is allowed. If provided more than one, the error will be raised.
-  """
-  value: AssignedAttributeValueInput
 }
 
 type PageTypeCountableConnection @doc(category: "Pages") {
--- a/src/graphql/types.generated.ts
+++ b/src/graphql/types.generated.ts
@@ -6005,6 +6005,12 @@ export type ProductVariantWhereInput = {
   AND?: InputMaybe<Array<ProductVariantWhereInput>>;
   /** A list of conditions of which at least one must be met. */
   OR?: InputMaybe<Array<ProductVariantWhereInput>>;
+  /**
+   * Filter by attributes associated with the variant.
+   *
+   * Added in Saleor 3.22.
+   */
+  attributes?: InputMaybe<Array<AssignedAttributeWhereInput>>;
   ids?: InputMaybe<Array<Scalars['ID']>>;
   metadata?: InputMaybe<Array<MetadataFilter>>;
   /** Filter by product SKU. */


package.json
```diff
@@ -38,6 +38,7 @@
     "@material-ui/lab": "^4.0.0-alpha.61",
     "@material-ui/styles": "^4.11.4",
     "@radix-ui/react-portal": "^1.1.4",
+    "lucide-react": "^0.534.0",
```
witoszekdev, line 87: omg yes ❤️

src/orders-v2/status-pill.tsx
```diff
@@ -0,0 +1,39 @@
...
+}: {
+  status: PillStatusType;
+  children: React.ReactNode;
+}) => {
+  const getFillColorsBasedOnStatus = (status: PillStatusType) => {
+    switch (status) {
+      case "success":
+        return {
+          __backgroundColor: "oklch(0.92 0.16 145)",
+          __borderColor: "oklch(0.72 0.27 145)",
```
witoszekdev: discussion, not for this PR: if we're adding new palette of colors, maybe we could use a tool like https://www.radix-ui.com/colors/docs/overview/custom-palettes - and generate a full color pallete we could then use in our designs. I'm worried about hardcoding these we'll get a bunch of colors that slightly differ from one another 😅

src/orders-v2/order-detail-page.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { OrderDetailsFragment } from "@dashboard/graphql";
+import React from "react";
+
+import { OrderHeader } from "./order-header";
+
+export const OrderDetailsPage = ({ order }: { order: OrderDetailsFragment }) => {
+  return (
+    <>
+      {/* __height="100vh" - Disables savebar space */}
+      <DetailPageLayout backgroundColor="default2" padding={4} __height="100vh">
```
witoszekdev: suggestion: Maybe we could make it an explicit prop, like so:


```suggestion
      <DetailPageLayout backgroundColor="default2" padding={4} disableSavebarSpace>
```

.storybook/preview.tsx
```diff
@@ -15,13 +16,15 @@ if (typeof window !== "undefined") {
...
   decorators: [
     Story => (
-      <IntlProvider locale="en" onError={() => {}}>
-        <LegacyThemeProvider overrides={themeOverrides} palettes={paletteOverrides}>
-          <ThemeProvider>
-            <Story />
-          </ThemeProvider>
-        </LegacyThemeProvider>
-      </IntlProvider>
+      <Router>
```
witoszekdev: suggestion: I don't think we need full-blown router here since it won't work anyway. I guess the only reason we need it is because we have some hooks from react-router or Links in our components.

In that case maybe we could use `MemoryRouter` instead: https://v5.reactrouter.com/web/api/MemoryRouter

We can also try mocking modules instead (when navigating nothing happens 😄) this could be even faster: https://storybook.js.org/docs/writing-stories/mocking-data-and-modules/mocking-modules

locale/defaultMessages.json
```diff
@@ -6378,6 +6381,9 @@
     "context": "submit button",
     "string": "Add manual capture"
   },
+  "aGx3Z5": {
+    "string": "Created {formattedDate}"
+  },
```
witoszekdev, line 6386: suggestion: I think we already have similar strings in our translations, we could reuse them instead of creating new ;)

src/components/FilterPresetsSelect/FilterPresetItem.tsx
```diff
@@ -49,6 +49,7 @@ export const FilterPresetItem = ({
           >
             <RemoveIcon
               data-test-id="preset-delete-button"
+              // @ts-expect-error - check why style is not accepted in types, but can be used
```
witoszekdev, line 52: tip: This was fixed with last macaw ui version: https://github.com/saleor/macaw-ui/commit/eca6ef88442d76fbbd4b925cc3ae07e3c748b417

src/orders-v2/fixtures.ts
witoszekdev, line 1: issue (non blocking for PR): These looks dreadful 🙈.  We already used to have hundreds of fixtures for our stories in Dashboard. I think this is fine for this PR, but we probably should think about something else for the future so we don't get back to our old problem.

I have some rough idea, we can discuss this later to use some sort of auto generation for GraphQL fragments using https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-mock-data + https://the-guild.dev/graphql/tools/docs/mocking#mocking-a-schema-using-introspection and perhaps mock-service-workers (https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-msw) / [utility for Apollo](https://github.com/apollographql/graphql-testing-library). We could also try https://www.graphql-mocks.com/docs/guides/fake-data/

src/config.ts
```diff
@@ -6,7 +6,8 @@ export const getAppDefaultUri = () => "/";
 export const getAppMountUri = () => window?.__SALEOR_CONFIG__?.APP_MOUNT_URI || getAppDefaultUri();
 export const getApiUrl = () => window.__SALEOR_CONFIG__.API_URL;
 export const SW_INTERVAL = parseInt(process.env.SW_INTERVAL ?? "300", 10);
-export const IS_CLOUD_INSTANCE = window.__SALEOR_CONFIG__.IS_CLOUD_INSTANCE === "true";
+// TODO: figure out why this is needed in storybook
+export const IS_CLOUD_INSTANCE = window.__SALEOR_CONFIG__?.IS_CLOUD_INSTANCE === "true";
```
witoszekdev: This adds button in our navigation and enables posthog. In storybook it should be set to false

witoszekdev (dismissed):

.storybook/preview.tsx
lkostrowski: We will need router to navigate to nested pages (say order/refunds) but I agree that probably memory router should can be considered. Not sure how storybook recommends that

locale/defaultMessages.json
lkostrowski: I would be careful with that, because with reusing, you lose context.

package.json
lkostrowski: Let's instantly deprecate all other icons

src/components/AppLayout/AppLayout.tsx
```diff
@@ -12,6 +12,7 @@ import { useStyles } from "./styles";
 interface AppLayoutProps {
   children: React.ReactNode;
   fullSize?: boolean;
+  disableBorder?: boolean;
```
lkostrowski: I think it's better to swap it

```
border?: boolean
```

It's better to pass `border: false` than `disableBorder: true` imo

src/orders-v2/fixtures.ts
lkostrowski: I think we should group them to logical partials. For example, all prices can be a single fixture, but transactions can be another one. It will make us easier to compose, eg 

```
const fixture = new OrderFixture.fulfilled().withGrantedRefund().withEvents([...]).build()
```

src/orders-v2/mocked-app-layout.tsx
```diff
@@ -0,0 +1,33 @@
+import { Box } from "@saleor/macaw-ui-next";
+import React from "react";
+
+export const MockedAppLayout = ({ children }: { children: React.ReactNode }) => {
```
lkostrowski: Why mocked? Is it used for storybook? Can we keep "mocked" stuff in some __tests__ or "storybook" folder?

src/orders-v2/order-detail-page.stories.tsx
lkostrowski, line 1: can we keep it in separate folder like `orders-v2/storybook`?

src/orders-v2/order-header.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { Box, Button, Text } from "@saleor/macaw-ui-next";
+import { Package } from "lucide-react";
+import moment from "moment-timezone";
+import React from "react";
+import { useIntl } from "react-intl";
+
+import { StatusPill } from "./status-pill";
+import { UnderlineLink } from "./underline-link";
+
+interface Props {
```
lkostrowski: let's ensure all components extends `BoxProps`, it's a basic composition that allows us to position components by parents

src/orders-v2/order-header.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+export const OrderHeader = ({ status, orderNumber, created, channel }: Props) => {
+  const intl = useIntl();
+
+  const { localized, status: orderStatus } = transformOrderStatus(status, intl);
+
+  const formattedDate = moment(created).format("Do MMM, YY HH:mm");
+
+  return (
+    <>
```
lkostrowski: wrap with <Box> so we can spread BoxProps

src/orders-v2/order-header.tsx
```diff
@@ -0,0 +1,85 @@
...
+    name: string;
+  };
+}
+
+export const OrderHeader = ({ status, orderNumber, created, channel }: Props) => {
+  const intl = useIntl();
+
+  const { localized, status: orderStatus } = transformOrderStatus(status, intl);
+
+  const formattedDate = moment(created).format("Do MMM, YY HH:mm");
```
lkostrowski: Why don't you use Intl? We should probably deprecate moment.js

src/orders-v2/status-pill.tsx
```diff
@@ -0,0 +1,39 @@
...
+}) => {
+  const getFillColorsBasedOnStatus = (status: PillStatusType) => {
+    switch (status) {
+      case "success":
+        return {
+          __backgroundColor: "oklch(94% 0.0898 164deg)",
+          __borderColor: "oklch(79% 0.204 153deg)",
+        };
+    }
+  };
```
lkostrowski: small optimization but if you move it outisde react component, it will not get re-instantionated on every react render

src/orders-v2/underline-link.tsx
```diff
@@ -0,0 +1,16 @@
+import { Text, TextProps } from "@saleor/macaw-ui-next";
+import React from "react";
+import { Link } from "react-router-dom";
+
+interface Props extends TextProps {
+  href: string;
+  children: React.ReactNode;
+}
+
+export const UnderlineLink = ({ href, children, ...props }: Props) => (
```
lkostrowski, line 11: probably also macaw-candidate?

lkostrowski (changes requested):

src/orders-v2/order-detail-page.tsx
kzuraw: I will remove it from this page

src/orders-v2/fixtures.ts
kzuraw: I will create a linear issue for order fixture factories

src/config.ts
kzuraw: I will check that as it seems like storybook is not properly loading custom override for window in `preview.tsx`

src/orders-v2/mocked-app-layout.tsx
kzuraw: I will rename it `StorybookAppLayout` and I will move it to `stories` folder

src/orders-v2/order-header.tsx
kzuraw: I will make sure that other `Date` components are deprecated as well + moment.js as dep in eslint

src/orders-v2/fixtures.ts
kzuraw: Done in https://linear.app/saleor/issue/ENG-575

src/components/AppLayout/AppLayout.tsx
kzuraw: turns out this was leftover and it is not needed now

package.json
kzuraw: Yes - to be tracked in https://linear.app/saleor/issue/ENG-447

src/components/FilterPresetsSelect/FilterPresetItem.tsx
kzuraw: It didn't work in https://github.com/saleor/macaw-ui/releases/tag/v1.2.0 - we will need to take a look why - I will do it in another issue: https://linear.app/saleor/issue/ENG-579

lkostrowski (approved):

PR #5766, opened 2025-08-04
feat: add ESLint rule to deprecate Material-UI and Macaw-UI icons

krzysztofzuraw:
- Add new ESLint rule 'no-deprecated-icons' to warn against importing icons from @material-ui/icons and @saleor/macaw-ui-next
- Add deprecation JSDoc comments to all existing icon components with migration guidance to lucide-react
- Configure ESLint to use the new rule as a warning
- Update 34 icon files with deprecation notices and usage examples

This helps migrate away from deprecated UI libraries towards lucide-react icons.

## Changes
- **New ESLint rule**: `lint/rules/no-deprecated-icons.js` - warns when importing icons from deprecated libraries
- **ESLint config**: Added the rule as a warning in `.eslintrc.json`
- **Icon deprecations**: Added JSDoc deprecation notices to 34 icon components with migration examples
- **Rule registration**: Updated `eslint-local-rules.js` to include the new rule

## Migration Path
Developers will now see warnings when importing icons from:
- `@material-ui/icons`
- `@saleor/macaw-ui-next` (icon imports)

The warnings guide them to use `lucide-react` icons instead.

### DIFF
--- a/lint/rules/no-deprecated-icons.js
+++ b/lint/rules/no-deprecated-icons.js
@@ -1,0 +1,60 @@
+module.exports = {
+  create: context => ({
+    ImportDeclaration: node => {
+      let iconImports = [];
+      let messageId = "";
+      let libraryName = "";
+
+      // Check if importing from @saleor/macaw-ui-next
+      if (node.source.value === "@saleor/macaw-ui-next") {
+        iconImports = node.specifiers.filter(specifier => {
+          return specifier.type === "ImportSpecifier" && specifier.imported.name.endsWith("Icon");
+        });
+        messageId = "noMacawUiIcons";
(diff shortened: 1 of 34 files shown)


lint/rules/no-deprecated-icons.js
```diff
@@ -0,0 +1,60 @@
+module.exports = {
```
lkostrowski, line 1: really you need custom rule? We have rule for restricted imports

kzuraw: you can have `import { Box, DeprecatedIcon } from '@macaw'` and you want warn only for `DeprecatedIcon` not `Box` itself - hence why I've added custom rule.

PR #5767, opened 2025-08-04
Remove unused GitHub workflows for cypress when releasing / auto-pick

krzysztofzuraw:
## 🧹 Cleanup: Remove Unused GitHub Workflows

This PR removes unused workflow files and scripts that were identified as no longer needed.

#### Workflows:
  
- **`deploy-and-run-cypress.yml`** - Manual testing workflow
  - ❌ Very infrequent usage (last run Feb 2025)
  - ❌ Rarely used manual workflow
  
- **`auto-pick`** - Picking changes for release

#### Supporting Scripts:
- **`cleanEnvironments.js`** - Environment cleanup script
  - ❌ Only used by removed clean-envs workflow
  - ❌ No other references found
  
- **`postTestsResults.js`** - Test results posting script
  - ❌ Orphaned script with no references in any workflow files

### Analysis Method:
- ✅ Checked workflow run history using `gh run list`
- ✅ Analyzed git commit history for maintenance activity
- ✅ Searched codebase for references to ensure safe removal
- ✅ Verified remaining files are actively used

### Impact:
- 📉 **518 lines of code removed**
- 🧹 **Cleaner workflow directory**
- ⚡ **Reduced maintenance overhead**
- ✅ **No impact on active workflows**

All remaining workflow files and scripts are actively referenced and needed for current operations.


.github/workflows/auto-pick.yml
witoszekdev, line 1: damn i didn't know we had that, would have been usefull for releases past 3.21 🫤

package.json
```diff
@@ -84,7 +76,6 @@
     "@material-ui/lab": "^4.0.0-alpha.61",
     "@material-ui/styles": "^4.11.4",
     "@radix-ui/react-portal": "^1.1.4",
-    "lucide-react": "^0.534.0",
```
witoszekdev, line 87: Was this an accident? Why did we remove it in this PR?

.github/actions/get-env-vars/action.yml
witoszekdev, line 1: We cannot delete this, it's still used in `run-tests-on-release.yaml`, line 127

```yaml
      - name: Get environment variables
        uses: ./.github/actions/get-env-vars
        id: get-environment-variables
        with:
          version: ${{ env.CUSTOM_VERSION }}
          token: ${{ secrets.CLOUD_ACCESS_TOKEN }}
          repo_token: ${{ secrets.GITHUB_TOKEN }}
          project: ${{ env.PROJECT }}
```

and line 61

```
      IS_OLD_VERSION: ${{ steps.get-environment-variables.outputs.IS_OLD_VERSION }}

```

.github/workflows/run-tests-on-release.yml
witoszekdev, line 1: We need to remove this since we're no longer deciding wether to use cypress vs playwright:

```
      FRAMEWORK: ${{ steps.check-framework.outputs.result }}
```

witoszekdev (changes requested):
FYI i asked codex to also try to remove Cypress workflows, maybe we should use this instead: https://github.com/witoszekdev/saleor-dashboard/pull/1/files

witoszekdev:
One more thing:

```
          SALEOR_CLOUD_SERVICE: ${{ needs.initialize-cloud.outputs.SALEOR_CLOUD_SERVICE }}
```

there's no `initialize-cloud` step, maybe we should check how it worked previously? maybe another action is outputting this VAR, e.g. `needs.add-check-and-prepare-instance.outputs.SALEOR_CLOUD_SERVICE`

witoszekdev:
There's also reference to `send-slack-message` but we don't have such job

package.json
kzuraw: Lucide react was not removed it is still in package.json but sorted - 
2025-08-05 at 15 36 51 - CleanShot@2x

.github/workflows/auto-pick.yml
kzuraw: I runs on main on every merge but it is doing nothing so I don't think there is a need to keep it

kzuraw:
Closing in favour of https://github.com/saleor/saleor-dashboard/pull/5777

PR #5768, opened 2025-08-05
Update MacawUI to 1.3.0

krzysztofzuraw:
## Scope of the change

No functional changes - it only deprecates icons via https://github.com/saleor/macaw-ui/releases/tag/v1.3.0


package.json
```diff
@@ -84,11 +84,10 @@
     "@material-ui/lab": "^4.0.0-alpha.61",
     "@material-ui/styles": "^4.11.4",
     "@radix-ui/react-portal": "^1.1.4",
-    "lucide-react": "^0.534.0",
```
witoszekdev: question: shouldn't we have lucide installed if it's a peer dependency for macaw?

kzuraw: 2025-08-05 at 15 49 21 - CleanShot@2x
it is there - but sorted 😄

PR #5769, opened 2025-08-05
copy id quick action

lkostrowski:

### DIFF
--- a/src/components/NavigatorSearch/useActionTriggers.tsx
+++ b/src/components/NavigatorSearch/useActionTriggers.tsx
@@ -75,14 +75,14 @@ interface TriggerDescriptor {
 
 const allActions: TriggerDescriptor[] = [
   {
-    section: {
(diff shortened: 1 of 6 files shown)


src/components/NavigatorSearch/useActionTriggers.tsx
```diff
@@ -33,373 +59,407 @@ const ActionLinkItem = ({ href, children }: { href: string; children: React.Reac
...
-    onClick?: (event: React.MouseEvent<HTMLAnchorElement>) => void;
+    onAction: (cb: Function, event: React.MouseEvent<HTMLAnchorElement | HTMLDivElement>) => void;
   }>;
+  id?: string;
 }
 
 const allActions: TriggerDescriptor[] = [
   {
-    section: {
+    section: defineMessage({
```
lkostrowski, line 77: We need to add defineMessage so eslint works

PR #5776, opened 2025-08-05
Add order details right sidebar

krzysztofzuraw:
## Scope of the change

2025-08-05 at 15 34 25 - CleanShot@2x


src/orders-v2/underline-link.tsx
```diff
@@ -10,7 +10,13 @@ interface Props extends TextProps {
 // TODO: move to MacawUI
 export const UnderlineLink = ({ href, children, ...props }: Props) => (
```
lkostrowski, line 11:
```
type Props extends LinkProps {
  textProps?: TextProps
}

```

src/orders-v2/underline-link.tsx
```diff
@@ -10,7 +10,13 @@ interface Props extends TextProps {
...
   <Link to={href}>
-    <Text as="span" textDecoration="underline" {...props} data-macaw-ui-candidate>
```
lkostrowski, line 13: lets export two components from this file
1. UnderlineText which is not a link (so we have low level API to display something and control it's routing / clicking)
2. UnderlinkLink which is a higher level component that includes router Link inside

props should be spreaded on top level node

lkostrowski (approved):

PR #5777, opened 2025-08-05
Remove Cypress leftovers in GitHub workflows, npm scripts

witoszekdev:
## Scope

This PR removes Cypress which was dropped in 3.19 from our GitHub actions and scripts.

We need further work on our workflows - because this change doesn't fully remove dead code:
- there are references to Slack workflow which no longer exists
- there are still jobs to decide which tool (cypress vs playwright) should be used, left for compatibility reasons (needs to be done in separate PR)


krzysztofzuraw (dismissed):
Looks good - fix package.json conflict and it is ready to be merged

krzysztofzuraw (approved):

PR #5778, opened 2025-08-05
Update react to 18

lkostrowski:
This PR upgrades React and React Dom to v18

It bascially changes 2 things
- How react dom is rendered (createRoot)
- Removal of React.FC usage, due to breaking change in types

e2e tests are failing, but not due to this change, they are flaky. Tested manually and seems to work

I made typescript green, which mean some "Stupid" things around legacy packages (old macaw, icons). Some ts-ignore and type assertions are added, but will be refactored in the future

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,10 +59,6 @@
       "schemaVersion": "main"
     }
   },
-  "resolutions": {
-    "@types/react": "18.3.23",
-    "@types/react-dom": "18.3.7"
-  },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -59,6 +59,10 @@
       "schemaVersion": "main"
     }
   },
+  "resolutions": {
```
kzuraw: **Question:** do we need that? I mean we have types updated as well to proper versions

PR #5784, opened 2025-08-07
Restrict legacy packages

lkostrowski:


eslint.config.mjs
```diff
@@ -153,27 +152,19 @@ export default tseslint.config(
...
       "no-console": ["error", { allow: ["warn", "error"] }],
       "no-restricted-imports": [
         "error",
         {
-          paths: [
-            "lodash",
-            "@material-ui/icons/Delete",
-            "classnames",
-          ],
+          paths: ["lodash", "@material-ui/icons/Delete", "classnames"],
```
kzuraw: do we need `@material-ui/icons/Delete` here? maybe we can remove it as below we have warn for `"@material-ui/*"`

lkostrowski: this line is "error". but i can change it

PR #5786, opened 2025-08-07
Fix attribute filtering on Products list

witoszekdev:
Reference attribute filtering was broken after GraphQL schema changes in saleor/saleor#18016. The dashboard was using the deprecated `valueNames` field instead of the new `AssignedAttributeReferenceInput` structure.

Now we are using referenced object ids instead of names, which should prevent issues with duplicate entries.


witoszekdev:
I've checked failing e2e tests, they are due to a failure on `main` in vouchers: https://github.com/saleor/saleor-dashboard/actions/runs/16830336854

CleanShot 2025-08-08 at 17 07 12@2x
CleanShot 2025-08-08 at 17 07 18@2x

.changeset/fix-reference-attribute-filtering.md
```diff
@@ -0,0 +1,5 @@
+---
+"@saleor/dashboard": patch
+---
+
+Fix reference attribute filtering in products list page after GraphQL schema changes. Product filters using reference attributes (pages, products, variants) now work correctly with latest Saleor schema.
```
lkostrowski, line 5: this is unreleased so we don't need a changeset. I mean this broken build never went to production so this entry is a noise

lkostrowski (approved):

PR #5796, opened 2025-08-18
Add advanced CodeQL setup

witoszekdev:
This PR adds advanced CodeQL configuration and explicit workflow that triggers on changes to `main`, protected branches and on pull request to `main` and protected branches.

Previously when making PRs to version branches (e.g. 3.21) or to `main` CodeQL workflow would not run, even though it is required by branch protection rules.

Now we have exlpicit workflow that should prevent this.

This PR also adds configuration for CodeQL to exclude files not being used on production like Playwright tests.

Relevant: https://github.com/orgs/community/discussions/159026


witoszekdev:
@patrys This should fix our issues with running CodeQL for patches in Dashboard. Previously we had workflow being stuck on "waiting for status report", however we need to enable this manually in settings for this PR to work.

witoszekdev:
@patrys we still need to change settings in the repo to merge this: https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/configuring-advanced-setup-for-code-scanning#configuring-advanced-setup-for-code-scanning-with-codeql

> CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled

patrys:
I know but it has to be merged first (not sure if `main` is enough or if the advanced config is read from the target branch of each individual PR)

.github/codeql/codeql-config.yml
```diff
@@ -0,0 +1,34 @@
+name: "CodeQL Config"
+
+# Paths to exclude from CodeQL analysis
+paths-ignore:
```
kzuraw: maybe we can instead use `paths`? Thanks to that you can have:

```yml
paths:
- "src/**"
- "scripts/**"
```
instead of blacklist of files (also we don't commit `node_modules` and other paths here - I don't think they all are necessary)

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,30 @@
...
+  push:
+    branches: [ main, protected, "3.21" ]
+  pull_request:
+    branches: [ main, protected, "3.21" ]
+
+jobs:
+  analyze:
+    name: Analyze JavaScript/TypeScript
+    runs-on: ubuntu-latest
+    timeout-minutes: 360
```
kzuraw: what about 60? 360 is 6h of timeout 😄

patrys (approved):

PR #5802, opened 2025-08-26
New order lines with fulfillments

krzysztofzuraw:
## Scope of the change

This PR adds fulfillments and order lines sections to the new order page view:
2025-08-26 at 10 56 35 - CleanShot@2x


src/orders-v2/fixtures/order-fixture.ts
witoszekdev, line 1: this comment is not for this PR specifically but in general:

Do we really want to use fixtures for our storybook? I think when we used this previously it was a big headache for us, we had to make a lot of changes whenever we made a small change

I think for GraphQL mocking we could automate this a bit, to generate the mock data for us.

src/orders-v2/order-fulfillments/order-fulfillments.tsx
```diff
@@ -0,0 +1,116 @@
...
+                  <OrderFulfillmentTrackingNumber trackingNumber={fulfillment.trackingNumber} />
+                )}
+              </Box>
+              <Box display="flex" alignItems="center" gap={2}>
+                <Text size={6} fontWeight="medium">
+                  {intl.formatMessage(
+                    {
+                      defaultMessage:
+                        "{fulfillmentStatus, select, CANCELED {Canceled} FULFILLED {Fulfilled} REFUNDED {Refunded} REFUNDED_AND_RETURNED {Refunded and Returned} REPLACED {Replaced} RETURNED {Returned} WAITING_FOR_APPROVAL {Waiting for Approval} other {Other}}",
+                      id: "Re3mG5",
```
witoszekdev: suggestion: I think having a switch for different statuses would be better, since it's a pattern we already use and a bit more clear to read the translation string 😅

src/orders-v2/order-detail-page.tsx
```diff
@@ -17,7 +19,17 @@ export const OrderDetailsPage = ({ order }: { order: OrderDetailsFragment }) =>
...
+          __height="100%"
+          gridColumn="9"
+          borderColor="default1"
+          borderTopWidth={1}
+          borderLeftWidth={1}
+          borderRightWidth={0}
+          borderTopLeftRadius={5}
+          borderBottomWidth={0}
+          borderStyle="solid"
+        >
```
witoszekdev, line 32: question: shouldn't this be a new component or it's variant? It looks to me like we're overriding quite a lot for this new view 😄

src/orders-v2/order-lines.tsx
```diff
@@ -0,0 +1,43 @@
...
+              {intl.formatMessage({ id: "QDxJib", defaultMessage: "Fulfill" })}
+            </Button>
+          )}
+        </Box>
+      </Box>
+      <OrderDetailsDatagrid
+        lines={lines}
+        loading={false}
+        onShowMetadata={() => {
+          alert("Metadata show");
```
witoszekdev, line 43: issue: missing intl.formatMessage for string


```suggestion
          alert(intl.formatMessage({defaultMessage: "Metadata show"}));
```

witoszekdev (changes requested):
just one issue with intl, rest are just suggestions

src/orders-v2/order-lines.tsx
kzuraw: This alert will be removed anyway - `onShowMetadata` should render modal. I will add `TODO` comment to make sure we don't miss it

src/orders-v2/order-detail-page.tsx
kzuraw: This is planned for the future - for now I wanted to have all necessary structure of order page first

PR #5807, opened 2025-08-27
Fix links to old Saleor docs

krzysztofzuraw:

### DIFF
--- a/src/plugins/components/PluginsListPage/PluginsListPage.tsx
+++ b/src/plugins/components/PluginsListPage/PluginsListPage.tsx
@@ -64,7 +64,7 @@ const PluginsListPage = ({
               {intl.formatMessage(pluginsListPageMessages.appStoreWarning)}{" "}
               <ExternalLinkNext
                 target="_blank"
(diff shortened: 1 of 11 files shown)


src/plugins/components/PluginsListPage/PluginsListPage.tsx
```diff
@@ -64,7 +64,7 @@ const PluginsListPage = ({
               {intl.formatMessage(pluginsListPageMessages.appStoreWarning)}{" "}
               <ExternalLinkNext
                 target="_blank"
-                href="https://docs.saleor.io/docs/3.x/developer/app-store/overview"
+                href="https://docs.saleor.io/developer/app-store/overview"
```
witoszekdev, line 67: suggestion: maybe we should put all of these links into a single file to it's easier to track when stuff changes?

e.g. we already have: https://github.com/saleor/saleor-dashboard/blob/2ce03d136ca5740b92c02ed747856df07583bb48/src/links.ts

witoszekdev (approved):

PR #5808, opened 2025-08-29
Narrow down reference attributes 

IKarbowiak:
## Scope of the change
Implementation for https://github.com/saleor/saleor/pull/18084


src/products/views/ProductUpdate/handlers/utils.ts
```diff
@@ -248,3 +253,56 @@ export function inferProductChannelsAfterUpdate(
...
+
+export const useReferencePageSearch = (refAttr: any) => {
+  const ids = React.useMemo(
+    () => getAllowedReferenceTypeIds(refAttr, ReferenceType.PageType),
+    [refAttr],
+  );
+  const variables = React.useMemo(
+    () => buildReferenceSearchVariables(ids, ReferenceWhereKey.PageType),
+    [ids],
+  );
```
witoszekdev: suggestion: let's replace `React.useMemo` and just use `useMemo` imported directly instead

src/products/views/ProductUpdate/handlers/utils.ts
```diff
@@ -248,3 +253,56 @@ export function inferProductChannelsAfterUpdate(
...
+
+export const buildReferenceSearchVariables = (
+  allowedIds: string[] | undefined,
+  whereKey: ReferenceWhereKey,
+) => ({
+  ...DEFAULT_INITIAL_SEARCH_DATA,
+  ...(allowedIds?.length ? { where: { [whereKey]: { oneOf: allowedIds } } } : {}),
+});
+
+export const useReferenceProductSearch = (refAttr: any) => {
```
witoszekdev: issue: can we provide some type here? `any` disables any checks, maybe we can just use this:

```suggestion
export const useReferenceProductSearch = (refAttr: unknown) => {
```

src/products/views/ProductUpdate/handlers/utils.ts
```diff
@@ -248,3 +253,56 @@ export function inferProductChannelsAfterUpdate(
...
+  PageType = "pageType",
+}
+
+export const getAllowedReferenceTypeIds = (refAttr: any, type: ReferenceType): string[] => {
+  if (refAttr?.referenceTypes?.[0]?.__typename === type) {
+    return (refAttr.referenceTypes ?? []).map((t: any) => t?.id).filter(Boolean);
+  }
+
+  return [];
+};
```
witoszekdev: same here let's not use `any` for `refAttr` this way we won't another one when filtering

```suggestion
export const getAllowedReferenceTypeIds = (refAttr: <some_type_here>, type: ReferenceType): string[] => {
  if (refAttr?.referenceTypes?.[0]?.__typename === type) {
    return (refAttr.referenceTypes ?? []).map((t) => t?.id).filter(Boolean);
  }

  return [];
};
```

witoszekdev (commented):
LGTM, we just need to get rid of `any`

We can also change this PR to target `main` branch, let's not create feature branches, since we want to release this in 3.22

witoszekdev:
@IKarbowiak Is there a way I can test this out manually? I don't know how to setup `referenceTypes` since we don't have UI yet. Could you please setup such attribute in the testing environment for this PR: https://pr-5808.dashboard.saleor.rocks/ 🙏🏻

witoszekdev (approved):

PR #5815, opened 2025-09-04
Improve cancel fulfilment 

IKarbowiak:
## Scope of the change

When a waiting for approval fulfilment is cancelled, providing a warehouse is not required. The backend ignores the warehouse in this case, so I omitted it.

Cancel fulfilment with `waiting for approval` status
image

Cancel for fulfilments with another status:
image


.changeset/sharp-shrimps-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Improve order fulfillment cancel dialog; in case the fulfillment is in waiting for approval state, do not require providing a warehouse
```
witoszekdev: nitipck: we can reword this a bit, since it's used for generating changelog

```suggestion
Improved order fulfillment cancel dialog: in case the fulfillment is in waiting for approval state, providing a warehouse will be no longer required
```

IKarbowiak: Done

witoszekdev (approved):

witoszekdev:
@IKarbowiak We need to run `npm run extract-messages` + commit files, othwerwise build will fail; already fixed that ;)

PR #5816, opened 2025-09-04
Allow defining reference types

IKarbowiak:
## Scope of the change
Part of the implementation for https://github.com/saleor/saleor/pull/18084

- Allow defining the `referenceTypes` in `AttributeCreate` and `AttributeUpdate` page only for `reference` attributes
- Allow defining and searching `product types` for `product` and `variant` reference attributes
- Allow defining and searching `page types` for `page` reference attributes

image

image

image

### DIFF
--- a/locale/defaultMessages.json
+++ b/locale/defaultMessages.json
@@ -6520,10 +6520,6 @@
   "aMwxYb": {
     "string": "Countries"
   },
-  "aOZXYd": {
-    "context": "helper explaining the effect of assigning reference types",
-    "string": "Assigning product types narrows the list of items this attribute can reference."
-  },
   "aOelhW": {
     "context": "tab name",
     "string": "All Plugins"


src/attributes/components/AssignReferenceTypesDialog/AssignReferenceTypesDialog.tsx
```diff
@@ -0,0 +1,39 @@
...
+        confirmBtn: intl.formatMessage(messages.confirmBtn),
+        ...labels,
+      }}
+      {...rest}
+    />
+  );
+};
+
+AssignReferenceTypesDialog.displayName = "AssignReferenceTypesDialog";
+export default AssignReferenceTypesDialog;
```
lkostrowski: we are in the middle of the migration, but please do not use `export default`, it's an anti-pattern

instead just do `export const AssignReferenceTypesDialog = ...`

(ESLint should warn about this)

src/attributes/components/AssignReferenceTypesDialog/index.ts
lkostrowski, line 1: This file is also not needed. We don't have static lint rule for that, but we remove barrel files (another antipattern). This leads to complicated resolving, circular dependencies and worse DX in general.

Instead, just import directly from src/attributes/components/AssignReferenceTypesDialog/AssignReferenceTypesDialog.tsx

src/attributes/components/AttributeReferenceTypesSection/AttributeReferenceTypesSection.tsx
```diff
@@ -0,0 +1,75 @@
...
+    inputType,
+    entityType,
+    selectedTypes,
+    disabled,
+    onAssignClick,
+    onRemoveType,
+}) => {
+    const intl = useIntl();
+    
+    if (!canShow(inputType, entityType)) return null;
```
lkostrowski: maybe this check should happen in the parent component? I think it's clearer than instantionate this, just to return nothing

src/components/ChipField/ChipField.tsx
```diff
@@ -0,0 +1,78 @@
...
+  label: ReactNode;
+  onClose?: () => void;
+  loading?: boolean;
+  url?: string;
+} & BoxProps;
+
+export const ChipField = React.forwardRef<HTMLDivElement, ChipFieldProps>(
+  (
+    { label, onClose, loading, url, className, style, ...props },
+    ref,
```
lkostrowski: is ref actually being used?

IKarbowiak: `ref` is used in `Box` component

src/attributes/components/AssignReferenceTypesDialog/AssignReferenceTypesDialog.tsx
IKarbowiak: Done

src/attributes/components/AssignReferenceTypesDialog/index.ts
IKarbowiak: Done

src/attributes/components/AttributeReferenceTypesSection/AttributeReferenceTypesSection.tsx
IKarbowiak: Done

src/attributes/components/AttributePage/AttributePage.tsx
```diff
@@ -1,5 +1,6 @@
-import { attributeListPath } from "@dashboard/attributes/urls";
-import { ATTRIBUTE_TYPES_WITH_DEDICATED_VALUES } from "@dashboard/attributes/utils/data";
+import {AssignReferenceTypesDialog, ReferenceTypes} from "@dashboard/attributes/components/AssignReferenceTypesDialog/AssignReferenceTypesDialog";
```
witoszekdev: issue: Please run `npm run lint` to fix these formatting issues (spaces in curly braces are different than the rest, wrong spacing in some files)

src/attributes/components/AttributePage/AttributePage.tsx
```diff
@@ -148,11 +166,57 @@ const AttributePage = ({
...
+          activeRefSearch.result.data?.search,
+        );
+    
+        const fetchMoreReferenceTypes = getSearchFetchMoreProps(
+          activeRefSearch.result as CommonSearchOpts,
+          activeRefSearch.loadMore,
+        );
+
+        // Clear reference types in case entityType changes, as it may affect available options
+        const handleChange = (eventOrName: any, value?: any) => {
```
witoszekdev: issue: This method has incorrect signature `AttributeDetails` component has `onChange` prop with type:

```ts
  onChange: (event: ChangeEvent) => void;
```

but this is later on used in input which has type:

```ts
  onChange?: React.ChangeEventHandler<HTMLTextAreaElement | HTMLInputElement>;
```

so we need to:
- update types in `AttributeDetails`
- remove `any` from `handleChange`

src/attributes/components/AttributePage/AttributePage.tsx
```diff
@@ -148,11 +166,57 @@ const AttributePage = ({
...
+          const fieldName =
+            typeof eventOrName === "string" ? eventOrName : eventOrName?.target?.name;
```
witoszekdev: issue: `eventOrName` will never have value `string` since it's an event object

src/attributes/components/AttributePage/AttributePage.tsx
```diff
@@ -243,6 +315,23 @@ const AttributePage = ({
                 />
               </Savebar>
             </DetailPageLayout>
+            <AssignReferenceTypesDialog
+              key={data.referenceTypes.map(o => o.value).sort().join("|")}
```
witoszekdev: question: Why do we need this? When `key` changes React will unmount this dialog and mount a new one with fresh state, but I don't think we need that here?

This can be fine, but we should add a comment why we need `key` (e.g. because we need to completely rerender dialog because we now choose different type and we have some side-effects). However this can also signal that we might have a bug in the modal itself because it doesn't clean up properly.

One more thing: if we need `key` it should be in `useMemo` so that we don't re-calculate it's value on each rerender

src/attributes/components/AttributePage/AttributePage.tsx
```diff
@@ -243,6 +315,23 @@ const AttributePage = ({
...
+              open={params.action === "assign-reference-types"}
+              onClose={onCloseAssignReferenceTypes}
+              confirmButtonState={"default"}
```
witoszekdev: question: If this stays as a constant value `default` imho we can remove this prop and just hardcode it inside the Dialog itself

src/attributes/components/AttributeReferenceTypesSection/messages.tsx
```diff
@@ -0,0 +1,31 @@
...
+    id: "UvZuf7",
+    defaultMessage: "Reference Types",
+    description: "section title for reference types selector",
+  },
+  referenceTypesHelper: {
+    id: "aOZXYd",
+    defaultMessage:
+      "Assigning product types narrows the list of items this attribute can reference.",
+    description: "helper explaining the effect of assigning reference types",
+  },
```
witoszekdev: question: What's the difference between this message and `referenceTypesHelp`? I think this one is more clear, but we would probably need another one for model types afaik

src/components/ChipField/ChipField.tsx
witoszekdev: (response to previous comment) I think we don't actually need this, `ref` was used in `SortableChip` because drag and drop library does some magic with it so that users can drag and drop the chips, but here we don't want anyone to use it with drag and drop 😄 

`ref` on Box is optional, we don't actually have to provide it

witoszekdev:
@IKarbowiak I updated schema in this PR because it wouldn't build due to differences with Saleor (on feature branch we have new fields in schema)

src/attributes/components/AttributeReferenceTypesSection/messages.tsx
IKarbowiak: I delete this one

src/components/ChipField/ChipField.tsx
IKarbowiak: Removed, thanks!

src/attributes/components/AttributePage/AttributePage.tsx
IKarbowiak: Done

src/attributes/components/AttributePage/AttributePage.tsx
IKarbowiak: Fixed

src/attributes/components/AttributePage/AttributePage.tsx
IKarbowiak: Fixed

src/attributes/components/AttributePage/AttributePage.tsx
IKarbowiak: `key` deleted

PR #5810, opened 2025-09-02
Eng 658 stock section improv

krzysztofzuraw:
## Scope of the change

Improved stock section when editing product variant. After this change `Assign Warehouse` button will open modal with warehouse select instead of opening dropdown with the list.

PR #5821, opened 2025-09-08
Improve stock selection on product variant

krzysztofzuraw:
## Scope of the change

Backport of https://github.com/saleor/saleor-dashboard/pull/5810

### DIFF
--- a/src/products/components/ProductStocks/components/ProductStocksAssignWarehouses.tsx
+++ b/src/products/components/ProductStocks/components/ProductStocksAssignWarehouses.tsx
@@ -1,16 +1,24 @@
+import BackButton from "@dashboard/components/BackButton";
+import Checkbox from "@dashboard/components/Checkbox";
+import { InfiniteScroll } from "@dashboard/components/InfiniteScroll";
+import { DashboardModal } from "@dashboard/components/Modal";
+import ResponsiveTable from "@dashboard/components/ResponsiveTable";
+import TableRowLink from "@dashboard/components/TableRowLink";
 import { WarehouseFragment } from "@dashboard/graphql";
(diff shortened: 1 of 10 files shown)


src/products/components/ProductStocks/components/ProductStocksAssignWarehouses.tsx
```diff
@@ -1,81 +1,143 @@
...
-    // In case there is no more warehouses to assign, load more warehouses
-    // Example: from first page all warehouse has been already assigned, so we need to load more warehouses to assign
-    if (hasMoreWarehouses && !warehousesToAssign.length) {
-      loadMoreWarehouses();
-    }
-  }, [hasMoreWarehouses, warehousesToAssign.length]);
+  const [warehouses, setWarehouses] = useState<Option[]>([]);
+  const intl = useIntl();
+  const [open, setOpen] = useState(false);
+  const [query, onQueryChange, queryReset] = useSearchQuery(searchWarehouses);
```
lkostrowski, line 35: for the future, I would consider avoiding distributed states for the same thing. If our logic is
- option A: Modal closed, nothing in context
- option B: Modal open, warehouse(s) in context

Then we have finite state machine and option C & D (like open modal without context) are invalid

in such case we can model

```
class ModalOpenState {
  constructor(wh: Warehouse[]) {}

  readonly type = "modal_open"
  wh: Warehouse[]
}

class ModalClosedState {
  type= "modal_closed"
}

type PossibleStates = ModalOpenState | ModalCloseState

const [state, setState] = useState<PossibleStates>(new ModalOpenState([...]))
```

This may be a little more verbose but it's more readalbe in maintainable, also less buggy, especially for complex parts

PR #5826, opened 2025-09-08
Fix reference filtering and add message for empty state in search views for reference attributes

IKarbowiak:
## Scope of the change

- Fix filtering by categories and collection in reference attributes for product, variant and model views
- Add empty states for filtering references

Empty states:
image

image

image

image

### DIFF
--- a/locale/defaultMessages.json
+++ b/locale/defaultMessages.json
@@ -4488,6 +4488,10 @@
     "context": "link",
     "string": "Activate"
   },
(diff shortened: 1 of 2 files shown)


IKarbowiak:
Skipping change-set as the collection and category reference support has been added to main version only, and already mentioned in change set ([related PR](https://github.com/saleor/saleor-dashboard/pull/5698)).

PR #5828, opened 2025-09-08
Add filters for Order fulfilment warehouse

witoszekdev:
This PR adds `fulfilments.warehouse` filtering to Orders list:

CleanShot 2025-09-09 at 15 50 53@2x


src/components/ConditionalFilter/API/providers/OrderFilterAPIProvider.tsx
```diff
@@ -72,6 +73,10 @@ const createAPIHandler = (
     return new LegacyChannelHandler(client, inputValue);
   }
 
+  if (rowType === "fulfillmentWarehouse") {
+    return new WarehouseHandler(client, inputValue);
+  }
```
lkostrowski, line 78: suggestion & non blocking: this entire block would be more readable with switch/case block

PR #5837, opened 2025-09-10
Fix product stock warehouse assign

kzuraw:
## Scope of the change

Fixed product stocks section for variant details:

- After this change searching for warehouses to assign won't trigger re-rendering of stocks section.
- Added empty state message if there is no warehouses left to choose.
- Assign warehouse button is now hidden if product stocks section is rendered in product variant create page.

### DIFF
--- a/src/products/components/ProductStocks/ProductStocks.tsx
+++ b/src/products/components/ProductStocks/ProductStocks.tsx
@@ -220,7 +220,7 @@ export const ProductStocks: React.FC<ProductStocksProps> = ({
           onWarehouseSelect={handleWarehouseStockAdd}
           loading={loading}
           searchWarehouses={searchWarehouses}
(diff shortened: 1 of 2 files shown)


src/products/components/ProductStocks/ProductStocks.tsx
```diff
@@ -222,6 +220,7 @@ export const ProductStocks: React.FC<ProductStocksProps> = ({
           onWarehouseSelect={handleWarehouseStockAdd}
           loading={loading}
           searchWarehouses={searchWarehouses}
+          hideAssignWarehousesButton={isCreate}
```
lkostrowski: I think in general it's better to avoid negations. Non blocking but I would do 

```
showAssignWarehousesButton={!isCreate}
```

lkostrowski (approved):

PR #5840, opened 2025-09-11
Fix product stock warehouse assign

kzuraw:
## Scope of the change

Backport of https://github.com/saleor/saleor-dashboard/pull/5837

### DIFF
--- a/.changeset/fix-order-payment-refunds-safety.md
+++ b/.changeset/fix-order-payment-refunds-safety.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Fixed null safety issues in OrderPayment component when handling gift card amounts and improved variable naming consistency. This prevents potential runtime errors when order data is not fully loaded.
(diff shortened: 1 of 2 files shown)


.changeset/soft-rice-compete.md
```diff
@@ -0,0 +1,9 @@
+---
+"saleor-dashboard": patch
+---
+
+Fixed product stocks section for variant details:
```
witoszekdev: question: If this was already merged to `3.21` and is in the changelog do we want to also add changeset for `3.22`? 🤔

locale/defaultMessages.json
```diff
@@ -2179,6 +2179,10 @@
     "context": "webhooks and events section name",
     "string": "Webhooks & Events"
   },
+  "BHr3fj": {
```
witoszekdev, line 2182: question: how this link actually works? in other messages we've build we have a pattern to leave `{{  }}` and format it using a React component, e.g.:

```
  "408KSO": {
    "context": "gift card history message",
    "string": "Gift card was used as a payment method on order {orderLink}"
  },

```

IMHO this is much better, because messages don't have to know anything about HTML, and are less prone to issues if someone forgets to format this properly in other language translation

.changeset/soft-rice-compete.md
kzuraw: I will remove it - it doesn't make sense to keep it

locale/defaultMessages.json
kzuraw: Afaik `<a>` has used here to allow text inside - if you use `{orderLink}` you won't get ability to:

```tsx
<FormattedMessage
  {...messages.configureWarehouseForProduct}
  values={{
    a: chunks => <Link onClick={onWarehouseConfigure}>{chunks}</Link>,
  }}
/>
```

We can change that but then it should be done in different PR with ports to 3.21 as well - in this PR I changed only copy of text.

PR #5841, opened 2025-09-11
Improve CI/CD and husky workflow for testing code

kzuraw:


.github/workflows/test.yml
```diff
@@ -1,10 +1,15 @@
...
+on:
+  push:
+    branches: [main]
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review, labeled]
 
 jobs:
   check-lock:
     runs-on: ubuntu-latest
+    if: github.event_name == 'push' || (github.event.pull_request.draft == false || contains(github.event.pull_request.labels.*.name, 'Run QA checks'))
```
witoszekdev, line 12: suggestion: I think we could use YAML anchors to keep this condition in one place and reuse for each job:

e.g.
```yaml
jobs:
  .job_conditions: &job_condition
    if: github.event_name == 'push' || (github.event.pull_request.draft == false || contains(github.event.pull_request.labels.*.name, 'Run QA checks'))

 check-lock:
    runs-on: ubuntu-latest
    # Use the alias (*) to insert the condition
    <<: *job_condition
    steps:
      - uses: actions/checkout@v4 # Note: v2 is outdated
      - name: Validate lock file
        run: |
          npx lockfile-lint --path package-lock.json --allowed-hosts npm yarn

# ...
```

kzuraw: Looks interesting but I'm wondering if it won't mean worse readability so I suggest we stick with repetition for now

PR #5835, opened 2025-09-09
WIP: Order summary

krzysztofzuraw:
use released macaw-ui

PR #5843, opened 2025-09-11
Improve order v2 page

kzuraw:
## Scope of the change

* Adds missing icons from Figma
* Fix issues from https://github.com/saleor/saleor-dashboard/pull/5835


src/orders-v2/icons/order-details-header.tsx
witoszekdev, line 1: question: Is this going to be used only for orders? If not then maybe we can throw that to our icons folder

kzuraw: I will keep it for now in orders not to leak redesign all over the place but in the future for sure we should move it into `icons` folder

PR #5844, opened 2025-09-11
Remove extensions feature flag and dead code

witoszekdev:
This PR removes old `/app`, `/custom-app`, `/plugins` views which were replaced with `/extensions`. This feature was added in 3.21 as a feature flag, which was also deleted in this PR.

This required to make some changes in existing code:
- remove that uses feature flags (e.g. in onboarding)
- removed `view-webhooks` onboarding step which is no longer used
- moved queries, mutations that were previously only in `/apps` to `/extensions` (we couldn't add them to `/extensions` before due to codegen not allowing queries with the same name being duplicated)
- added components that were not migrated to `/extensions` yet from `/apps`
- fixed app routing that rely on legacy `/app/...` routes in app bridge navigation events to use hardcoded URLs (previously they imported methods from `/app` URLs, which are now deleted)

> [!CAUTION]
> Changeset check is failing due to too large diff, we have changeset for this change


witoszekdev:
We'll merge this after 3.22 release. We need to make sure we didn't remove any changes that were made in `/apps`

lkostrowski:
done in other prs

PR #5845, opened 2025-09-12
Narrow down references for product create

IKarbowiak:
## Scope of the change

Implementation for https://github.com/saleor/saleor/pull/18084

- Continuation of https://github.com/saleor/saleor-dashboard/pull/5808
- The PR restricted the reference choices on the product update page


src/searches/useReferenceSearch.tsx
```diff
@@ -0,0 +1,64 @@
...
+type AttributeWithReferenceTypes =
+  | NonNullable<NonNullable<ProductVariantAttributesFragment["attributes"]>[number]["attribute"]>
+  | NonNullable<NonNullable<ProductTypeQuery["productType"]>["productAttributes"]>[number];
+
+export const getAllowedReferenceTypeIds = (
+  refAttr: AttributeWithReferenceTypes | undefined,
+  type: ReferenceType,
+): string[] => {
+  if (refAttr?.referenceTypes?.[0]?.__typename === type) {
+    return (refAttr.referenceTypes ?? []).map((t: any) => t?.id).filter(Boolean);
```
kzuraw: **Question:** why `any`? what will happen if there is no `any` here?

src/searches/useReferenceSearch.tsx
```diff
@@ -0,0 +1,64 @@
...
+  PageType = "PageType",
+}
+
+export enum ReferenceWhereKey {
+  ProductType = "productType",
+  PageType = "pageType",
+}
+
+type AttributeWithReferenceTypes =
+  | NonNullable<NonNullable<ProductVariantAttributesFragment["attributes"]>[number]["attribute"]>
```
kzuraw: I'm wondering if we have fragment for that so we can have `ProductVariantAttributeFragment`?

src/searches/useReferenceSearch.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const buildReferenceSearchVariables = (
+  allowedIds: string[] | undefined,
+  whereKey: ReferenceWhereKey,
+) => ({
+  ...DEFAULT_INITIAL_SEARCH_DATA,
+  ...(allowedIds?.length ? { where: { [whereKey]: { oneOf: allowedIds } } } : {}),
+});
+
+export const useReferenceProductSearch = (refAttr: AttributeWithReferenceTypes | undefined) => {
+  const ids = React.useMemo(
```
kzuraw: what will happen if there is no `useMemo`?

src/searches/useReferenceSearch.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const variables = React.useMemo(
+    () => buildReferenceSearchVariables(ids, ReferenceWhereKey.ProductType),
+    [ids],
+  );
+
+  return useProductSearch({ variables });
+};
+
+export const useReferencePageSearch = (refAttr: AttributeWithReferenceTypes | undefined) => {
+  const ids = React.useMemo(
```
kzuraw: we recently introduced eslint rule to ban `React.` - I suggest you use `useMemo` and import it from `import {useMemo} from React` instead

src/searches/useReferenceSearch.tsx
IKarbowiak: I adjust fragments to use `AttributeDetails` here

src/searches/useReferenceSearch.tsx
IKarbowiak: fixed

src/searches/useReferenceSearch.tsx
IKarbowiak: Done

src/searches/useReferenceSearch.tsx
IKarbowiak: Eslint is crying otherwise

src/fragments/attributes.ts
```diff
@@ -36,29 +36,37 @@ export const attributeFragment = gql`
...
   fragment AttributeDetails on Attribute {
     ...Attribute
-    ...Metadata
     availableInGrid
-    inputType
     entityType
-    unit
     storefrontSearchPosition
     valueRequired
     referenceTypes {
```
witoszekdev, line 50: tip: I think we could add `__typename` like this so that we can detect wether we left an unsupported entity in our code:


```suggestion
    referenceTypes {
    ... on Node {
      __typename
    }
```

This way later in our code we can detect it like this:

```ts
switch (referenceType) {
  case ProductType: ...
  case PageType: ...
  default:
    Sentry.captureException(...);
    throw new Error("Unhandled reference type");
}
```

IKarbowiak: Will update in another PR

PR #5849, opened 2025-09-12
Add prettier to lint-staged, add CI check, fix more files

witoszekdev:
This PR adds auto-formatting using prettier to lint-stgaged. It also adds new CI check for PRs to make sure files are formatted


lint-staged.config.js
```diff
@@ -2,7 +2,8 @@
  * @type {import('lint-staged').Configuration}
  */
 const config = {
-  "./{src,playwright}/**/*.{js,jsx,ts,tsx}": ["eslint --cache --fix"],
+  "./{src,playwright}/**/*.{js,jsx,ts,tsx}": ["prettier --write", "eslint --cache --fix"],
```
kzuraw: Order of those items matters - I think we first should lint and then format:
https://github.com/lint-staged/lint-staged?tab=readme-ov-file#task-concurrency

package.json
```diff
@@ -26,6 +26,7 @@
     "extract-messages": "npm run extract-json-messages",
     "fetch-local-schema": "node scripts/fetch-local-schema.js",
     "fetch-schema": "curl https://raw.githubusercontent.com/saleor/saleor/${npm_package_config_saleor_schemaVersion}/saleor/graphql/schema.graphql > schema.graphql",
+    "format:check": "prettier --check src/*",
```
kzuraw: maybe we should check playwright folder as well?

lint-staged.config.js
lkostrowski: yup, otherwise eslint will overwrite formatting

package.json
witoszekdev: done: https://github.com/saleor/saleor-dashboard/pull/5850

PR #5851, opened 2025-09-12
[Orders v2] Order refunds

kzuraw:
## Scope of the change

2025-09-15 at 14 12 30 - CleanShot@2x
https://saleor-dashboard-storybook-git-eng-710-or-ccf746-saleorcommerce.vercel.app/?path=/story/order-detail-page-content-refunds--success

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -160,12 +160,15 @@ export default tseslint.config(
       "no-restricted-imports": [
         "error",
         {
-          paths: ["lodash", "classnames",
+          paths: [
+            "lodash",
+            "classnames",
             {
               name: "react",
               importNames: ["default", "React"],
-              message: "Import directly the needed functions, e.g. 'import {useState} from \"react\"'",
-            }
+              message:
+                "Import directly the needed functions, e.g. 'import {useState} from \"react\"'",
+            },
           ],
         },
       ],
@@ -233,7 +236,7 @@ export default tseslint.config(
             {
               name: "moment-timezone",
               message: "Use react-intl formatDate instead of moment-timezone.",
-            }
+            },
           ],
         },
       ],


src/orders/components/OrderDetailsRefundTable/refunds.ts
kzuraw, line 1: Moved into `src/orders-v2/order-refunds/order-refunds-view-model.ts`

src/orders/components/OrderDetailsRefundTable/utils.ts
kzuraw, line 1: Moved into `src/orders-v2/order-refunds/order-refunds-view-model.ts`

src/orders-v2/order-refunds/order-refunds-view-model.ts
```diff
@@ -0,0 +1,176 @@
...
+    return "refundable";
+  }
+
+  static canEditRefund(refund: OrderRefundDisplay): boolean {
+    const isSuccessful = refund.status === OrderGrantedRefundStatusEnum.SUCCESS;
+    const isPending = refund.status === OrderGrantedRefundStatusEnum.PENDING;
+    const isManual = refund.type === "manual";
+
+    // Can only edit refunds that are not successful, not pending, and not manual
+    return !isSuccessful && !isPending && !isManual;
```
lkostrowski, line 55: image


;)

src/orders-v2/order-refunds/order-refunds-view-model.ts
```diff
@@ -0,0 +1,176 @@
...
+      if (!acc[event.pspReference]) {
+        acc[event.pspReference] = [];
+      }
+
+      acc[event.pspReference].push(event);
+
+      return acc;
+    }, {});
+  };
+
```
lkostrowski, line 71: I know you copy-pasted it, but there is a bug (you can check my PR with refund reasons)

1. Create refund that fails
2. Create another refund that fails

Both of them have the same PSP (from transaction id? not sure). Eventually they are all merged and you will see only the last event.

I think we should see all of them

(non blocking, we can park it for later, details are visible in events list anyway)

src/orders-v2/order-refunds/order-refunds-view-model.ts
```diff
@@ -0,0 +1,176 @@
...
+  ): OrderRefundDisplay["user"] => {
+    if (creator?.__typename === "User") {
+      return {
+        email: creator.email,
+        firstName: creator.firstName,
+        lastName: creator.lastName,
+      };
+    }
+
+    return null;
```
lkostrowski, line 104: we don't render app?

lkostrowski: not sure when its null. Its automatic, created by sAelor? if so, maybe we should show saleor logo or something similar

src/orders-v2/order-refunds/refund-list-item.tsx
```diff
@@ -0,0 +1,85 @@
...
+      alignItems="center"
+      gap={7}
+      as="li"
+      {...props}
+    >
+      <Box display="flex" alignItems="center" gap={3}>
+        {refund.user ? (
+          <UserAvatar initials={getUserInitials(refund.user) || ""} />
+        ) : (
+          <Box width={8} height={8} />
```
lkostrowski, line 32: refund can be initiated by app so I think we should render it if possible

src/orders-v2/order-refunds/status-badge.tsx
```diff
@@ -0,0 +1,35 @@
...
+    case OrderGrantedRefundStatusEnum.PENDING:
+      return {
+        status: "warning",
+        label: "Pending",
+      } as const;
+    case OrderGrantedRefundStatusEnum.NONE:
+    default:
+      return {
+        status: "warning",
+        label: "Draft",
```
lkostrowski, line 26: missing translations?

src/orders-v2/user-avatar.tsx
lkostrowski, line 1: lets scan other avatar components and deprecate them

src/orders-v2/order-refunds/order-refunds-view-model.ts
kzuraw: Sure - I will add todo comment with issue to check that before going live

locale/defaultMessages.json
witoszekdev, line 1: suggestion: let's add context for these messages, so that it's easier to translate them

src/orders-v2/order-refunds/order-refunds-view-model.ts
kzuraw: yes - I copy / pasted it from previous implementation - I will change that in the next PR

src/orders-v2/fixtures/order-fixture.ts
```diff
@@ -480,6 +482,57 @@ export class OrderFixture {
...
+    refundPendingAmount: {
+      __typename: "Money" as const,
+      amount: 0,
+      currency: "USD",
+    },
+    cancelPendingAmount: {
+      __typename: "Money" as const,
+      amount: 0,
+      currency: "USD",
+    },
```
witoszekdev, line 533: nitpick: we could add utility function to create these money object:

```typescript
const createMockMoney = (amount: number, currency = "USD"): Money => ...
```

src/orders-v2/order-refunds/refund-list-item.tsx
kzuraw: As above - I will fix it in next PR

src/orders-v2/order-refunds/order-refunds-view-model.ts
```diff
@@ -0,0 +1,176 @@
...
+  static getRefundState(transactions: OrderDetailsFragment["transactions"]): OrderRefundState {
+    if (transactions.length === 0) {
+      return "noTransactionsToRefund";
+    }
+
+    if (
+      transactions.every(transaction => !transaction.actions.includes(TransactionActionEnum.REFUND))
+    ) {
+      return "allTransactionsNonRefundable";
+    }
```
witoszekdev, line 44: question: What about manual refunds? Previously we allowed creating new transactions with `refundedAmount` to mark cases just like that: when there are no transactions that can be refunded and store has refunded customer in some other way (e.g. wire transfer)

witoszekdev (commented):
suggestion: we could use grid to render list of refunds, so that we don't have slight misalignment when status string is shorter / longer:

CleanShot 2025-09-15 at 15 43 42@2x

src/orders-v2/order-refunds/order-refunds-view-model.ts
kzuraw: logic is the same as it was before but instead of `true/false` I'm using state that combines boolean with reason. Previous logic:

```ts
export const canAddRefund = ({
  transactions,
  intl,
}: {
  transactions: OrderDetailsFragment["transactions"];
  intl: IntlShape;
}): CanAddRefund => {
  if (transactions.length === 0) {
    return {
      canRefund: false,
      reason: intl.formatMessage(refundGridMessages.noTransactionsToRefund),
    };
  }

  if (
    transactions.every(transaction => !transaction.actions.includes(TransactionActionEnum.REFUND))
  ) {
    return {
      canRefund: false,
      reason: intl.formatMessage(refundGridMessages.allTransactionsNonRefundable),
    };
  }

  return {
    canRefund: true,
  };
```

PR #5852, opened 2025-09-12
Inject @saleor/sdk package into Dashboard source

lkostrowski:
We want:
1. Archive `@saleor/sdk` package (use auth-sdk)
2. Remove dependency to `@saleor/sdk`, e.g. to upgrade dependnent packages, without updating the lib

First step is copy-pasting this library into the Dashboard itself. It now behaves the same, but without extenral dependency. Now if we want to e.g. update Apollo, we can do this easily within Dashboard. We can migrate to auth-sdk as well if needed


codegen.ts
```diff
@@ -1,8 +1,9 @@
 import { CodegenConfig } from "@graphql-codegen/cli";
 
 const config: CodegenConfig = {
-  schema: "./schema.graphql",
+  schema: ["./schema.graphql", "./src/legacy-sdk/local-schema.graphql"],
```
witoszekdev, line 4: question: why do we need additional schema for legacy-sdk?

witoszekdev (approved):

codegen.ts
lkostrowski: Because it adds apollo-specific, local directives (@client-decorated). They are merged into the root schema.
We can drop this but it will require us to change implementation of this lib. Which is fine, but the scope of this change is to pull sdk and make it run only

Qasem-h:
Hi @krzysztofzuraw 
Do you have an estimated timeline for when this will be merged into the dashboard main ?

lkostrowski:
@Qasem-h I will finish this, but I can't put any date

Why is this important for you? We do this, because we want to remove this npm package (it's a cleanup)

lkostrowski:
Merged as the other PR and sdk was archived

PR #5856, opened 2025-09-15
Fix assigning category and collection reference on product variant views

IKarbowiak:
## Scope of the change
Previously, assigning category and collection reference attributes in the ProductVariant create and update views did not display the assigned values.

After the changes:
image


IKarbowiak:
Skipping changeset, as a category and collection reference are not released yet.

src/products/views/ProductVariantCreate.tsx
```diff
@@ -48,7 +48,7 @@ interface ProductVariantCreateProps {
   params: ProductVariantAddUrlQueryParams;
 }
 
-export const ProductVariant = ({ productId, params }: ProductVariantCreateProps) => {
+const ProductVariant = ({ productId, params }: ProductVariantCreateProps) => {
```
witoszekdev: let's keep using named export instead of default:

```suggestion
export const ProductVariant = ({ productId, params }: ProductVariantCreateProps) => {
```

eslint.config.mjs
witoszekdev, line 1: please let's revert these changes, i have separate PR that fixes ESLint config: https://github.com/saleor/saleor-dashboard/pull/5858

PR #5858, opened 2025-09-15
Fix ESLint import rule

witoszekdev:
This PR fixes ESLint import rule which didn't work for declared `error`'s - this is because ESLint merging rules doesn't allow to declare the same rule that will be both `warning` and `error`.

This means for imports we need to keep using `warning` for now, and migrate to `error` next.


eslint.config.mjs
```diff
@@ -233,7 +220,23 @@ export default tseslint.config(
             {
               name: "moment-timezone",
               message: "Use react-intl formatDate instead of moment-timezone.",
-            }
+            },
+
+            // Note: these should be errors but we cannot use "warn" and "error" rules in ESLint together
```
kzuraw, line 225:
```suggestion
            // Note: these should be errors but we cannot use "warn" and "error" rules in ESLint together for the same rule
```

krzysztofzuraw (approved):

PR #5863, opened 2025-09-16
[3.21] Update fulfillment section with tracking number & update checkbox copy

kzuraw:


package-lock.json
kzuraw, line 1: Autoupdated when I installed deps - only change is dashboard version number

src/graphql/types.generated.ts
kzuraw, line 1: Autoupdated when I ran dev server

lkostrowski: have you checked diff? looks like we updated schema without many "deprecation" comments 🤔 strange

lkostrowski (dismissed):
Code is fine, I'd just check if this is expected schema we fetched (looks like we removed deprecation fields, why?)

PR #5859, opened 2025-09-15
Fix knip to delete unused files

witoszekdev:
This PR fixes knip to remove unused files and unused exports.

Previous config had deleted code that was actually used (e.g. in tests), now it should be handled properly

It also fixes `scripts/` directory so that it can be properly detected by knip and ESLint, it adds dependencies from scripts which were previously not declared in package.json

PR #5866, opened 2025-09-16
Knip: remove unused code, remove unused exports

witoszekdev:
Continuation of #5859

This PR deletes dead-code that wasn't used anywhere and removes exports if they weren't used outside of local file.

> [!CAUTION]
> **Do not open "Files changed" tab**, it will freeze your browser 🙈 
> Open changes using `github.dev`: https://github.dev/saleor/saleor-dashboard/pull/5866/files

## Other changes

**These changes were needed to make this cleanup possible**

### ESLint

Added new ESLint rules that were used during this cleanup and will be useful afterwards as well:

- `no-unused-vars` - similar to our existing TypeScript config, but will ignore `catch (error)`, and vars starting with `_`

- `eslint-unused-imports-plugin` - to auto-delete unused imports (useful in normal development to get rid of annoying errors from TypeScript)

- [eslint-plugin-unicorn](https://github.com/sindresorhus/eslint-plugin-unicorn) with rule `unicorn/no-empty-file` - to detect and remove empty files after our cleanup (due to running all rules, file content was completely removed) - we might want to use more rules from this plugin 👀 

### TypeScript

Removed checking unused vars and imports, since this is now done by ESLint. This might improve TypeScript performance.

### Other

Added `OrderGiftCardEventBalanceFragment` to fix TypeScript error in `src/orders/components/OrderSummaryCard/utils.ts`

## Steps taken

For future prosperity here were the steps taken to cleanup unused files:

1. `npm run knip:fix`
2. `npm run linter` 
3. `npx eslint --quiet -f json | npx remove-unused-vars`
4. Repeat steps 1-3 a few times
5. `npm run check-types`
8. Fix manually some errors

### Manual fixes

- Playwright tests: they use `let someVar` and then initialize it in `beforeEach` block. If variable wasn't actually used in test, knip will delete it. But it won't delete the `beforeEach` initailization - this needs to be deleted manually
- Empty files - `unicorn/no-empty-file` doesn't support auto-fix, do some bash magic by simply: running eslint, exporting result to json, parse it using jd, pass to `rm` command ಠ‿↼
- Enums: for some reason unused enums are not deleted by `remove-unused-vars` - these need to be removed manually
- Barrel files: this evil monstrosity breaks knip and/or eslint when used along with another enemy: default exports. In order to resolve issue we must remove default exports and then `export {default as something} from "./some-file"` line in barrel file. Ideally barrel files should be gone, but we already have 900+ files changed 😄


eslint.config.mjs
```diff
@@ -24,6 +26,7 @@ export default tseslint.config(
     "type-policies.ts",
     "playwright/auth.js",
     "**/*.generated.ts",
+    ".github/**/*.js",
```
witoszekdev, line 29: We had some scripts for github workflows, this probably can't be detected properly by knip or eslint so I just added it to ignore

eslint.config.mjs
```diff
@@ -64,7 +67,16 @@ export default tseslint.config(
       "@typescript-eslint/no-non-null-asserted-optional-chain": "warn",
       "@typescript-eslint/ban-types": "off",
       "@typescript-eslint/no-explicit-any": "off",
-      "@typescript-eslint/no-unused-vars": "off",
+      "@typescript-eslint/no-unused-vars": [
+        "error",
```
witoszekdev, line 71: This config is a bit more forgiving than previous TypeScript config, because it:
- allows to leave `catch (error)` as is
- checks only local variables, and not globals (e.g. on `window`)
- checks unused arguments in functions, methods after last used (e.g. `unusedArg1, arg2, arg3, unusedArg4` where last used was `arg3`, only `unusedArg4` will raise error)
- ignores unused variables and arguments that start with `_` which is a common pattern

package-lock.json
```diff
@@ -135,6 +135,8 @@
         "eslint-plugin-react-hooks": "^5.2.0",
         "eslint-plugin-react-refresh": "^0.4.20",
         "eslint-plugin-simple-import-sort": "^12.1.1",
+        "eslint-plugin-unicorn": "^61.0.2",
```
witoszekdev, line 138: FYI: most added lines come from package-lock

PR #5868, opened 2025-09-17
Fix reference types dialog when changing entity type

IKarbowiak:
## Scope of the change
- The reference types weren't cleared properly. After change it's reset properly when entity type change.
- Update graphql schema

Switching from `ProductVariant` to `Product` entity type
image

image
image


src/attributes/components/AssignReferenceTypesDialog/AssignReferenceTypesDialog.tsx
```diff
@@ -11,24 +12,33 @@ interface AssignCollectionDialogProps
...
+  const title =
+    entityType === AttributeEntityTypeEnum.PAGE
+      ? intl.formatMessage(messages.titleModelTypes)
+      : intl.formatMessage(messages.titleProductTypes);
+
   return (
     <AssignContainerDialog
+      // Force react re-render to clear selected reference types when changing entity type
+      // Should be removed after moving the selectedContainers outside AssignContainerDialog
+      key={entityType}
```
IKarbowiak, line 41: Hack for now as I don't have time to refactor `AssignContainerDialog`

witoszekdev (approved):

PR #5869, opened 2025-09-17
Use Knip in warn mode

krzysztofzuraw:

### DIFF
--- a/.knip.json
+++ b/.knip.json
@@ -1,20 +1,20 @@
 {
   "$schema": "https://unpkg.com/knip@5/schema.json",
   "rules": {
-    "types": "error",
+    "types": "warn",
     "enumMembers": "off",
-    "files": "error",
-    "exports": "error",
-    "duplicates": "error",
-    "classMembers": "error",
-    "nsTypes": "error",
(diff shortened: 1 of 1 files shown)


.knip.json
```diff
@@ -1,20 +1,20 @@
...
+    "unresolved": "warn",
+    "nsExports": "warn",
+    "unlisted": "warn",
     "binaries": "off",
-    "optionalPeerDependencies": "error",
-    "dependencies": "error",
-    "devDependencies": "error"
+    "optionalPeerDependencies": "warn",
+    "dependencies": "warn",
+    "devDependencies": "warn"
```
witoszekdev: suggestion: This can be actually left as error, since we had it like this before my changes 😄 - this blocked PR if you didn't add new dependency to package.json by accident, which imho makes sense as it can break build if it wasn't installed properly (e.g. some package uses as it's own dependency)

```suggestion
    "optionalPeerDependencies": "error",
    "dependencies": "error",
    "devDependencies": "error"
```

kzuraw: Fixed

witoszekdev (approved):

PR #5885, opened 2025-09-19
Cleanup `package.json` scripts

witoszekdev:
This PR cleans up package.json scripts. It removes old unused cypress scripts (it's no even installed anymore).
It cleans up existing scripts to have simpler names.

ESLint and Prettier were merged into a single script so that we always run them both in correct order for proper formatting


lkostrowski (approved):

lkostrowski:
@witoszekdev 

knip has complained about unused cypress-dependencies so instead of silencing him, I just removed these. If cypress is not used/installed, looks like we don't need its reporters (or mocha in general)

see 
[`6da4ef7` (#5885)](https://github.com/saleor/saleor-dashboard/pull/5885/commits/6da4ef79385a2c5b9fcf4f7d1c8bc5c9ebcbf1f3)

@krzysztofzuraw we need your review here

package.json
```diff
@@ -22,30 +22,21 @@
     "dep-status": "depcruise --config .dependency-cruiser.js src",
     "predev": "npm run generate",
     "dev": "vite --host",
-    "extract-json-messages": "formatjs extract 'src/**/*.{ts,tsx}' --ignore '**/*.d.ts' --out-file locale/defaultMessages.json --format scripts/formatter.cjs",
-    "extract-messages": "npm run extract-json-messages",
+    "e2e": "npx playwright test --grep \"#e2e\"",
+    "e2e:ui": "npx playwright test --ui",
+    "eslint": "cross-env NODE_OPTIONS=--max-old-space-size=8192 eslint \"{src,playwright}/**/*.@(tsx|ts|jsx|js)\" --fix",
+    "eslint:check-progress": "eslint-nibble \"src/**/*.@(tsx|ts|jsx|js)\"",
```
kzuraw: **Question:** do we need that?

package.json
```diff
@@ -22,30 +22,21 @@
...
+    "format": "prettier --ignore-unknown --write .",
     "format:check": "prettier --check .",
-    "format:code": "prettier --ignore-unknown --write .",
     "generate": "graphql-codegen",
-    "heroku-postbuild": "npm run build",
     "knip": "knip --reporter markdown",
     "knip:fix": "knip --fix --allow-remove-files",
-    "lint": "cross-env NODE_OPTIONS=--max-old-space-size=8192 eslint \"{src,playwright}/**/*.@(tsx|ts|jsx|js)\" --fix",
-    "lint:check-progress": "eslint-nibble \"src/**/*.@(tsx|ts|jsx|js)\"",
+    "lint": "npm run eslint && npm run format",
```
kzuraw: I suggest we add `lint:eslint` / `lint:format` so either we can have `npm-run-all` (for npm) or use pnpm filtering in the future

package.json
witoszekdev: I've tried to check if it even works, and eslint-nibble doesn't seem to be working with flatconfig

lkostrowski (approved):

PR #5888, opened 2025-09-22
Introduce "ripples" to dashboard

lkostrowski:
Dashboard now shows  "ripples" with new stuff we add

CleanShot 2025-09-22 at 13 41 24

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,3 +3,5 @@
 <!-- Describe changed made in this PR. You can attach screenshots or mention related issues as well. -->
 
 <!-- External contributors: Please attach GitHub issue number. -->
+
+- [] I confirm I added change ripples (see src/ripples) or my feature doesn't contain any
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     }
   },
   "dependencies": {
+    "jotai": "2.14.0",
     "@apollo/client": "3.4.17",
     "@dnd-kit/core": "^6.3.1",
     "@dnd-kit/modifiers": "^9.0.0",
--- a/src/modeling/ripples/pages-are-models.tsx
+++ b/src/modeling/ripples/pages-are-models.tsx
@@ -5,6 +5,7 @@ export const ripplePagesAreModels: Ripple = {
   ID: "pages-are-models",
   TTL: 60 * 60 * 24 * 2, // 2 days
   content: {
+    oneLiner: "Pages are now called Models",
     contextual: "Pages are now called Models",
     global:
       "We have renamed Pages to Models. API still uses the old naming, but we it will change in the future.",
--- a/src/modeling/ripples/pages-are-models.tsx
+++ b/src/modeling/ripples/pages-are-models.tsx
@@ -15,13 +15,12 @@ export const ripplePagesAreModels: Ripple = {
   actions: [
     {
       label: defineMessage({
-        defaultMessage: "Test",
-        id: "xu6eM8",
+        defaultMessage: "Read a blog post",
+        id: "7DMWSy",
       }),
-      onClick() {},
+      onClick() {
+        window.open("https://saleor.io/blog/modeling", "_blank");
+      },
     },
   ],
 };
-
-// todo
-// https://saleor.io/blog/modeling
--- a/src/modeling/ripples/pages-are-models.tsx
+++ b/src/modeling/ripples/pages-are-models.tsx
@@ -22,3 +22,6 @@ export const ripplePagesAreModels: Ripple = {
     },
   ],
 };
+
+// todo
+// https://saleor.io/blog/modeling
--- a/src/modeling/ripples/pages-are-models.tsx
+++ b/src/modeling/ripples/pages-are-models.tsx
@@ -1,0 +1,23 @@
+import { Ripple } from "@dashboard/ripples/types";
+import { defineMessage } from "react-intl";
+
+export const ripplePagesAreModels: Ripple = {
+  ID: "pages-are-models",
+  TTL: 60 * 60 * 24 * 2, // 2 days
+  content: {
+    contextual: "Pages are now called Models",
+    global:
+      "We have renamed Pages to Models. API still uses the old naming, but we it will change in the future.",
+  },
+  // Approx June it was changed in the dashboard
+  dateAdded: new Date(2025, 5, 1),
+  actions: [
+    {
+      label: defineMessage({
+        defaultMessage: "Test",
+        id: "xu6eM8",
+      }),
+      onClick() {},
+    },
+  ],
+};
--- a/src/ripples/components/Ripple.tsx
+++ b/src/ripples/components/Ripple.tsx
@@ -2,7 +2,7 @@ import { RippleAnimation } from "@dashboard/ripples/components/RippleAnimation";
 import { useRippleStorage } from "@dashboard/ripples/hooks/useRipplesStorage";
 import type { Ripple as RippleModel } from "@dashboard/ripples/types";
 import { TooltipMountWrapper } from "@saleor/macaw-ui";
-import { Box, Button, Text, Tooltip } from "@saleor/macaw-ui-next";
+import { Box, BoxProps, Button, Text, Tooltip } from "@saleor/macaw-ui-next";
 import { useIntl } from "react-intl";
 
 /**
@@ -14,7 +14,7 @@ import { useIntl } from "react-intl";
  * TODO
  * Make tooltip working on click instead of hover
  */
-export const Ripple = (props: { model: RippleModel }) => {
+export const Ripple = (props: { model: RippleModel; triggerProps?: BoxProps }) => {
   const content = props.model.content.contextual;
   const isPlainString = typeof content === "string";
   const intl = useIntl();
@@ -32,7 +32,7 @@ export const Ripple = (props: { model: RippleModel }) => {
     >
       <Tooltip.Trigger>
         <TooltipMountWrapper>
-          <RippleAnimation cursor="pointer" />
+          <RippleAnimation cursor="pointer" {...props.triggerProps} />
         </TooltipMountWrapper>
       </Tooltip.Trigger>
       <Tooltip.Content align="start" side="bottom">
@@ -48,6 +48,8 @@ export const Ripple = (props: { model: RippleModel }) => {
                 onClick={e => {
                   e.preventDefault();
                   e.stopPropagation();
+
+                  setManuallyHidden();
 
                   act.onClick();
                 }}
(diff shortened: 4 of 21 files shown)


package.json
```diff
@@ -61,6 +61,7 @@
     }
   },
   "dependencies": {
+    "jotai": "2.14.0",
```
lkostrowski, line 64: introduced jotai finally to seamless state management. its a drop-in replacement for useState often. for example, useLocalStorage is not global state but only local. jotai binding makes it global when needed

src/modeling/ripples/pagesAreModels.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export const ripplePagesAreModels: Ripple = {
+  ID: "pages-are-models",
+  TTL: 60 * 60 * 24 * 2, // 2 days
+  content: {
+    oneLiner: "Pages are now called Models",
+    contextual: "Pages are now called Models",
+    global: (
+      <>
+        <Paragraph>
```
kzuraw, line 14: nitpick to dicuss - do we need Paragraph in macaw-ui when we have a Text?

src/modeling/ripples/pages-are-models.tsx
```diff
@@ -0,0 +1,40 @@
+import Link from "@dashboard/components/Link";
+import { Ripple } from "@dashboard/ripples/types";
+import { Paragraph } from "@saleor/macaw-ui-next";
+import { defineMessage } from "react-intl";
+
+export const ripplePagesAreModels: Ripple = {
+  ID: "pages-are-models",
+  TTL: 60 * 60 * 24 * 2, // 2 days
```
kzuraw: what about changing `TTL` to `TTL_MS`? thanks to that I will know what unit we here expect

src/modeling/ripples/pagesAreModels.tsx
```diff
@@ -0,0 +1,40 @@
...
+          We have renamed Pages to Models. API still uses the old naming, but it will change in the
+          future.
```
kzuraw, line 16: do we want to translate this?

src/ripples/hooks/useRipplesStorage.ts
kzuraw, line 1: should we have common pattern for naming files? sometimes we have `camelCase` and sometimes we have `kebab-case`

src/ripples/hooks/useRipplesStorage.ts
```diff
@@ -0,0 +1,98 @@
+import { allRipples } from "@dashboard/ripples/all-ripples";
+import { Ripple } from "@dashboard/ripples/types";
+import { useAtom } from "jotai";
+import { atomWithStorage } from "jotai/vanilla/utils";
+import lodashSet from "lodash/set";
+
+type StoredRipple = {
+  manuallyHidden: boolean;
+  firstSeenAt: number; // seconds
```
kzuraw, line 9:
```suggestion
  firstSeenAtSeconds: number; // seconds
```

src/ripples/hooks/useRipplesStorage.ts
witoszekdev: components use pascal case: `MyComponent.tsx`
hooks use camel case: `useHook` and pretty much anything else (`utils.ts`)

afaik we don't use kebab-case anywhere

.changeset/hip-women-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Added "ripples" indicating new features added to the dashboard, directly for staff users
```
witoszekdev, line 5: tip: we could add images here, they will be automatically visible in changelog

locale/defaultMessages.json
```diff
@@ -924,6 +924,9 @@
     "context": "ProductTypeDeleteWarningDialog with items multiple description",
     "string": "You are about to delete multiple product types. Some of them are assigned to products. Deleting those product types will also delete those products"
   },
+  "3eR8iC": {
+    "string": "Show me"
+  },
```
witoszekdev, line 929: suggestion: let's add context for these messages (e.g. `Ripple tooltip for new features`)

locale/defaultMessages.json
```diff
@@ -6195,6 +6201,10 @@
   "YZl6cv": {
     "string": "Miscellaneous"
   },
+  "YcJdUV": {
+    "context": "recent changes shortcut",
+    "string": "What’s New"
```
witoszekdev, line 6206: suggestion: I think we should stick to the `'` apostrophe, instead of `’` since it's used everywhere else

```suggestion
    "string": "What's New"
```

src/orders/ripples/newRefundReasons.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const rippleNewRefundReasons: Ripple = {
+  ID: "new-refund-reasons",
+  TTL_seconds: 60 * 60 * 24 * 2, // 2 days
+  content: {
+    oneLiner: "Structured refund reasons",
+    contextual: "Refund reasons are now more powerful with pre-defined choices",
+    global: (
+      <>
+        <Paragraph>
+          Refund reasons are now more powerful. You can add reason to the &#34;manual refund&#34;.
```
witoszekdev, line 15: issue: Let's just use normal quotationmarks instead of HTML codes:


```suggestion
          Refund reasons are now more powerful. You can add reason to the "manual refund".
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,3 +3,5 @@
 <!-- Describe changed made in this PR. You can attach screenshots or mention related issues as well. -->
 
 <!-- External contributors: Please attach GitHub issue number. -->
+
+- [ ] I confirm I added ripples for changes (see src/ripples) or my feature doesn't contain any user-facing changes
```
witoszekdev, line 7: suggestion: We should also add entry to add tracking events like in this PR

src/ripples/ripples/introducedRipples.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Ripple } from "@dashboard/ripples/types";
+import { Paragraph } from "@saleor/macaw-ui-next";
+import { getDefaultStore } from "jotai";
+import { defineMessage } from "react-intl";
+
+const defaultStore = getDefaultStore();
+
+export const rippleIntroducedRipples: Ripple = {
+  TTL_seconds: 60 * 60 * 24 * 2,
+  ID: "introducing-ripples",
```
witoszekdev, line 11: suggestion: I think for IDs we could use a single file for all of them and create a `new Symbol` for each ID, this way we will know it's not duplicated and there won't be a need to throw error in runtime in development

src/ripples/hooks/useRipplesStorage.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  constructor(
+    storedState: StoredRipplesRecord,
+    updateState: (newState: StoredRipplesRecord) => void,
+    allRipples: Ripple[],
+  ) {
+    this.storedState = storedState;
+    this.updateState = updateState;
+    this.allRipples = allRipples;
+  }
```
witoszekdev, line 33: suggestion: we cam simplify this syntax

```suggestion
  constructor(
    private storedState: StoredRipplesRecord,
    private updateState: (newState: StoredRipplesRecord) => void,
    private allRipples: Ripple[],
  ) {
  }
```

src/ripples/hooks/useRipplesStorage.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  setFirstSeenFlag(ripple: Ripple): void {
+    // Do not override, we only store the first event
+    if (this.storedState[ripple.ID]?.firstSeenAt) {
+      return;
+    }
+
+    const newState = structuredClone(this.storedState);
+
+    lodashSet(newState, `${ripple.ID}.firstSeenAt`, this.now);
```
witoszekdev, line 65: suggestion: we could replace lodash with this:

```ts
  newState[ripple.ID] = newState[ripple.ID] || {};
  newState[ripple.ID].firstSeenAt = this.now;
```

src/ripples/components/RippleAnimation.tsx
```diff
@@ -0,0 +1,77 @@
...
+          0% {
+            transform: scale(0);
+            opacity: 1;
+          }
+          100% {
+            transform: scale(4);
+            opacity: 0;
+          }
+        }
+      `}</style>
```
witoszekdev, line 38: issue: for custom CSS we could use CSS Modules, it's already used in dashboard

src/ripples/components/RippleAnimation.tsx
```diff
@@ -0,0 +1,77 @@
+"use client";
```
witoszekdev, line 1: i think we can get rid of this, since we don't use React Server Components in Dashboard

```suggestion
```

src/ripples/hooks/useRipplesStorage.ts
lkostrowski: yea I reverted, we should probably add eslint rule

src/ripples/components/Ripple.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <Box>
+          <RippleAnimation cursor="pointer" />
+        </Box>
+      </Tooltip.Trigger>
+      <Tooltip.Content align="start" side="bottom">
+        <Tooltip.Arrow />
+        <Tooltip.ContentHeading>Hint</Tooltip.ContentHeading>
+        <Box marginBottom={4}>{isPlainString ? <Text>{content}</Text> : content}</Box>
+        <Box display="flex" justifyContent="flex-end" gap={2}>
+          {props.model.actions?.map((act, index) => {
```
witoszekdev, line 33: suggestion: Could we rename this to what we're iterating on? `act` looks weird like it's a test 😅 

```suggestion
          {props.model.actions?.map((model, index) => {
```

src/orders/ripples/newRefundReasons.tsx
lkostrowski: we have rule "(react/no-unescaped-entities" enabled, if we want to drop it, lets drop it separately and refactor codebase

.changeset/hip-women-scream.md
lkostrowski: Its tricky on PR level, you need to host image somewhere, IMO better to do this when we prepare release

src/ripples/ripples/introducedRipples.tsx
lkostrowski: Current solution should work too. For now ripples are per-module. If we want IDs in single file, we should discuss if all ripples should be in the same file (maybe it's good, idk).

Symbols dont solve the problem because IDs are needed in local storage, you can't serialize symbol

PR #5891, opened 2025-09-23
Store last login method

kzuraw:
## Scope of the change
Backport of https://github.com/saleor/saleor-dashboard/pull/5880

### DIFF
--- a/src/auth/components/LoginPage/LoginPage.tsx
+++ b/src/auth/components/LoginPage/LoginPage.tsx
@@ -1,3 +1,4 @@
+import { LastLoginMethod } from "@dashboard/auth/hooks/useLastLoginMethod";
 import { UserContextError } from "@dashboard/auth/types";
 import { passwordResetUrl } from "@dashboard/auth/urls";
 import { ButtonWithLoader } from "@dashboard/components/ButtonWithLoader/ButtonWithLoader";
@@ -13,6 +14,7 @@ import { Link } from "react-router-dom";
 
 import useStyles from "../styles";
 import LoginForm, { LoginFormData } from "./form";
+import { LastLoginIndicator } from "./LastLoginIndicator";
 import { getErrorMessage } from "./messages";
 
 interface LoginCardProps {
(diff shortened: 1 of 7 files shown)


src/auth/hooks/useLastLoginMethod.ts
```diff
@@ -0,0 +1,12 @@
+import useLocalStorage from "@dashboard/hooks/useLocalStorage";
+
+export type LastLoginMethod = string | null;
+
+export const useLastLoginMethod = () => {
+  const [lastLoginMethod, setLastLoginMethod] = useLocalStorage<LastLoginMethod>(
+    "lastLoginMethod",
```
witoszekdev, line 7: suggestion: let's have this localStorage key as a separate variable

src/auth/hooks/useAuthProvider.ts
```diff
@@ -218,6 +220,8 @@ export function useAuthProvider({ intl, notify, apolloClient }: UseAuthProviderO
 
       await logoutNonStaffUser(result.data?.externalObtainAccessTokens!);
 
+      setLastLoginMethod(pluginId);
```
witoszekdev, line 223: question: shouldn't this use a prefix to indicate it's a plugin signin? e.g. `plugin:google`

src/auth/components/LoginPage/LoginPage.tsx
```diff
@@ -152,8 +158,10 @@ const LoginPage = (props: LoginCardProps) => {
                 transitionState={
                   optimisticLoaderAuthId === externalAuthentication.id ? "loading" : "default"
                 }
+                position="relative"
               >
                 {externalAuthentication.name}
+                {lastLoginMethod === externalAuthentication.id && <LastLoginIndicator />}
```
witoszekdev, line 164: question: shouldn't we display a single LastLoginIndicator component? 🤔

src/auth/hooks/useLastLoginMethod.ts
kzuraw: I will add it in different PR as this is backport

src/auth/hooks/useAuthProvider.ts
kzuraw: we don't have such use-case now - we rather wants to know if user is logged via password

src/auth/components/LoginPage/LoginPage.tsx
kzuraw: No - we have button per external auth method, hence last login indicator is checked for each of button

PR #5894, opened 2025-09-24
Improve Account Settings

krzysztofzuraw:
## Scope of the change
Backport of https://github.com/saleor/saleor-dashboard/pull/5892


src/auth/utils.ts
```diff
@@ -98,3 +101,6 @@ export async function handleQueryAuthError(
     showAllErrors({ notify, error });
   }
 }
+
+export const getNewPasswordResetRedirectUrl = () =>
+  urlJoin(window.location.origin, getAppMountUriForRedirect(), newPasswordUrl().replace(/\?/, ""));
```
lkostrowski, line 106: we could use some test here

lkostrowski (approved):

src/auth/utils.ts
kzuraw: Will do

PR #5908, opened 2025-09-30
Remove some usages of extensions

lkostrowski:
## Scope of the change

This PR doesn't remove all dead code, but it:
1. Tracks the feature flag to remove
2. Remove dead code which was branched from that flag

As a follow up, I will remove `apps/` directory incrementally to remove the dead code after code was duplicated. 


src/configuration/index.tsx
```diff
@@ -37,10 +32,7 @@ import { ConfigurationPage } from "./ConfigurationPage";
 import { MenuSection } from "./types";
 
 // TODO: Remove hideOldExtensions once "extensions" feature flag is removed
```
kzuraw: Seems like this todo can be removed as well?

PR #5910, opened 2025-09-30
wip

lkostrowski:
## Scope of the change


.featureFlags/new_order_ui.md
lkostrowski, line 1: @witoszekdev @krzysztofzuraw I had to add this flag ahead of time, because current implementation fails if there are no flags (most likely typescript only). I decided to do this instead of fixing it, because likely we will have flags all the time (and new order is coming as well)

PR #5934, opened 2025-10-02
Fix error on Order details page

witoszekdev:
## Scope of the change

This PR fixes error when opening order draft details. It also adds strict mode in a file


krzysztofzuraw (approved):

src/orders/utils/data.ts
```diff
@@ -62,17 +61,17 @@ export function getWarehousesFromOrderLines<T extends OrderLineWithStockWarehous
...
-  return (
-    getOrderCharged(order as OrderDetailsFragment) &&
-    order?.total?.gross &&
-    subtractMoney(getOrderCharged(order as OrderDetailsFragment), order?.total?.gross)
-  );
+  const orderDetails = order as OrderDetailsFragment;
+  const charged = getOrderCharged(orderDetails);
+  const total = orderDetails?.total?.gross;
+
+  return charged && total ? subtractMoney(charged, total) : charged;
```
lkostrowski, line 74: can you please add test? this is where we should be aligned - we don't need to test html but the fact there was a bug, we should ensure its tested now

witoszekdev: bug fix is only here: [`4d0ad0d` (#5934)](https://github.com/saleor/saleor-dashboard/pull/5934/commits/4d0ad0d3d34efacd471c84f1c86f75050e85083b) - the rest is to fix TypeScript errors, after I removed `@ts-strict-ignore` for this file.

Issue itself would have been caught by TypeScript if we didn't disable strict mode for this file, so I'm not sure we need to test that since it's a job for TypeScript compiler.

src/orders/utils/data.ts
```diff
@@ -367,40 +371,43 @@ export const getLineAvailableQuantityInWarehouse = (
 
 export const getLineAllocationWithHighestQuantity = (
   line: OrderFulfillLineFragment,
-): OrderFulfillLineFragment["allocations"][number] | undefined =>
-  line.allocations.reduce((prevAllocation, allocation) => {
-    if (!prevAllocation || prevAllocation.quantity < allocation.quantity) {
-      return allocation;
-    }
+): NonNullable<OrderFulfillLineFragment["allocations"]>[number] | undefined | null =>
```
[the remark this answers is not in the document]
witoszekdev: We need to leave it as is in order to not break usage of this utils

PR #5959, opened 2025-10-10
Remove Content Attributes from All Products column picker

IKarbowiak:
## Scope of the change

Remove Content Attributes from All Products column picker.

Now the column picker shows only `product_type` attributes
image

### DIFF
--- a/.changeset/blue-jokes-greet.md
+++ b/.changeset/blue-jokes-greet.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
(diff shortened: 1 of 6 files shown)


.changeset/large-showers-check.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Remove Content Attributes from `All Products` column picker.
```
witoszekdev:
```suggestion
Remove Content Attributes from `All Products` column picker on product list page. These attributes were never showing values if chosen.
```

PR #5976, opened 2025-10-14
Remove dead code and reorganize apps/extensions GraphQL layer

lkostrowski:
## Summary

This PR removes dead code from the apps module and reorganizes GraphQL queries/mutations into the extensions module where they are actually used.

### Changes

- **Removed dead code:**
  - `src/apps/components/AppFrame/useAppActions.ts` (86 lines)
  - `src/apps/components/AppFrame/usePostToExtension.ts` (20 lines)
  - `src/apps/mutations.ts` (175 lines of unused mutations)
  - `src/apps/queries.ts` (103 lines of unused queries)

- **Reorganized code:**
  - Moved `getPermissionsDiff.ts` and its test from `apps/` to `extensions/` module
  - Extended `extensions/mutations.ts` and `extensions/queries.ts` with the actually used GraphQL operations

- **Updated imports:**
  - Updated all import references across the codebase to reflect the new structure
  - Regenerated GraphQL types and hooks

### Impact

- Reduces codebase size by ~400 lines of dead code
- Improves code organization by placing GraphQL operations in the module where they're consumed
- No functional changes - purely refactoring

### DIFF
--- a/.changeset/fair-zebras-admire.md
+++ b/.changeset/fair-zebras-admire.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Added another bunch of refactors around apps and extensions. Less code, no functional changes


.changeset/fair-zebras-admire.md
witoszekdev, line 1: question: do we even need such changesets? maybe we can skip this for refactors?

lkostrowski: I decided to keep them because it will be visible what we release. You know, for ourselves. If we decide we want to hide them, we can remove them later in changelog

I don't have strong opinion here - I see benefits of both, but I just try to be defensive here because changes are large

PR #6010, opened 2025-10-28
Fix app extension URL validation to accept relative URLs

lkostrowski:
## Summary

This PR fixes the app extension manifest validation logic that was incorrectly rejecting relative URLs for POPUP, NEW_TAB, and WIDGET extension targets.

## What changed

The previous validation logic required all non-APP_PAGE extensions to use absolute URLs, but this was overly restrictive. The updated logic:

- **Accepts** relative URLs (starting with `/`) for POPUP, NEW_TAB, and WIDGET targets
- **Requires** relative URLs for APP_PAGE targets (rejects absolute URLs)
- Simplifies the validation logic for better readability

## Why this matters

Apps should be able to use relative URLs for their extension endpoints regardless of the target type. Only APP_PAGE extensions need to be constrained to relative URLs since they integrate directly with the dashboard routing.

## Test plan

- [x] Updated existing tests to reflect the new validation rules
- [x] Tests now verify that relative URLs are accepted for all non-APP_PAGE targets
- [x] APP_PAGE target still correctly rejects absolute URLs

### DIFF
--- a/.changeset/slimy-bikes-wink.md
+++ b/.changeset/slimy-bikes-wink.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Add missing logic for handling product variant metadata when accessed from order line modal (on order details page).
--- a/locale/defaultMessages.json
+++ b/locale/defaultMessages.json
@@ -312,10 +312,6 @@
     "context": "channels alphabetically title",
     "string": "Channels from A to Z"
   },
-  "/mwSjm": {
-    "context": "modal subheader, read-only product variant metadata",
-    "string": "This is a metadata of the variant that is being used in this ordered item"
-  },
   "/n+NRO": {
     "string": "Note id"
   },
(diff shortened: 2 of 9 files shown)


krzysztofzuraw (approved):

src/extensions/domain/app-extension-manifest.ts
```diff
@@ -60,21 +60,17 @@ export const appExtensionManifest = z
       const url = data.url;
       const target = data.target;
 
-      // Relative URL validation
-      if (url.startsWith("/")) {
-        // APP_PAGE can use relative URLs
-        return target === "APP_PAGE";
-      }
+      const isAppPage = target === "APP_PAGE";
+      const isRelativeUrl = url.startsWith("/");
```
witoszekdev, line 64: question: this looks to me more like an absolute URL, not relative 😅, shouldn't it start with `./`? 🤔

lkostrowski: it's not file, its url framgent:
- absolute - starts with http(s)
- relative - start with `/`, e.g. `/widget`

PR #6017, opened 2025-10-31
Update schema and set proper branch

lkostrowski:
## Scope of the change

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
   },
   "config": {
     "saleor": {
-      "schemaVersion": "main"
+      "schemaVersion": "3.22"
     }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -52,7 +52,7 @@
   },
   "config": {
     "saleor": {
-      "schemaVersion": "main"
+      "schemaVersion": "3.22"
```
lkostrowski, line 55: at this moment we should rely on 3.22 to avoid unnecessary break once we release

in the future we may add multi-schema build for conditional operations and switch them

krzysztofzuraw (approved):

PR #6030, opened 2025-11-05
Add real-time translation form integration with app extensions

lkostrowski:
## Summary

This PR adds support for real-time integration between translation forms and app extensions, enabling external apps to receive live translation data and suggest translations that are immediately reflected in the UI.

### Key Changes

**1. Real-time value tracking**
- Added `onValueChange` callbacks to all translation field components (Short, Long, Rich)
- Tracks current field values as users type (before submission)
- Uses a ref-based cache to maintain state without causing re-renders

**2. App extension integration**
- Emits translation form state to app extensions via `useActiveAppExtension` hook
- Provides original, current, and translated values for each field
- Receives translation suggestions from app extensions through "frames"
- Automatically displays app-suggested translations in the UI

**3. Multi-field editing support**
- Changes `activeField` from `string` to `string | string[]` to support editing multiple fields simultaneously
- Adds `getMultipleUrlValues()` utility to parse multiple query parameters with the same name
- Updates URL handling to support multiple `activeField` query parameters

**4. Improved navigation and state management**
- Fields remain in edit mode after submission (no automatic close)
- Individual fields can be discarded while keeping others active
- Better URL state synchronization with active fields
- Automatically opens fields that were modified by app extensions

### Technical Details

The implementation uses a combination of:
- **Data cache**: Ref-based cache tracks current values without triggering re-renders
- **Frame-based communication**: App extensions communicate suggestions via frames
- **Dirty field detection**: Compares app extension responses with cache to identify modified fields
- **URL state management**: Uses query parameters to track which fields are being edited

### Example Flow

1. User opens a translation field to edit
2. As they type, current values are cached and emitted to app extensions
3. App extension receives the data and can suggest translations
4. Suggested translations appear immediately in the form
5. Fields modified by the app extension are automatically opened for review
6. User can submit individual fields or discard changes per field

### Closes

https://github.com/saleor/saleor-dashboard/pull/5893/files#diff-3b7681f1d1389cb45fbb66a3ccef868a50f55306d445fccfabce6f0cf99a84b0

### DIFF
--- a/src/translations/urls.test.ts
+++ b/src/translations/urls.test.ts
@@ -1,0 +1,532 @@
+import {
+  languageEntitiesPath,
+  languageEntitiesUrl,
+  languageEntityPath,
+  languageEntityUrl,
+  languageListUrl,
+  productUrl,
+  productVariantUrl,
+  TranslatableEntities,
+} from "./urls";
+
+describe("translations/urls", () => {
+  describe("languageListUrl", () => {
+    // Arrange // Act // Assert
+
+    it("should return translations section path", () => {
+      expect(languageListUrl).toBe("/translations/");
+    });
+  });
+
+  describe("languageEntitiesPath", () => {
+    // Arrange // Act // Assert
+
+    it("should generate path for language code", () => {
+      const result = languageEntitiesPath("en");
+
+      expect(result).toBe("/translations/en");
+    });
+
+    it("should handle uppercase language codes", () => {
+      const result = languageEntitiesPath("EN");
+
+      expect(result).toBe("/translations/EN");
+    });
+
+    it("should handle language codes with region", () => {
+      const result = languageEntitiesPath("en-US");
+
+      expect(result).toBe("/translations/en-US");
+    });
+
+    it("should handle special characters in language code", () => {
+      const result = languageEntitiesPath("zh_CN");
+
+      expect(result).toBe("/translations/zh_CN");
+    });
+  });
+
+  describe("languageEntitiesUrl", () => {
+    // Arrange // Act // Assert
+
+    it("should generate URL with query parameters", () => {
+      const result = languageEntitiesUrl("en", {
+        after: "cursor123",
+        before: "cursor456",
+      });
+
+      expect(result).toContain("/translations/en?");
+      expect(result).toContain("after=cursor123");
+      expect(result).toContain("before=cursor456");
+    });
+
+    it("should handle tab parameter", () => {
+      const result = languageEntitiesUrl("en", {
+        tab: TranslatableEntities.products,
+      });
+
+      expect(result).toBe("/translations/en?tab=products");
+    });
+
+    it("should handle query parameter", () => {
+      const result = languageEntitiesUrl("en", {
+        query: "search term",
+      });
+
+      expect(result).toContain("/translations/en?");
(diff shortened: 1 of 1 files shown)


src/extensions/components/ExtensionsButtonSelector/ExtensionsButtonSelector.tsx
```diff
@@ -39,6 +39,40 @@ export const ButtonGroupWithDropdown = ({
...
+    return (
+      <Box {...boxProps}>
+        <Button
+          __height="50px"
+          variant={variant}
+          onClick={onClick}
+          data-test-id={testId}
+          __minWidth="80px"
+          disabled={disabled}
+          // TODO: fix this in Macaw UI - allow overriding border radius
```
witoszekdev, line 54: suggestion: i don't see any override for borderRadius 😉 

```suggestion
```

src/translations/components/TranslationFields/TranslationFields.tsx
```diff
@@ -166,6 +168,11 @@ const TranslationFields = (props: TranslationFieldsProps) => {
                         saveButtonState="default"
                         onDiscard={onDiscard}
                         onSubmit={undefined}
+                        onValueChange={v => {
+                          if (onValueChange) {
+                            onValueChange(field, v);
+                          }
+                        }}
```
witoszekdev, line 175: suggestion: we could make it a bit more readable with something like this

```
const getValueChangeHandler = (field) => v => onValueChange ? onValueChange(field, v) : undefined;
```

src/translations/components/TranslationFields/TranslationFieldsShort.tsx
```diff
@@ -14,6 +14,8 @@ interface TranslationFieldsShortProps {
   saveButtonState: ConfirmButtonTransitionState;
   onDiscard: () => void;
   onSubmit: (data: string) => SubmitPromise<any[]>;
+  // todo add to every field
```
witoszekdev, line 17: question: is this still valid?


```suggestion
```

lkostrowski: I will remove on next PR with cleanup

src/translations/components/TranslationFields/TranslationFields.tsx
lkostrowski: IMO it's much less readable :D

src/extensions/components/ExtensionsButtonSelector/ExtensionsButtonSelector.tsx
lkostrowski: let me fix that on cleanup pr

src/translations/components/TranslationsProductsPage/TranslationsProductsPage.tsx
```diff
@@ -55,6 +57,105 @@ const TranslationsProductsPage = ({
     productId,
     translationLanguage: languageCode,
   });
+  const { attachFormState, active, framesByFormType } = useActiveAppExtension();
+  // A hack to access field that are currently being edited in nested form.
+  const dataCache = useRef<
```
kzuraw, line 62: Consider exporting this to its own hook

PR #6042, opened 2025-11-07
Allow empty value for metadata

witoszekdev:
## Scope of the change

This PR adds possibility to search only for metadata key existence, by omitting `value` field in metadata input. This only works for filters where `WHERE` api is used instead of `FILTER`.

It also adds UI change to show "No value" in italics when user intentionally ommits value field.

It also adds validation to metadata, so that users cannot omit `key` - it's required for both APIs (where and filter)

## Example

**WHERE API**

https://github.com/user-attachments/assets/4297382e-2edf-414c-b40c-1d998c531bec

Shows order:
CleanShot 2025-11-07 at 12 24 43@2x

**FILTER API**

https://github.com/user-attachments/assets/c833658f-7905-4697-9834-b7b27231e69f

### DIFF
--- a/.changeset/extension-name-width.md
+++ b/.changeset/extension-name-width.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Increase extension name display width in InstalledExtensionsList from 200px to 400px to improve readability for extensions with longer names
--- a/src/extensions/views/InstalledExtensions/components/InstalledExtensionsList/InstalledExtensionsList.tsx
+++ b/src/extensions/views/InstalledExtensions/components/InstalledExtensionsList/InstalledExtensionsList.tsx
@@ -108,7 +108,7 @@ export const InstalledExtensionsList = ({
                     <Text
                       size={4}
                       fontWeight="bold"
-                      __maxWidth="200px"
+                      __maxWidth="400px"
                       overflow="hidden"
                       textOverflow="ellipsis"
                       whiteSpace="nowrap"


src/components/ConditionalFilter/FiltersQueryBuilder/queryVarsBuilders/MetadataFilterInputQueryVarsBuilder.test.ts
```diff
@@ -249,5 +249,24 @@ describe("MetadataAdvancedFilterQueryVarsBuilder", () => {
         AND: [{ someOtherField: "value" }, { metadata: { key: "color", value: { eq: "red" } } }],
       });
     });
+
+    it("includes key with empty value when both key and value are provided but value is empty", () => {
```
lkostrowski, line 253: what is the point of this? providing key without a value?

witoszekdev: yes, we have support for this via API, and when you search with `value: {eq: ""}` it doesn't return anything (even when the value is `""` - we'll need to check this on Saleor side as well). I think this might be useful if you want to check if you have a metadata _at all_, then you can filter for values.

witoszekdev: so now when you provide in the input only a key, it will return all items that have such key in their metadata, no matter the value, I've also added an indicator for this (see recording in PR description ;)

witoszekdev:
FYI we have error on `main` as well: https://github.com/saleor/saleor-dashboard/actions/runs/19166920617/job/54789684773?pr=6042

src/components/ConditionalFilter/Validation/metadata.ts
```diff
@@ -0,0 +1,25 @@
...
+  // Key is always required
+  if (!key || key.trim() === "") {
+    return {
+      row,
+      rightText: "Metadata key is required",
+    };
+  }
+
+  return false;
+};
```
lkostrowski, line 25: please avoid such cryptic functions. `metadata()` doesnt mean anything. Function has no typed return statement so we need to scan implementation what it does. Finally I see `metadata() === false` which is even more confusing

also if this is validator, common pattern is returning errors or null / undefined (and lets type function return statement)

PR #6044, opened 2025-11-07
Replace npm commands with pnpm in README

kzuraw:
## Scope of the change

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@
 
 - Node.js v22 (recommended)
 - A running instance of [Saleor](https://github.com/saleor/saleor/)
+- PNPM package manager - preferably installed via [corepack](https://pnpm.io/installation#using-corepack)
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,15 +59,15 @@ cd saleor-dashboard
 3. Install the dependencies:
 
 ```bash
-npm i
+pnpm install
```
NyanKiyoshi, line 63: Something like this instead? Not all users likely know how to install pnpm:

```suggestion
npm install -g pnpm
pnpm install
```

kzuraw: I think it is the best to use corepack - https://pnpm.io/installation#using-corepack as it reads pinned version of pnpm from package.json and installs that one: https://github.com/saleor/saleor-dashboard/blob/main/package.json#L242

kzuraw: Added it as prerequisite

PR #6047, opened 2025-11-10
Add order-level metadata dialog

kzuraw:
2025-11-10 at 14 23 44 - CleanShot@2x
2025-11-10 at 14 24 31 - CleanShot@2x

## Summary

- Move order metadata editing from inline card to dedicated dialog
- Add order line metadata dialog (based on existing variant metadata pattern)
- Add ripple indicator to highlight new feature location

### Order Metadata Dialog
- Created `OrderMetadataDialog` component for editing order-level metadata
- Moved metadata editing from inline card below transactions to modal dialog
- Integrated dialog in all three order views (Normal, Unconfirmed, Draft)
- Added "Update metadata" button in order details header (Code icon)

### Order Line Metadata Dialog
- Created `OrderLineMetadataDialog` component for editing order line metadata
- Follows same pattern as variant metadata dialog
- Handles both public and private metadata for order lines
- Integrated in fulfillment cards and unfulfilled products card

### Ripple Feature Indicator
- Added ripple indicator positioned at top-right corner of metadata button
- Configured with 3-day TTL to inform users of the new feature location
- Registered in ripples system

## Test plan

- [ ] Open order details page
- [ ] Verify ripple indicator appears on metadata button (Code icon)
- [ ] Click metadata button to open dialog
- [ ] Edit public and private metadata
- [ ] Verify changes saved correctly
- [ ] Test order line metadata dialog from line items
- [ ] Verify old inline metadata card removed from order details

### DIFF
--- a/.changeset/curly-squids-enter.md
+++ b/.changeset/curly-squids-enter.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Fixed broken sidebar app alert url to installed extension page
--- a/.changeset/free-numbers-wear.md
+++ b/.changeset/free-numbers-wear.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Dropped support for node 18. Migrated npm to pnpm
--- a/.github/workflows/dependency-check.yaml
+++ b/.github/workflows/dependency-check.yaml
@@ -52,10 +52,10 @@ jobs:
         env:
           latest: ${{ needs.current_state.outputs.main_statistics }}
         run: |
-          current=$(pnpx depcruise --config .dependency-cruiser.js src --output-type json | jq ".summary.info")
+          current=$(pnpm exec depcruise --config .dependency-cruiser.js src --output-type json | jq ".summary.info")
           difference=$(( $current - $latest ))
           echo "Imports in main: ${latest}"
-          echo "Incomming imports: ${current}"
+          echo "Incoming imports: ${current}"
           echo "Difference: ${difference}"
           echo "IMPORT_DIFFERENCE=${difference}" >> $GITHUB_ENV
       - name: "MacawUI regression reported"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 3.22.11
+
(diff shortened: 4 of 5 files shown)


src/orders/components/OrderLineMetadataDialog/OrderLineMetadataDialog.tsx
```diff
@@ -0,0 +1,247 @@
...
+}
+
+interface OrderMetadataDialogProps {
+  open: boolean;
+  onClose: () => void;
+  lineId: string;
+  orderId: string;
+}
+
+export const OrderLineMetadataDialog = ({
```
lkostrowski, line 35: Why don't we make it generic metadata dialog? I will look the same for any other metadata.

We can also use composition for that:

```
// generic UI
declare MetadataDialog = (props: {onClose, onSave, metadata: MetadataItem[], ...}) => ReactNode

const OrderMetadataDialog = (props: {onClose}) => {
  const { data, loading } = useMetadataValues({ orderId, lineId, open });
  const { onSubmit, lastSubmittedData, submitInProgress } = useHandleOrderLineMetadataSubmit({
    initialData: data,
    onClose,
  });
  
  // other order specific

  return <MetadataDialog metadata={data} onSave={makeMutation(...)} />

}
```

so we can compose any entity metadata almost automatically

src/ordersV2/OrderLines.tsx
lkostrowski, line 1: i guess this goes to trash

src/orders/components/OrderLineMetadataDialog/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    if (keys.some(key => key === "")) {
+      return intl.formatMessage({
+        defaultMessage: "Metadata key cannot be empty",
+        description: "metadata edit form, error message",
+        id: "lb5uDM",
+      });
+    }
+
+    return true;
```
lkostrowski, line 26: I thunk common for validators is to return either errors or null

current signature is bug prone because if you do

```
if (getValidateMetadata()){}
```

it will pass both values

instead

```
const errors = getValidateMetadata()

if(!errors) {...}
```

PR #6066, opened 2025-11-13
Order level metadata

kzuraw:
## Scope of the change
2025-11-13 at 14 30 05 - CleanShot@2x
image

### DIFF
--- a/locale/defaultMessages.json
+++ b/locale/defaultMessages.json
@@ -8158,6 +8158,9 @@
   "oJkeS6": {
     "string": "Attribute Code"
   },
+  "oL7VUz": {
+    "string": "Order Metadata"
+  },
   "oLMXDv": {
     "context": "order status",
     "string": "Ready to fulfill"


src/orders/components/OrderMetadataDialog/OrderMetadataDialog.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const privateMetadataErrors = useMemo(
+    () => flattenErrors(formState.errors.privateMetadata as FieldError),
+    [formState.errors.privateMetadata],
+  );
+
+  useEffect(() => {
+    if (!open) {
+      reset();
+    }
+  }, [open, reset]);
```
witoszekdev, line 155: question: shouldn't this be done in `MetadataDialog` itself?

witoszekdev (dismissed):

src/orders/components/OrderMetadataDialog/OrderMetadataDialog.tsx
kzuraw: I like that `MetadataDialog` is only for UI - no logic there (about forms etc)

PR #6145, opened 2025-11-29
Update throbber

mirekm:


.changeset/new-bottles-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Until we have better support for skeletons replace the spinning throbber with dash
```
lkostrowski: skeleton or not, we will have throbbers anyway IMO

src/components/Throbber/SaleorThrobber.tsx
```diff
@@ -0,0 +1,74 @@
+"use client";
```
lkostrowski: We don't have server side, this is redundant

lkostrowski (dismissed):
nitpick with "use client"

you need to resolve copilot comments before merging, check if any of them make sense, otherwise just close them

lkostrowski:
@mirekm tests are failing, probably you need to check if loading buttons are still accessible by e2e

PR #6148, opened 2025-12-01
Update order status and payment status pills

mirekm:
Side-by-side comparison of current Saleor implementation vs suggested semantic color system:

image

image

Semantics:
image

### DIFF
--- a/src/collections/components/CollectionListDatagrid/datagrid.test.ts
+++ b/src/collections/components/CollectionListDatagrid/datagrid.test.ts
@@ -15,7 +15,7 @@ describe("CollectionListDatagrid datagrid utils", () => {
     // Mock CSS.supports to enable OKLCH strings output
     global.CSS = {
       supports: () => true,
-      escape: s => s,
+      escape: (s: string) => s,
     } as any;
   });
 


lkostrowski:
Im not sure if partially returned and refunded is actually attention-level? Isn't partial refund the same severity as full refund? Why fully returned order has different meaning than order where only one product was returned?

src/components/Datagrid/customCells/PillCell.tsx
```diff
@@ -20,30 +20,36 @@ export const stringToHue = (str: string): number => {
 };
 
 const oklchIsSupported = typeof CSS !== "undefined" && CSS.supports("color: oklch(0% 0 0)");
-const oklch = (l: number, c: number, h: number) => {
+const oklch = (l: number, c: number, h: number, a?: number) => {
   if (oklchIsSupported) {
```
lkostrowski, line 24: I checked for compatbilitiy status and looks like its widely supported, feel free to simplify and assume its available

src/misc.ts
```diff
@@ -162,22 +162,22 @@ export const transformOrderStatus = (
...
       return {
         localized: intl.formatMessage(orderStatusMessages.unconfirmed),
-        status: StatusType.ERROR,
+        status: StatusType.NEUTRAL,
       };
     case OrderStatus.PARTIALLY_RETURNED:
       return {
         localized: intl.formatMessage(orderStatusMessages.partiallyReturned),
-        status: StatusType.INFO,
+        status: StatusType.ATTENTION,
```
lkostrowski, line 175: imo it should stay neutral, like before - both partial and full had the same color

mirekm:
> Isn't partial refund the same severity as full refund? Why fully returned order has different meaning than order where only one product was returned?

_Full refund_  → we're back to the initial state, the job is done
_Partial refund_  → is not back to the initial state (the job can be done too).

PR #6150, opened 2025-12-01
Fix and redesign Order history

mirekm:
Updated Order's timeline design:

image

### DIFF
--- a/.changeset/breezy-steaks-sit.md
+++ b/.changeset/breezy-steaks-sit.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Invalid event sent from app to dashboard will not throw anymore, but gracefully show notification
--- a/.changeset/eager-numbers-tan.md
+++ b/.changeset/eager-numbers-tan.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Update app-sdk package to 1.4.0
--- a/.changeset/frank-suns-attack.md
+++ b/.changeset/frank-suns-attack.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Changed Address component to use modern MacawUI text inputs
--- a/.changeset/happy-forks-drive.md
+++ b/.changeset/happy-forks-drive.md
@@ -1,6 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Updated `@saleor/app-sdk` to `1.5.0`.
-`next` should no longer be automatically installed with Saleor Dashboard. Previously this package installed into `node_modules`, even though it wasn't used.
--- a/.changeset/humble-boats-stay.md
+++ b/.changeset/humble-boats-stay.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Improve replaced icons visual consistency
--- a/.changeset/pretty-jobs-ring.md
+++ b/.changeset/pretty-jobs-ring.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Fixed attribute tooltip icon size
--- a/.changeset/small-baboons-eat.md
+++ b/.changeset/small-baboons-eat.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Clipboard operations no longer crash the website if browser permissions are not enabled
--- a/.changeset/sunny-bears-start.md
+++ b/.changeset/sunny-bears-start.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Icons overhaul part 1: Removed some old icons with fresh Lucide icons
--- a/.changeset/tangy-beds-camp.md
+++ b/.changeset/tangy-beds-camp.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Fix sizing of the main datagrid based views
--- a/.github/workflows/create-draft-release.yml
+++ b/.github/workflows/create-draft-release.yml
@@ -1,0 +1,95 @@
+name: Create Draft Release
+
+on:
+  push:
+    tags:
+      - "[0-9]+.[0-9]+.[0-9]+"
+
+permissions:
+  contents: write
+
+jobs:
+  create-draft-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # Fetch all history for all tags
+
+      - name: Extract changelog for version
+        id: changelog
+        run: |
+          VERSION="${GITHUB_REF#refs/tags/}"
+          echo "Creating release for version: $VERSION"
+
+          # Extract changelog section for this version
+          # This awk script extracts content between ## VERSION and the next ## line
+          CHANGELOG=$(awk -v version="$VERSION" '
+            /^## / {
+              if (found) exit;
+              if ($2 == version) found=1;
+              next;
+            }
+            found { print }
+          ' CHANGELOG.md)
+
+          if [ -z "$CHANGELOG" ]; then
+            echo "No changelog entry found for version $VERSION"
+            exit 1
+          fi
+
+          # Save changelog to file (multiline output)
+          echo "$CHANGELOG" > /tmp/changelog.txt
+          echo "changelog_file=/tmp/changelog.txt" >> $GITHUB_OUTPUT
+
+      - name: Get previous tag
+        id: prev_tag
+        run: |
+          CURRENT_TAG="${GITHUB_REF#refs/tags/}"
+
+          # Get all tags sorted by version, find the one before current
+          PREV_TAG=$(git tag --sort=-version:refname | grep -A1 "^${CURRENT_TAG}$" | tail -n1)
+
+          if [ -z "$PREV_TAG" ] || [ "$PREV_TAG" = "$CURRENT_TAG" ]; then
+            echo "No previous tag found"
+            echo "has_prev_tag=false" >> $GITHUB_OUTPUT
+          else
+            echo "Previous tag: $PREV_TAG"
+            echo "prev_tag=$PREV_TAG" >> $GITHUB_OUTPUT
+            echo "has_prev_tag=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Create release body
+        id: release_body
+        run: |
+          CURRENT_TAG="${GITHUB_REF#refs/tags/}"
+          CHANGELOG=$(cat /tmp/changelog.txt)
+
+          # Start building the release body
+          {
+            echo "$CHANGELOG"
+            echo ""
+
+            # Add diff link if we have a previous tag
+            if [ "${{ steps.prev_tag.outputs.has_prev_tag }}" = "true" ]; then
+              PREV_TAG="${{ steps.prev_tag.outputs.prev_tag }}"
+              echo "**Full Changelog**: https://github.com/${{ github.repository }}/compare/${PREV_TAG}...${CURRENT_TAG}"
+            fi
+          } > /tmp/release_body.txt
+
+          echo "release_body_file=/tmp/release_body.txt" >> $GITHUB_OUTPUT
+
+      - name: Create draft release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          CURRENT_TAG="${GITHUB_REF#refs/tags/}"
+          RELEASE_BODY=$(cat /tmp/release_body.txt)
+
+          gh release create "$CURRENT_TAG" \
+            --draft \
+            --title "$CURRENT_TAG" \
+            --notes "$RELEASE_BODY"
+
+          echo "✅ Draft release created for $CURRENT_TAG"
--- a/.github/workflows/deploy-staging-and-prepare-release.yaml
+++ b/.github/workflows/deploy-staging-and-prepare-release.yaml
@@ -44,6 +44,13 @@ jobs:
           publish: "pnpm run release"
         env:
           GITHUB_TOKEN: ${{ steps.get-token.outputs.token }}
+      - name: Update PR title with version from package.json
+        if: steps.changesets.outputs.pullRequestNumber != ''
+        run: |
+          VERSION=$(jq -r .version package.json)
+          gh pr edit ${{ steps.changesets.outputs.pullRequestNumber }} --title "Prepare release ${VERSION}"
+        env:
+          GH_TOKEN: ${{ steps.get-token.outputs.token }}
 
   build:
     needs: prepare-release
@@ -151,7 +158,7 @@ jobs:
           CDN_DISTRIBUTIONS: ${{ secrets.AWS_CLOUD_STAGING_CDN_DISTRIBUTIONS }}
 
       - name: Update check
-        if: always() && ${{ steps.create-check.outputs.CHECK_ID }}
+        if: always() && steps.create-check.outputs.CHECK_ID
         uses: ./.github/actions/checks/update-pull-request-check
         with:
           CHECK_ID: ${{ steps.create-check.outputs.CHECK_ID }}
--- a/.github/workflows/new-current-patch-release.yml
+++ b/.github/workflows/new-current-patch-release.yml
@@ -1,0 +1,67 @@
+name: New Current Patch Release
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+jobs:
+  # Patching current stable version is basically merging commits from main to release branch
+  # Since we ensure main is compatible with 3.22, instead of cherry-picking and porting, we
+  # just branch from main to current stable and create a release candidate. On this branch we run tests
+  # and release changelogs and tags.
+  # This workflow is automating this and ensuring branch protection is omitted for authorized actor.
+  recreate-release-branch:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Get Token
+        id: get-token
+        uses: saleor/saleor-internal-actions/request-vault-token@v1.4.0
+        with:
+          vault-url: ${{ secrets.VAULT_URL }}
+          vault-jwt: ${{ secrets.VAULT_JWT }}
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          token: ${{ steps.get-token.outputs.token }}
+
+      - name: Configure Git
+        run: |
+          git config user.name "Saleor Deployments"
+
+      - name: Recreate current release branch
+        run: |
+          CURRENT_RELEASE='3.22'
+
+          echo "Recreating release branch: $CURRENT_RELEASE"
+
+          # Fetch all branches
+          git fetch origin
+
+          # Switch to main branch
+          git checkout main
+          git pull origin main
+
+          # Delete the old release branch locally if it exists
+          if git show-ref --verify --quiet refs/heads/$CURRENT_RELEASE; then
+            echo "Deleting local branch $CURRENT_RELEASE"
+            git branch -D $CURRENT_RELEASE
+          fi
+
+          # Delete the remote release branch if it exists
+          if git ls-remote --heads origin $CURRENT_RELEASE | grep -q $CURRENT_RELEASE; then
+            echo "Deleting remote branch $CURRENT_RELEASE"
+            git push origin --delete refs/heads/$CURRENT_RELEASE
+          fi
+
+          # Create new branch from main
+          echo "Creating new branch $CURRENT_RELEASE from main"
+          git checkout -b $CURRENT_RELEASE
+
+          # Push the new branch
+          echo "Pushing new branch $CURRENT_RELEASE to remote"
+          git push origin refs/heads/$CURRENT_RELEASE:refs/heads/$CURRENT_RELEASE
+
+          echo "✅ Successfully recreated release branch $CURRENT_RELEASE from main"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -21,6 +21,8 @@ Before completing changes make sure you run these commands:
 - `pnpm run check-types` - Run TypeScript type checking
 - `pnpm run knip` - Check for unused files/dependencies/exports
 
+- ALWAYS run linter with autoformatter after you change the code, BEFORE you try to manually fix linter errors
+
 #### Test Scripts
 
 - `pnpm run test:quiet <file_path>` - RECOMMENDED: Run specific test with minimal output (--silent flag)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,32 @@
 # Changelog
 
+## 3.22.20
+
+### Patch Changes
+
+- [#6195](https://github.com/saleor/saleor-dashboard/pull/6195) [`c9ed5f7`](https://github.com/saleor/saleor-dashboard/commit/c9ed5f7d2ac75658218fc860ef4886bd9de95fe1) Thanks [@wcislo-saleor](https://github.com/wcislo-saleor)! - Added "Allow legacy gift card use" flag to Channel configuration
+
+- [#6188](https://github.com/saleor/saleor-dashboard/pull/6188) [`7833de8`](https://github.com/saleor/saleor-dashboard/commit/7833de81a816fc4298fd73be58d5377de2b98a00) Thanks [@witoszekdev](https://github.com/witoszekdev)! - Updated `@saleor/macaw-ui-next` to `1.4.1`. This version should resolve potential issues with React 18+ timing crashes
+
+## 3.22.19
+
+### Patch Changes
+
+- [#6172](https://github.com/saleor/saleor-dashboard/pull/6172) [`5373ccd`](https://github.com/saleor/saleor-dashboard/commit/5373ccdb3571e1294008224cf911bcd79f1b56b7) Thanks [@lkostrowski](https://github.com/lkostrowski)! - Update app-sdk package to 1.4.0
+
+- [#6177](https://github.com/saleor/saleor-dashboard/pull/6177) [`cdf694d`](https://github.com/saleor/saleor-dashboard/commit/cdf694dd083de4a12ebbe67b86ba56ed50f7d79a) Thanks [@lkostrowski](https://github.com/lkostrowski)! - Changed Address component to use modern MacawUI text inputs
+
+- [#6174](https://github.com/saleor/saleor-dashboard/pull/6174) [`dbe7709`](https://github.com/saleor/saleor-dashboard/commit/dbe7709d27287a9280119073f1d4217c15a39210) Thanks [@witoszekdev](https://github.com/witoszekdev)! - Updated `@saleor/app-sdk` to `1.5.0`.
+  `next` should no longer be automatically installed with Saleor Dashboard. Previously this package installed into `node_modules`, even though it wasn't used.
+
+- [#6185](https://github.com/saleor/saleor-dashboard/pull/6185) [`995c330`](https://github.com/saleor/saleor-dashboard/commit/995c330af5564c110d6226239289ece3281a8b43) Thanks [@mirekm](https://github.com/mirekm)! - Improve replaced icons visual consistency
+
+- [#6180](https://github.com/saleor/saleor-dashboard/pull/6180) [`4de2650`](https://github.com/saleor/saleor-dashboard/commit/4de265041e98106868c79b2b86113fdbf517704e) Thanks [@lkostrowski](https://github.com/lkostrowski)! - Fixed attribute tooltip icon size
+
+- [#6179](https://github.com/saleor/saleor-dashboard/pull/6179) [`8584e5f`](https://github.com/saleor/saleor-dashboard/commit/8584e5fee1adff5efb930dc1ebc9a9c3fb967c4d) Thanks [@lkostrowski](https://github.com/lkostrowski)! - Icons overhaul part 1: Removed some old icons with fresh Lucide icons
+
+- [#6186](https://github.com/saleor/saleor-dashboard/pull/6186) [`adddfde`](https://github.com/saleor/saleor-dashboard/commit/adddfde2b97f31e7a7b0b0f63fd4a684dcbdcd47) Thanks [@mirekm](https://github.com/mirekm)! - Fix sizing of the main datagrid based views
+
 ## 3.22.18
 
 ### Patch Changes
--- a/locale/defaultMessages.json
+++ b/locale/defaultMessages.json
@@ -4194,6 +4194,10 @@
     "context": "amount of refunded money",
     "string": "Amount"
   },
+  "OqYCQ4": {
+    "context": "allow legacy gift card use description",
+    "string": "When enabled, gift card can be applied to a checkout using addPromoCode mutation."
+  },
   "OrMr/k": {
     "context": "window title",
     "string": "Create Channel"
@@ -8837,6 +8841,10 @@
   "sDMNfc": {
     "context": "bulk issue structure item",
     "string": "Bulk Issue"
+  },
+  "sE7fI/": {
+    "context": "allow legacy gift card use label",
+    "string": "Allow legacy gift card use"
   },
   "sFynTT": {
     "context": "grant refund, refund card title",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "saleor-dashboard",
-  "version": "3.22.18",
+  "version": "3.22.20",
   "repository": {
     "type": "git",
     "url": "git://github.com/saleor/saleor-dashboard.git"
@@ -82,7 +82,7 @@
     "@reach/auto-id": "^0.16.0",
     "@react-editor-js/core": "2.1.0",
     "@saleor/macaw-ui": "npm:@saleor/macaw-ui@0.7.4",
-    "@saleor/macaw-ui-next": "npm:@saleor/macaw-ui@1.4.0",
+    "@saleor/macaw-ui-next": "npm:@saleor/macaw-ui@1.4.1",
     "@saleor/sdk": "0.6.0",
     "@sentry/react": "^8.55.0",
     "@sentry/vite-plugin": "^2.23.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -90,8 +90,8 @@ importers:
         specifier: npm:@saleor/macaw-ui@0.7.4
         version: 0.7.4(@material-ui/core@4.12.4(@types/react@18.3.23)(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@material-ui/icons@4.11.3(@material-ui/core@4.12.4(@types/react@18.3.23)(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@types/react@18.3.23)(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@material-ui/lab@4.0.0-alpha.61(@material-ui/core@4.12.4(@types/react@18.3.23)(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@types/react@18.3.23)(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@types/react@18.3.23)(react-dom@18.3.1(react@18.3.1))(react-helmet@6.1.0(react@18.3.1))(react@18.3.1)
       "@saleor/macaw-ui-next":
-        specifier: npm:@saleor/macaw-ui@1.4.0
-        version: "@saleor/macaw-ui@1.4.0(@types/react-dom@18.3.7(@types/react@18.3.23))(@types/react@18.3.23)(@vanilla-extract/css@1.17.5(babel-plugin-macros@3.1.0))(lucide-react@0.555.0(react@18.3.1))(react-dom@18.3.1(react@18.3.1))(react@18.3.1)"
+        specifier: npm:@saleor/macaw-ui@1.4.1
+        version: "@saleor/macaw-ui@1.4.1(@types/react-dom@18.3.7(@types/react@18.3.23))(@types/react@18.3.23)(@vanilla-extract/css@1.17.5(babel-plugin-macros@3.1.0))(lucide-react@0.555.0(react@18.3.1))(react-dom@18.3.1(react@18.3.1))(react@18.3.1)"
       "@saleor/sdk":
         specifier: 0.6.0
         version: 0.6.0(@apollo/client@3.4.17(graphql@16.11.0)(react@18.3.1))(graphql@16.11.0)(react@18.3.1)
@@ -4254,10 +4254,10 @@ packages:
       react-dom: ^16.8.0 || ^17.0.0
       react-helmet: ^6.1.0
 
-  "@saleor/macaw-ui@1.4.0":
+  "@saleor/macaw-ui@1.4.1":
     resolution:
       {
-        integrity: sha512-491oI8J6qrAXOsuWDoefdNZh6U42KpuJ4idaRFBMjaIIa2uYMyY+skOTTKGEWT+YMrzkf19FEXxMaIckjSrkJg==,
+        integrity: sha512-87ZkKnuZAhMrAprcy32zOTmguN33Ygb1JLeL0eaKmVM8/d2Pk+fYFn0tCMqrdhH7gAGNpT5h0FgL9/QlK7QlGQ==,
       }
     engines: { node: 20 || 22, pnpm: "10" }
     peerDependencies:
@@ -16322,7 +16322,7 @@ snapshots:
     transitivePeerDependencies:
       - "@types/react"
 
-  "@saleor/macaw-ui@1.4.0(@types/react-dom@18.3.7(@types/react@18.3.23))(@types/react@18.3.23)(@vanilla-extract/css@1.17.5(babel-plugin-macros@3.1.0))(lucide-react@0.555.0(react@18.3.1))(react-dom@18.3.1(react@18.3.1))(react@18.3.1)":
+  "@saleor/macaw-ui@1.4.1(@types/react-dom@18.3.7(@types/react@18.3.23))(@types/react@18.3.23)(@vanilla-extract/css@1.17.5(babel-plugin-macros@3.1.0))(lucide-react@0.555.0(react@18.3.1))(react-dom@18.3.1(react@18.3.1))(react@18.3.1)":
     dependencies:
       "@dessert-box/react": 0.4.0(react@18.3.1)
       "@floating-ui/react-dom": 2.1.6(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
--- a/schema-main.graphql
+++ b/schema-main.graphql
@@ -5446,6 +5446,20 @@ type CheckoutSettings {
   Added in Saleor 3.20.
   """
   automaticallyCompleteFullyPaidCheckouts: Boolean!
+
+  """
+  The time in minutes to wait after a checkout is fully paid before automatically completing it.
+  
+  Added in Saleor 3.22.
+  """
+  automaticCompletionDelay: Minute
+
+  """
+  The date time defines the earliest checkout creation date on which fully paid checkouts can begin to be automatically completed. 
+  
+  Added in Saleor 3.22.
+  """
+  automaticCompletionCutOffDate: DateTime
 }
 
 """Represents the channel-specific payment settings."""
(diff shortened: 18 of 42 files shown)


src/components/Timeline/Timeline.tsx
```diff
@@ -61,66 +13,72 @@ interface TimelineAddNoteProps {
...
+  buttonLabel?: string | React.ReactNode;
+  label?: string;
 }
 
-export const Timeline = (props: TimelineProps) => {
-  const { children } = props;
-  const classes = useStyles(props);
-
-  return <div className={classes.root}>{children}</div>;
+export const Timeline: React.FC<TimelineProps> = ({ children }) => {
```
lkostrowski: Use `props: TimelineProps` syntax instead of React.FC

src/components/Timeline/Timeline.tsx
```diff
@@ -61,66 +13,72 @@ interface TimelineAddNoteProps {
...
+}) => {
   const intl = useIntl();
-  const submit = e => {
+  const submit = (e: React.FormEvent<any>) => {
     reset();
     onSubmit(e);
   };
 
+  const handleKeyDown = (e: React.KeyboardEvent<HTMLTextAreaElement>) => {
+    if ((e.metaKey || e.ctrlKey) && e.key === "Enter" && !disabled && message.trim()) {
```
lkostrowski: usage of message.trim is confusing

lets change it to

```
const isMessageEmpty = message.trim().length === 0

if (... & !isMessageEmpty)
```

src/components/Timeline/TimelineEvent.tsx
```diff
@@ -1,7 +1,190 @@
...
+  });
+
+  if (!document.getElementById(styleId)) {
+    const style = document.createElement("style");
+
+    style.id = styleId;
+    style.textContent = eventStyles;
+    document.head.appendChild(style);
+  }
+}
```
lkostrowski: this is some strange workaround of injecting custom styles. I don't know exactly what is recommended way in Macaw/vanilla-extract, but you can check how to do this in Vite (probably you just import css file)

src/components/Timeline/TimelineEventHeader.tsx
```diff
@@ -10,13 +12,26 @@ export interface TitleElement {
   link?: string;
 }
 
+export interface TimelineUser {
```
lkostrowski: What if App makes an event?

src/components/Timeline/TimelineNote.tsx
```diff
@@ -1,13 +1,73 @@
...
+  }
+  .timeline-edit-link:hover,
+  .timeline-user-link:hover,
+  .timeline-edit-link:hover span,
+  .timeline-user-link:hover span {
+    text-decoration: underline;
+  }
+`;
+
+// Inject styles once
```
lkostrowski: get rid of css injection

src/components/Timeline/utils.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Safely stringify an object for display, handling circular references
+ * and removing __typename fields.
+ */
+export const safeStringify = (data: unknown): string => {
```
lkostrowski, line 10: should be tested

src/components/Timeline/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+
+          seen.add(value);
+        }
+
+        return value;
+      },
+      2,
+    );
+  } catch {
+    return "Unable to serialize";
```
lkostrowski: This should throw, not return string. This solution introduces nasty bugs exposed to client, that can't be properly caught

src/orders/components/OrderHistory/OrderHistory.tsx
```diff
@@ -2,25 +2,92 @@
...
+// Date group header component
+const DateGroupHeader = ({ label }: { label: string }) => (
+  <Box paddingY={3}>
+    <Text
+      size={2}
+      fontWeight="medium"
+      color="default2"
+      style={{ textTransform: "uppercase", letterSpacing: "0.05em" }}
+    >
+      {label}
```
lkostrowski: label is grouping key, not internationalized, so either accept plain value here and add translation in parent, or set finite prop/enum and translate it here

lkostrowski (changes requested):

src/components/Timeline/Timeline.tsx
```diff
@@ -61,66 +13,116 @@ interface TimelineAddNoteProps {
...
+  disabled,
+  placeholder,
+  buttonLabel,
+  label,
+}: TimelineAddNoteProps) => {
   const intl = useIntl();
-  const submit = e => {
+  const [isFocused, setIsFocused] = useState(false);
+  const isMessageEmpty = message.trim().length === 0;
+  const isMac =
```
lkostrowski: if you plan to add more keyboard helpers like this, we should extract this to separate components

src/components/Timeline/Timeline.tsx
```diff
@@ -61,66 +13,116 @@ interface TimelineAddNoteProps {
...
+    (navigator as any).userAgentData?.platform?.toLowerCase().includes("mac") ||
```
lkostrowski: why assertion on `any`? try `window.navigator`

src/components/Timeline/utils.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Safely stringify an object for display, handling circular references
+ * and removing __typename fields.
+ */
+export const safeStringify = (data: unknown): string => {
```
lkostrowski, line 10: add test

src/orders/components/OrderHistory/ExtendedTimelineEvent.tsx
```diff
@@ -143,107 +91,159 @@ interface ExtendedTimelineEventProps {
...
-                  amount,
-                  currency: orderCurrency,
-                }}
-              />
-            </>
-          )}
-          {shippingCostsIncluded && <Text>{intl.formatMessage(messages.refundedShipment)}</Text>}
-        </>
-      )}
+      {lines && lines.length > 0 ? (
```
lkostrowski: these are multiple nested conditions, lets split them to smaller components

src/orders/components/OrderHistory/ExtendedTimelineEvent.tsx
```diff
@@ -143,107 +91,159 @@ interface ExtendedTimelineEventProps {
...
+            {intl.formatMessage(messages.transactionReference)}
+          </Text>
+          <Text size={2}>{transactionReference}</Text>
+        </Box>
+      ) : null}
     </TimelineEvent>
   );
 };
 
+// eslint-disable-next-line import/no-default-export
```
lkostrowski: we are getting rid of `default` exports so best if instead of shutting down llinter, you just remove default export and use named one (export const ExtendedTimelineEvent)

src/components/Timeline/Timeline.tsx
```diff
@@ -61,66 +14,82 @@ interface TimelineAddNoteProps {
...
+            />
+          )}
+        </Button>
+      </Box>
+    </Box>
   );
 };
 
 Timeline.displayName = "Timeline";
+// eslint-disable-next-line import/no-default-export
```
mirekm: Fixing...

src/components/Timeline/Timeline.tsx
witoszekdev: suggestion: let's not use `React.FC`, I see that we already have `children` defined there, so it would clash with our definition

```js
const Timeline = ({children}: TimelineProps})
```

src/components/Timeline/Timeline.tsx
```diff
@@ -61,66 +13,72 @@ interface TimelineAddNoteProps {
...
+  onSubmit,
+  reset,
+  disabled,
+  placeholder,
+  buttonLabel,
+  label,
+}) => {
   const intl = useIntl();
-  const submit = e => {
+  const submit = (e: React.FormEvent<any>) => {
```
witoszekdev: issue: let's not use `any`, since we have already some `onSubmit` let's re-use type from there or use `unknown`

src/components/Timeline/Timeline.tsx
```diff
@@ -61,66 +13,72 @@ interface TimelineAddNoteProps {
...
+            />
+          )}
+        </Button>
+      </Box>
+    </Box>
   );
 };
 
 Timeline.displayName = "Timeline";
+// eslint-disable-next-line import/no-default-export
```
witoszekdev: issue: I think it's good idea to remove default exports, we want to get rid of all barell files

src/components/Timeline/TimelineEventHeader.tsx
```diff
@@ -26,50 +41,76 @@ const TimelineEventHeader = ({
...
+  tooltip,
+  user,
 }: TimelineEventHeaderProps) => {
   const elements = titleElements?.filter(Boolean) ?? [];
+  const userName = user ? getUserName(user, true) : null;
+
+  const userAttribution =
+    user && userName ? (
+      <Text size={3} color="default2" as="span" marginLeft={1}>
+        by{" "}
```
witoszekdev: issue: we should use `<FormattedMessage>` from react-intl for that, it has a way to format such strings, e.g. `by {userNameElement}` and in that placeholder we can render React component

src/components/Timeline/TimelineEventHeader.tsx
```diff
@@ -26,50 +41,76 @@ const TimelineEventHeader = ({
...
-            }
-
-            return (
-              <Text size={3} marginRight={0.5} key={`timeline-event-${text}`}>
-                {text}
-              </Text>
-            );
-          })}
+                );
+              })}
```
witoszekdev: suggestion: This could be separated into separate component to avoid nested if else logic in a `.map`

src/components/Timeline/TimelineNote.tsx
```diff
@@ -1,13 +1,73 @@
...
+  .timeline-user-link {
+    text-decoration: none;
+  }
+  .timeline-edit-link:hover,
+  .timeline-user-link:hover,
+  .timeline-edit-link:hover span,
+  .timeline-user-link:hover span {
+    text-decoration: underline;
+  }
+`;
```
witoszekdev: issue: let's use CSS Modules if we need to have custom styles 🙏🏻 - we already have 3 different ways of customizing styles and CSS Modules are by far the most used and cleanest

src/components/Timeline/TimelineNote.tsx
```diff
@@ -1,13 +1,73 @@
...
+  });
+
+  if (!document.getElementById(styleId)) {
+    const style = document.createElement("style");
+
+    style.id = styleId;
+    style.textContent = noteCardStyles;
+    document.head.appendChild(style);
+  }
+}
```
witoszekdev: issue: please, let's not use hacks to inject CSS, there is a global CSS file in Dashboard if it _must_ be done, but CSS modules are the way to go

src/components/Timeline/TimelineNote.tsx
```diff
@@ -23,6 +83,9 @@ interface TimelineNoteProps {
   relatedId?: string;
   onNoteUpdate?: (id: string, message: string) => Promise<unknown>;
   onNoteUpdateLoading?: boolean;
+  dateNode?: React.ReactNode;
+  eventData?: any;
```
witoszekdev: issue: let's not use `any`, if this is some expected object it should have it's own type

src/components/Timeline/TimelineNote.tsx
```diff
@@ -79,79 +114,171 @@ export const TimelineNote = ({
...
-                <FormattedMessage defaultMessage="Note id" id="/n+NRO" />: {id}
+                        if (card) {
+                          card.style.transition = "border-color 0.3s ease";
+                          card.style.borderColor = "hsla(0, 0%, 0%, 0.3)";
+                          setTimeout(() => {
+                            card.style.borderColor = "";
+                          }, 2000);
+                        }
+                      }
+                    }}
```
witoszekdev: suggestion: let's move this handler outside of `return`, let's have `const onNoteClick = ...` in the component

issue: We should assign styles by a state in react - e.g. we can have `const isFocused = useState()`, then in onClick we set state for `isFocused`, then we conditionally render className (we can merge different class names using cslx method)

src/components/Timeline/TimelineNoteEdit.tsx
```diff
@@ -1,9 +1,31 @@
...
+  const styleId = "timeline-edit-textarea-styles";
+
+  if (!document.getElementById(styleId)) {
+    const style = document.createElement("style");
+
+    style.id = styleId;
+    style.textContent = editTextareaStyles;
+    document.head.appendChild(style);
+  }
+}
```
witoszekdev: issue: let's use CSS modules

src/components/Timeline/utils.ts
witoszekdev, line 10: suggestion: I would rename this to `stringifyGraphQLObject` or something like this so that it's clear what we want this to be used for

src/components/UserAvatar/UserAvatar.tsx
```diff
@@ -7,11 +7,26 @@ interface UserAvatarProps {
...
+    backgroundColor: vars.colors.background.default3,
+    color: vars.colors.text.default1,
+  };
+  const mergedStyle = { ...defaultStyle, ...style };
+
+  return url ? (
+    <Avatar.User scheme={scheme} src={url} size={size} style={mergedStyle} {...rest} />
   ) : (
-    <Avatar.User scheme={scheme} initials={initials} {...rest} />
+    <Avatar.User scheme={scheme} initials={initials} size={size} style={mergedStyle} {...rest} />
```
witoszekdev, line 30: issue: This can be done via `backgroundColor` and `color` props, which already have `"default3"` and `"default1"` as options

src/orders/components/OrderHistory/ExtendedTimelineEvent.tsx
```diff
@@ -143,107 +91,161 @@ interface ExtendedTimelineEventProps {
...
+          </Text>
+          <Text size={2}>{transactionReference}</Text>
+        </Box>
       )}
     </TimelineEvent>
   );
 };
 
+// eslint-disable-next-line import/no-default-export
 export default ExtendedTimelineEvent;
```
witoszekdev: suggestion: let's remove default export since we're already changing this component (we can remove `index.ts` as well if it's used there)

src/orders/components/OrderHistory/OrderHistory.tsx
```diff
@@ -2,25 +2,92 @@
...
+const getDateGroupKey = (date: string | null): string => {
+  if (!date) {
+    return "UNKNOWN";
+  }
+
+  const eventDate = moment(date);
+  const now = moment();
+  const today = moment().startOf("day");
+  const yesterday = moment().subtract(1, "day").startOf("day");
+  const daysAgo = now.diff(eventDate, "days");
```
witoszekdev: issue: let's not use moment.js - we should get rid of it in Dashboard, since most of the operations can be done natively or with smaller utility libs like `date-fns` and simmilar.

src/orders/components/OrderHistory/OrderHistory.tsx
```diff
@@ -2,25 +2,92 @@
...
 
 import ExtendedTimelineEvent from "./ExtendedTimelineEvent";
 import { HistoryComponentLoader } from "./HistoryComponentLoader";
 import LinkedTimelineEvent from "./LinkedTimelineEvent";
 import { getEventMessage } from "./messages";
+import { OrderHistoryDate } from "./OrderHistoryDate";
 import { getEventSecondaryTitle, isTimelineEventOfType } from "./utils";
 
+// Helper to get date group key - smart grouping based on age
+const getDateGroupKey = (date: string | null): string => {
```
witoszekdev: let's have this in a separate file exported, so that we can test it

src/orders/components/OrderHistory/OrderHistory.tsx
```diff
@@ -2,25 +2,92 @@
...
+    return "LAST 7 DAYS";
+  } else if (daysAgo < 30) {
+    return "LAST 30 DAYS";
+  } else {
+    return "OLDER";
+  }
+};
+
+// Group events by date - preserves insertion order
+const groupEventsByDate = (events: OrderEventFragment[]): Array<[string, OrderEventFragment[]]> => {
```
witoszekdev: suggestion: let's have this in a separate file, so we can test this

src/orders/components/OrderHistory/OrderHistoryDate.tsx
```diff
@@ -0,0 +1,62 @@
...
+            } else if (diffDays < 7) {
+              dateStr = `${diffDays}d ago`;
+            } else {
+              dateStr = m.format("MMM D, YYYY");
+            }
+
+            // Full date with GMT for tooltip
+            const offset = m.format("Z");
+            const shortOffset = offset.endsWith(":00") ? offset.slice(0, -3) : offset;
+            const fullDate = `${m.format("MMM D, YYYY, h:mm A")} GMT${shortOffset}`;
```
witoszekdev: issue: this should be a separate function, so we can test this, we also shouldn't use moment.js. Date formating can be done via react-intl for correct language (this implementation has it hardcoded to English)

witoszekdev (changes requested):
For blocking comments from my side please see comments with `issue:`

src/components/Timeline/Timeline.tsx
mirekm: Outdated.

src/components/Timeline/Timeline.tsx
mirekm: Outdated

src/components/Timeline/TimelineEventHeader.tsx
mirekm: Outdated

src/orders/components/OrderHistory/ExtendedTimelineEvent.tsx
mirekm: Outdated

src/orders/components/OrderHistory/OrderHistory.tsx
mirekm: Outdated

src/orders/components/OrderHistory/OrderHistoryDate.tsx
mirekm: Outdated

src/orders/components/OrderHistory/OrderHistory.tsx
mirekm: Outdated

mirekm:
> For blocking comments from my side please see comments with `issue:`

Thanks for the review. Resolved.

src/orders/components/OrderHistory/utils.tsx
```diff
@@ -1,23 +1,86 @@
...
+  | "LAST_7_DAYS"
+  | "LAST_30_DAYS"
+  | "OLDER"
+  | "UNKNOWN";
+
+// Helper to check if two dates are the same day
+const isSameDay = (date1: Date, date2: Date): boolean =>
+  date1.getFullYear() === date2.getFullYear() &&
+  date1.getMonth() === date2.getMonth() &&
+  date1.getDate() === date2.getDate();
```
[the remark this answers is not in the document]
mirekm: Grouping is shown in local browser time as it's purely informative for the user. Authoritative UTC is show in the event's date tooltip.

mirekm:
~In progress: addressing additional feedback from @korycins~
Done.

src/components/CopyableText/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./CopyableText";
```
lkostrowski: do not create index files

src/components/KeyboardShortcut/KeyboardShortcut.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+interface KeyboardShortcutHintProps {
+  visible?: boolean;
+}
+
+/**
+ * Displays a keyboard shortcut hint (e.g., "Press ⌘ ↵ to send").
+ * Automatically uses the correct modifier key for the platform.
+ */
+export const KeyboardShortcutHint = ({
```
lkostrowski: 1. I think better name it "KeyboardShortcut" because hint sounds like some tooltip around it
2. It hardcodes `↵`, this should be used from props

src/components/SendFormKeyboardShortcutHint/SendFormKeyboardShortcutHint.tsx
```diff
@@ -0,0 +1,59 @@
...
+    }}
+  >
+    <Text size={1} color="default2">
+      <FormattedMessage
+        id="ILrXJV"
+        defaultMessage="Press {key1} {key2} to send"
+        values={{
+          key1: <KeyboardKey>{getModifierKey()}</KeyboardKey>,
+          key2: <KeyboardKey>↵</KeyboardKey>,
+        }}
```
lkostrowski, line 46: Why this is used as translatable string?

src/components/Timeline/TimelineEvent.module.css
lkostrowski, line 1: this should be css module

src/components/Timeline/TimelineEventHeader.tsx
```diff
@@ -10,68 +13,122 @@ export interface TitleElement {
...
 }
 
+export interface TimelineUser {
+  id: string;
+  email: string;
+  firstName?: string;
+  lastName?: string;
+  avatar?: {
+    url?: string;
+  } | null;
```
lkostrowski: this is ui component, you can simplify props to `avatarUrl?: string`, it doesnt have to map graphql

src/components/Timeline/TimelineEventHeader.tsx
```diff
@@ -10,68 +13,122 @@ export interface TitleElement {
...
+}
+
+export interface TimelineApp {
+  id: string;
+  name?: string | null;
```
lkostrowski: try to reduce ambiguity, either `?` or `null`

src/components/Timeline/TimelineEventHeader.tsx
```diff
@@ -10,68 +13,122 @@ export interface TitleElement {
...
-  secondaryTitle,
   hasPlainDate,
   children,
+  dateNode,
+  tooltip,
+  user,
+  app,
 }: TimelineEventHeaderProps) => {
   const elements = titleElements?.filter(Boolean) ?? [];
+  const userName = user ? getUserName(user, true) : null;
```
lkostrowski: you provide entire use just to get its name. This is wrong design. Component should accept only prop it needs. can be

```
actorName: string;
actorType: "app" | "user"
```

lkostrowski (changes requested):

src/components/Timeline/TimelineNote.tsx
```diff
@@ -81,79 +41,194 @@ export const TimelineNote = ({
...
-            {id && (
-              <Text size={2} color="defaultDisabled">
-                <FormattedMessage defaultMessage="Note id" id="/n+NRO" />: {id}
+                        const card = element.querySelector("[data-note-card]") as HTMLElement;
+
+                        if (card) {
+                          card.style.borderColor = "hsla(0, 0%, 0%, 0.3)";
+                          setTimeout(() => {
+                            card.style.borderColor = "";
+                          }, 2000);
```
lkostrowski: its unsafe usage of timeout which cause memory leaks

you need to extract this to useEffect and ensure timeout is cleaned on unmount

also, for code clarity, dont keep non-trivial callbacks inline, but extract them to component body

src/components/Timeline/TimelineNote.tsx
```diff
@@ -25,52 +26,11 @@ interface TimelineNoteProps {
   relatedId?: string;
   onNoteUpdate?: (id: string, message: string) => Promise<unknown>;
   onNoteUpdateLoading?: boolean;
+  dateNode?: React.ReactNode;
```
lkostrowski:
```
date: string | ReactNode
```

```
const dateToRender = typeof date === "string"? <DateTime>...</DateTime> : date
```

src/components/KeyboardShortcut/KeyboardShortcut.tsx
mirekm: But it's a hint. Not keyboard shortcut itself. I would keep the current name.

mirekm: > It hardcodes ↵, this should be used from props

👍

src/orders/components/OrderHistory/OrderHistoryDate.tsx
```diff
@@ -0,0 +1,119 @@
+// @ts-strict-ignore
```
lkostrowski: don't add this clause. It's for OLD code that is using non-strcit / invalid types. We are refactoring this (by removing this statement and fixing type errors), but new code must be always strict

src/components/SendFormKeyboardShortcutHint/SendFormKeyboardShortcutHint.tsx
mirekm: Is there an argument why `Press {key1} {key2} to send` should not be translatable?

src/orders/components/OrderHistory/OrderHistoryDate.tsx
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Calculate relative date string and full date for tooltip.
+ * Extracted for testability.
+ */
+export const getRelativeDate = (
+  date: string,
+  intl: IntlShape,
+  tz?: string,
+  now = new Date(),
+): RelativeDateResult => {
```
lkostrowski: Use named attributes (object destructuring), otherwise it causes problems and code looks like this

```
getRelativeDate(date, intl, undefined, date2)
```

which sucks, because you can't check function signature without accessing it's implementation

instead

```
const getRelativeDate = (params: {  date: string,
  intl: IntlShape,
  tz?: string,
  now?: Date}) {}
```

(like kwargs in python)

src/orders/components/OrderHistory/utils.tsx
```diff
@@ -1,23 +1,86 @@
...
+const startOfDay = (date: Date): Date => {
+  const result = new Date(date);
+
+  result.setHours(0, 0, 0, 0);
+
+  return result;
+};
+
+// Helper to get date group key - smart grouping based on age
+export const getDateGroupKey = (date: string | null, now?: Date): DateGroupKey => {
```
lkostrowski: why do you even pass "now" to component? can "now" be not now?

lkostrowski (changes requested):

src/components/SendFormKeyboardShortcutHint/SendFormKeyboardShortcutHint.tsx
mirekm: Since it's not a generic reusable component for any shortcut (translation) I will change the name for now to `SendFormKeyboardShortcutHint`: we still need a reusable component like that, at the same time the translation works as is.

src/components/Timeline/TimelineNote.tsx
mirekm: Yep, the timeout is not needed there in the first place. It should be just css animation. Fixing.

src/orders/components/OrderHistory/utils.tsx
mirekm: Was added for testibility :) but it should be removed. Fixing.

src/components/SendFormKeyboardShortcutHint/SendFormKeyboardShortcutHint.tsx
lkostrowski, line 1: rename file to match component name

src/components/KeyboardShortcut/KeyboardShortcut.tsx
```diff
@@ -0,0 +1,67 @@
...
+        }}
+      />
+    </Text>
+  </Box>
+);
+
+/**
+ * Returns the correct modifier key symbol for the platform (⌘ for Mac, Ctrl for others).
+ */
+export { getModifierKey };
```
lkostrowski: why is it exported? if for tests only, it's antipattern, in such case better just assert on rendered value

src/components/Timeline/Timeline.module.css
```diff
@@ -0,0 +1,47 @@
+/* Timeline event styles */
```
lkostrowski: File name is "Timeline.module" which by convention means these are styles for "Timeline". But Timeline is not even using it.

Move styles matching components, eg `TimelineEvent.module.css`

src/components/Timeline/TimelineEvent.tsx
```diff
@@ -1,64 +1,363 @@
...
+    app,
+    eventType,
+    isLastInGroup,
+  } = props;
   const hasChildren = children && React.Children.toArray(children).filter(Boolean).length > 0;
 
+  // Transform user/app to actor props for TimelineEventHeader
+  const actorName = user ? getUserName(user, true) : (app?.name ?? undefined);
+  const actorType = user ? "user" : app ? "app" : undefined;
+  const actorId = user?.id ?? app?.id;
```
lkostrowski: you didn't get my comment. It's not about component internal to map 2 prop into one but the opposite - public API (props) should guard against passing invalid data, like passing both user and app at the same time


So it's about having one prop like `actor` or `userOrApp` with type that can be either of them and inside component you can extract what has been passed. 

Typescript will ensure only one of them is passed

And if it happens that your code doesn't have to distinguish between app and user, that's even better because html is cleaner

src/components/Timeline/TimelineEvent.tsx
```diff
@@ -1,64 +1,363 @@
...
+  id: string;
+  email: string;
+  firstName?: string;
+  lastName?: string;
+}
+
+interface TimelineEventApp {
+  id: string;
+  name?: string | null;
+}
```
lkostrowski: didn' t you loose the avatar? I think both user and app have it

src/components/Timeline/TimelineEvent.tsx
```diff
@@ -1,64 +1,363 @@
...
 
 export interface TimelineEventProps {
   children?: React.ReactNode;
   date: string;
-  secondaryTitle?: string;
   title?: React.ReactNode;
   titleElements?: TitleElement[];
   hasPlainDate?: boolean;
+  dateNode?: React.ReactNode;
```
lkostrowski: in my previous review I wrote about merging `date` and `dateNode` props

src/components/Timeline/TimelineEventHeader.tsx
```diff
@@ -14,64 +17,110 @@ interface TimelineEventHeaderProps {
...
-          {title}
+  const getActorLink = () => {
+    if (!actorId) return null;
+
+    if (actorType === "user") return staffMemberDetailsUrl(actorId);
+
+    if (actorType === "app")
+      return ExtensionsPaths.resolveViewManifestExtension(encodeURIComponent(actorId));
+
+    return null;
```
lkostrowski: why TimelineEvent and TimelineHeader do it differently?

src/components/Timeline/TimelineNote.tsx
```diff
@@ -1,22 +1,24 @@
...
 import { TimelineNoteEdit } from "./TimelineNoteEdit";
+import { safeStringify } from "./utils";
 
 type TimelineAppType =
   | NonNullable<GiftCardDetailsQuery["giftCard"]>["events"][0]["app"]
   | OrderEventFragment["app"];
 
 interface TimelineNoteProps {
-  date: string;
+  date: string | React.ReactNode;
```
lkostrowski, line 15: Apply this pattern to other components, not only where I added the comment

src/components/Timeline/TimelineNote.tsx
```diff
@@ -81,79 +41,200 @@ export const TimelineNote = ({
...
+    element.scrollIntoView({ behavior: "smooth", block: "center" });
+
+    const card = element.querySelector("[data-note-card]") as HTMLElement;
+
+    if (card) {
+      card.classList.add(styles.noteCardHighlight);
+      card.addEventListener("animationend", () => card.classList.remove(styles.noteCardHighlight), {
+        once: true,
+      });
+    }
```
lkostrowski, line 87: this is not valid react code that leaves memory leak

you can either use react state that will control the classes or do it vanilla-way, but you need to clean up event listeners on useEffect

src/orders/components/OrderHistory/ExtendedTimelineEvent.tsx
```diff
@@ -143,107 +129,128 @@ interface ExtendedTimelineEventProps {
...
   hasPlainDate?: boolean;
+  dateNode?: React.ReactNode;
+  isLastInGroup?: boolean;
 }
 
 const ExtendedTimelineEvent = ({
   event,
   orderCurrency,
   hasPlainDate,
+  dateNode,
```
lkostrowski: same thing with dates, this is really confusing that we allow two dates with ambiguity. I think root component like Timeline should make a decision what date is available and pass it down. Other components should not prop-drill not needed components just to low-level component make a decision which to display

src/orders/components/OrderHistory/OrderHistory.tsx
```diff
@@ -128,39 +200,174 @@ const OrderHistory = ({
...
                       hasPlainDate={false}
+                      dateNode={<OrderHistoryDate date={date} />}
+                      isLastInGroup={isLastInGroup}
                     />
                   );
                 }
 
-                if (isTimelineEventOfType("linked", type)) {
-                  return <LinkedTimelineEvent event={event} key={id} hasPlainDate={false} />;
+                // Events with message, relatedOrder, or lines - make them foldable
```
lkostrowski: instead of a comment, code should say that 

```
const isEventFoldable = Boolean(message || hasRelatedOrder || hasLines)
```

src/components/Timeline/TimelineNote.tsx
mirekm: Would clearing it up in the timeout be a reasonable balance?

``` typescript
const highlightTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);

  useEffect(() => {
    return () => {
      if (highlightTimeoutRef.current) {
        clearTimeout(highlightTimeoutRef.current);
      }
    };
  }, []);
  ```
  
  ``` typescript
  if (card) {
      card.classList.add(styles.noteCardHighlight);

      if (highlightTimeoutRef.current) {
        clearTimeout(highlightTimeoutRef.current);
      }

      highlightTimeoutRef.current = setTimeout(() => {
        card.classList.remove(styles.noteCardHighlight);
      }, 2000); // Match CSS animation duration
    }
```

src/components/KeyboardShortcut/KeyboardShortcut.tsx
mirekm: Leftover from the transition from generic shortcut hint component to the generic send form hint component. I've also moved CSS to the module as I discovered it was inline.

src/components/Timeline/Timeline.module.css
mirekm: Yes, those should be split between TimelineNote and TimelineEvents.

src/components/Timeline/TimelineEvent.tsx
mirekm: In the Timeline component icons represent the type of event. So, I decided not to show avatars on top of that at the moment. Instead of avatars the name of user/app is presented.

src/components/Timeline/TimelineEvent.tsx
mirekm: Got it. Refactored.

src/components/Timeline/TimelineEventHeader.tsx
mirekm: I unified it by moving some of the helper functions to utils

src/orders/components/OrderHistory/ExtendedTimelineEvent.tsx
mirekm: Ok, let's the caller decide about the format.

mirekm:
Code review issues addressed.

src/orders/components/OrderHistory/utils.tsx
```diff
@@ -1,25 +1,86 @@
...
+    const index = groupMap.get(key)!;
+
+    groups[index][1].push(event);
+  });
+
+  return groups;
+};
+
 const timelineEventTypes = {
   discount: [
```
witoszekdev, line 85: question: what does this change mean? we won't show these order events, or will they be shown in different category? 🤔

lkostrowski (dismissed):

src/orders/components/OrderHistory/OrderHistoryDate.tsx
```diff
@@ -0,0 +1,124 @@
...
+    const days = Math.floor(diffInSeconds / DAY);
+
+    dateStr = intl.formatMessage(
+      {
+        id: "pSFOFy",
+        defaultMessage: "{days}d ago",
+        description: "relative time in days",
+      },
+      { days },
+    );
```
witoszekdev, line 90: suggestion: this could use Intl.RelativeTimeFormat instead of strings for formatting, this will apply correct language without any work with translations

https://formatjs.github.io/docs/react-intl/components#formattedrelativetime

It's supported in all browsers since 2020: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat

src/orders/components/OrderHistory/OrderHistory.tsx
```diff
@@ -128,39 +197,167 @@ const OrderHistory = ({
...
+                      title={
+                        <Box display="flex" gap={1} alignItems="center" flexWrap="wrap">
+                          <FormattedMessage id="KpP/aW" defaultMessage="Order was marked as" />
+                          <Pill
+                            color="success"
+                            size="small"
+                            label={intl.formatMessage({
+                              id: "u/vOPu",
+                              defaultMessage: "Paid",
+                            })}
```
witoszekdev, line 233: suggestion: we should re-use existing translations strings / or add context within the message

src/orders/components/OrderHistory/OrderHistory.tsx
```diff
@@ -128,39 +197,167 @@ const OrderHistory = ({
...
+                            style={{ paddingLeft: 4, paddingRight: 4 }}
```
witoszekdev, line 234: question: what is this style unit? px / em / rem? we should probably use `paddingLeft` from macaw or `__paddingLeft`

src/orders/components/OrderHistory/OrderHistory.tsx
```diff
@@ -128,39 +197,167 @@ const OrderHistory = ({
...
+                  return (
+                    <TimelineEvent
+                      date={<OrderHistoryDate date={date} />}
+                      eventData={event}
+                      actor={toActor(user, app)}
+                      eventType={type}
+                      isLastInGroup={isLastInGroup}
+                      title={
+                        <Box display="flex" gap={1} alignItems="center" flexWrap="wrap">
+                          <FormattedMessage id="KpP/aW" defaultMessage="Order was marked as" />
```
witoszekdev, line 226: suggestion: This can get tricky for translations, if for some reason "paid" would need to be before this string for a proper sentence.

We can use react-intl for this as well, example:

```ts
<FormattedMessage
    id="order.statusChange"
    defaultMessage="Order was marked as {status}"
    values={{
      // We define 'status' as the actual Pill component
      status: (
        <Pill
          color="success"
          size="small"
          // The label inside the pill is still translated normally
          label={intl.formatMessage({
            id: "u/vOPu",
            defaultMessage: "Paid",
          })}
        />
      ),
    }}
  />
```

src/orders/components/OrderHistory/OrderHistory.tsx
```diff
@@ -91,32 +135,57 @@ const OrderHistory = ({
...
-              .map(event => {
+            {(() => {
+              const reversedHistory = history.slice().reverse();
+              const groupedEvents = groupEventsByDate(reversedHistory);
+
+              const renderEvent = (
+                event: OrderEventFragment,
+                index: number,
+                groupEvents: OrderEventFragment[],
+              ) => {
```
witoszekdev, line 173: suggestion: this looks like it could be it's own component not a render method

src/orders/components/OrderHistory/OrderHistory.tsx
```diff
@@ -91,32 +135,57 @@ const OrderHistory = ({
...
                 />
               )}
             </Form>
-            {history
-              .slice()
-              .reverse()
-              .map(event => {
+            {(() => {
+              const reversedHistory = history.slice().reverse();
+              const groupedEvents = groupEventsByDate(reversedHistory);
```
witoszekdev, line 167: suggestion: let's use `useMemo` for this - right now we'll have to calculate it on every re-render, it can get problematic if we have a lot of order events

src/orders/components/OrderHistory/OrderHistory.tsx
```diff
@@ -91,32 +135,57 @@ const OrderHistory = ({
...
+                  label={intl.formatMessage({
+                    id: "LgbKvU",
+                    defaultMessage: "Comment",
+                  })}
+                  buttonLabel={
+                    <FormattedMessage
+                      id="H5jL5+"
+                      defaultMessage="Add Comment"
+                      description="button"
+                    />
```
witoszekdev, line 160: suggestion: let's move messages to separate file, I'm sure we already have a string for "Comment", this way we can re-use existing translations

`description="button"` is not very clear for translation, better one would give where this is used, e.g. `order history, button`

src/orders/components/OrderHistory/OrderHistory.tsx
```diff
@@ -1,25 +1,84 @@
...
+// Date group header component with internationalized labels
+const DateGroupHeader = ({ groupKey }: { groupKey: string }) => {
+  const intl = useIntl();
+
+  const getLabel = (key: string): string => {
+    switch (key) {
+      case "TODAY":
+        return intl.formatMessage({
+          id: "zWgbGg",
+          defaultMessage: "Today",
```
witoszekdev, line 34: suggestion: missing description

```suggestion
          defaultMessage: "Today",
          description: "date group header",
```

src/components/Timeline/TimelineNoteEdit.tsx
```diff
@@ -1,9 +1,31 @@
...
+if (typeof document !== "undefined") {
+  const styleId = "timeline-edit-textarea-styles";
+
+  if (!document.getElementById(styleId)) {
+    const style = document.createElement("style");
+
+    style.id = styleId;
+    style.textContent = editTextareaStyles;
+    document.head.appendChild(style);
+  }
```
witoszekdev: issue: let's not inject styles, instead we can use CSS modules / `style={}` property or `Box` component

If we need global overrides there's also: `src/index.css` for this

src/components/Timeline/TimelineNote.tsx
```diff
@@ -1,159 +1,247 @@
...
+
+      if (highlightTimeoutRef.current) {
+        clearTimeout(highlightTimeoutRef.current);
+      }
+
+      highlightTimeoutRef.current = setTimeout(() => {
+        card.classList.remove(styles.noteCardHighlight);
+      }, 2000); // Match CSS animation duration
+    }
+  }, [relatedId]);
```
witoszekdev, line 88: issue: let's not query DOM elements directly and instead rely on React for this.

1. document.getElementById can be replaced with `useRef` and assigning `ref` to the element we want to track by

```ts
const timelineNoteRef = useRef();

// ...

<Box
  // ...
  ref={timelineNoteRef}
```

2. styles can be assigned by storing a value in useState and then applying it conditionally

```ts
const [cardHighlighted, setCardHighilighted] = useState(null);

// ...

useEffect(() => {
  if (cardHighlighted !== null) {
     // in here set timeout, assign it to ref
  }

  return () => {
    // clear timeout here
  }
}, [cardHighlighted])

// in handler
setCardHihglihted(id);


```

src/components/Timeline/TimelineNoteEdit.tsx
lkostrowski: agree, there are framework-specific ways to add styles, this should be css modules

src/components/Timeline/TimelineNote.tsx
lkostrowski: While for me it is rather non blocking, I agree with Jonatan that more react-way is to add it into state, which makes it more declarative too

src/components/Timeline/TimelineNoteEdit.tsx
mirekm: 💯

src/components/Timeline/TimelineNote.tsx
mirekm: I looked at it yesterday and decided not to fix it to ref on this stage but I agree, it's ugly and should be ref. I will fix it tho - let's promote the right patterns and do things idiomatically, should have done it once I spotted it.

witoszekdev: let's leave this for now then - i see now that this is a _list_ which would make this ref approach more tricky because we would need to store ref to all elements, or a single ref and then do a narrow down querySelector to find child, which at this point we can just use what we have

mirekm: Ok, I looked into it and because it's a cross-component interaction React-way may turn it into something more complex (maybe than we want). I would need to have a Provider with some context to track the references record or Notes. Not sure it is worth it.

PR #6165, opened 2025-12-03
Add modal with filtering logic for Products

witoszekdev:
Added filtering logic to Products modal:

CleanShot 2025-12-03 at 17 36 45@2x

It respects constraints, e.g. when selecting a product for reference attribute that has restriction to a specific variant:

CleanShot 2025-12-03 at 17 36 38@2x

### DIFF
--- a/.changeset/beige-gifts-say.md
+++ b/.changeset/beige-gifts-say.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Transaction actions: The Capture button is now directly visible on transaction cards for quicker access, while destructive actions like Cancel remain safely tucked in the menu.
--- a/.changeset/crisp-shoes-care.md
+++ b/.changeset/crisp-shoes-care.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Update order Transactions cards
--- a/.changeset/lemon-candies-crash.md
+++ b/.changeset/lemon-candies-crash.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Introduced a redesigned "Order summary" section that unifies order details and payment information across all order types, including Drafts and Unconfirmed orders. The updated "Order value" breakdown now clearly separates subtotal, taxes, discounts, and shipping. Additionally, a new "Payments summary" section has been added, featuring a dedicated "no data" state when no payments are present.
--- a/.changeset/wide-signs-love.md
+++ b/.changeset/wide-signs-love.md
@@ -1,5 +1,0 @@
----
-"saleor-dashboard": patch
----
-
-Fixed redirection between Customer Details -> See all orders to Orders List (which this customer selected). Now filter is properly set on the URL and only relevant orders are displayed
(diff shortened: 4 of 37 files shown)


src/components/AssignProductDialog/ModalProductFilterProvider.tsx
```diff
@@ -0,0 +1,61 @@
...
+          filterChannel,
+          clearFilters,
+          hasActiveFilters,
+        }}
+      >
+        {children}
+      </ModalProductFilterContext.Provider>
+    </ConditionalFilterContext.Provider>
+  );
+};
```
witoszekdev, line 51: todo: we should rewrite this to use jotai - but this would require bigger changes to our modals

src/components/AssignProductDialog/useModalProductFilter.ts
```diff
@@ -0,0 +1,240 @@
...
+/**
+ * Creates a wrapped value provider that injects a constraint element and handles
+ * GLOBAL constraint-specific behavior for persistence, counting, and token lookup.
+ *
+ * What it does:
+ * 1. Inject constraint element at the beginning of filter value
+ * 2. Strip GLOBAL constraints when persisting to URL
+ * 3. Exclude constraint from count
+ */
+export const createWrappedValueProvider = (
```
witoszekdev, line 128: this is needed so that we can show "productType" when we have reference attribute constraints

it's fragile because it wraps existing state management for filters, but otherwise we would need bigger refactor to ConditionalFilter and Condition

src/components/ConditionalFilter/useContainerState.ts
```diff
@@ -1,11 +1,23 @@
...
+export interface UseContainerStateOptions {
+  /**
+   * Only sync with valueProvider on first mount, not on subsequent changes.
+   *
+   * Use this for modals where `valueProvider.value` is wrapped/computed
+   * (e.g., with injected constraints). The wrapping creates new array
+   * references on each render, causing unwanted resyncs that overwrite
+   * user's unsaved edits.
+   */
+  syncOnce?: boolean;
```
witoszekdev, line 18: this is bandaid for poor design - we should refactor our state management depending on URLs into useSyncExternalStore in React 18+

src/products/components/ProductUpdatePage/ProductUpdatePage.tsx
```diff
@@ -125,6 +127,12 @@ interface ProductUpdatePageProps {
   onImageUpload: (file: File) => any;
   onMediaUrlUpload: (mediaUrl: string) => any;
   onSeoClick?: () => any;
+  onProductFilterChange?: (
+    filterVariables: ProductWhereInput,
+    channel: string | undefined,
+    query: string,
+  ) => void;
+  initialConstraints?: InitialConstraints;
```
witoszekdev, line 135: we should use jotai instead of adding yet another props but again... even more scope

witoszekdev (commented):
fyi some things are hacks around our filter logic, we should probably change our abstractions and component architecture as mentioned in my comments

src/components/AssignProductDialog/AssignProductDialog.tsx
```diff
@@ -15,36 +17,59 @@ export interface AssignProductDialogProps extends FetchMoreProps, DialogProps {
   productUnavailableText?: string;
   selectedIds?: Record<string, boolean>;
   loading: boolean;
-  onFetch: (value: string) => void;
+  onFilterChange?: (
+    filterVariables: ProductWhereInput,
+    channel: string | undefined,
+    query: string,
+  ) => void;
```
lkostrowski, line 24: make arguments object, we should need to write `f(a, undefined, b)`

PR #6169, opened 2025-12-04
minify and name chunks

lkostrowski:
https://github.com/saleor/saleor-dashboard/pull/6170


lkostrowski:
Replace with minification first https://github.com/saleor/saleor-dashboard/pull/6170

manual chunks require extra testing so lets do this later

PR #6175, opened 2025-12-04
Fix external auth buttons in React Strict Mode

mirekm:


src/auth/views/Login.tsx
```diff
@@ -21,18 +21,26 @@ const LoginView = ({ params }: LoginViewProps) => {
...
+  const isExternalAuthCallback = !!(code && state && isCallbackPath);
+
+  const { data: externalAuthentications, loading: externalAuthenticationsLoading } =
+    useAvailableExternalAuthenticationsQuery({
+      skip: isExternalAuthCallback,
+    });
+
   const { lastLoginMethod, setLastLoginMethod } = useLastLoginMethod();
+  // Track if external auth callback has been processed to avoid double-processing in Strict Mode
+  const externalAuthProcessedRef = useRef(false);
```
witoszekdev, line 43: issue: I don't think we should create hacks to work around strict mode, if we have issue there it probably means we have something wrong with our implementation. It's worth checking ESLint warnings.

There's also a possibility we could introduce a bug here, as copilot wrote: https://github.com/saleor/saleor-dashboard/pull/6175/files#r2589250212

src/auth/views/Login.tsx
```diff
@@ -61,38 +69,49 @@ const LoginView = ({ params }: LoginViewProps) => {
...
-    const externalAuthParamsExist = code && state && isCallbackPath;
     const externalAuthNotPerformed = !authenticating && !errors.length;
 
-    if (externalAuthParamsExist && externalAuthNotPerformed) {
-      handleExternalAuthentication(code, state);
+    // Guard against double-processing in React Strict Mode
+    if (isExternalAuthCallback && externalAuthNotPerformed && !externalAuthProcessedRef.current) {
+      externalAuthProcessedRef.current = true;
+      handleExternalAuthentication(code!, state!);
     }
```
witoszekdev, line 106: issue: again we probably have issue somewhere else, we shouldn't write specific fixes to Strict mode

src/auth/views/Login.tsx
mirekm: As far as I see it it's not necessarily a hack for React Strict Mode per se. What should we do when component remounts and Apollo has the lazy query in flight?

mirekm: Assuming we can't have that race condition realistically outside of strict mode the ref is indeed unnecessary but what do we do then for the strict mode so it works as expected?

mirekm: The [comment](https://github.com/saleor/saleor-dashboard/pull/6175/files#r2589250212) was I think already addressed.

witoszekdev: >  What should we do when component remounts and Apollo has the lazy query in flight?

I think apollo should cancel query if it is in flight and component re-renders, if it doesn't then probably there's some bug in Apollo, it might be because we use an old version though :/

lkostrowski:
@mirekm closing for now, please reopen if you decide you want to conitnue

PR #6176, opened 2025-12-04
Skip or fix failing e2e

szczecha:
## Scope of the change

Fix a few broken tests due to recent changes in Dashboard.
Skip tests that we don't have time to fix properly - better to have a small number of working tests instead of a lot of flaky. 

Even with my best effort, it still randomly fail a few tests, but now it's 2-3, not 20+ 🫠 


playwright/pages/ordersPage.ts
```diff
@@ -36,15 +36,14 @@ export class OrdersPage extends BasePage {
   constructor(
     page: Page,
     readonly createOrderButton = page.getByTestId("create-order-button"),
-    readonly markAsPaidButton = page.getByTestId("markAsPaidButton"),
+    readonly markAsPaidButton = page.getByText("Mark as Paid"),
```
witoszekdev, line 39: suggestion: imho we should add back `data-testid` to our code, just in case we decide to change text for some reason 😄

playwright/pages/ordersPage.ts
```diff
@@ -126,8 +125,6 @@ export class OrdersPage extends BasePage {
     console.log("Navigating to order details view: " + orderLink);
     await this.page.goto(orderLink);
     await this.waitForDOMToFullyLoad();
-    await this.waitForGrid();
-    await this.waitForGrid();
```
witoszekdev, line 130: question: why was this causing errors? why did we had to wait previously for grid and now it works fine 🤔

playwright/pages/warehousePage.ts
```diff
@@ -103,7 +103,8 @@ export class WarehousePage extends BasePage {
 
   async clickDeleteWarehouseButton(warehouseName: string) {
     await this.page
-      .getByTestId(`warehouse-entry-${warehouseName}`)
+      // .getByTestId failed to locate because of spaces in warehouse name
```
witoszekdev, line 106: suggestion: I think we could fix our `data-test-id` by replacing all whitespaces (not this PR though, this is fine)

witoszekdev (approved):

PR #6178, opened 2025-12-04
Update transactions

mirekm:


src/orders/components/OrderTransaction/components/TransactionEvents/components/EventType.tsx
```diff
@@ -2,48 +2,48 @@
 import { capitalize } from "@dashboard/misc";
 import { transactionEventTypeMap } from "@dashboard/orders/messages";
 import { TransactionEventType } from "@dashboard/orders/types";
-import { makeStyles } from "@saleor/macaw-ui";
-import { Box, InfoIcon, Tooltip } from "@saleor/macaw-ui-next";
+import { Box, InfoIcon, Text, Tooltip } from "@saleor/macaw-ui-next";
```
lkostrowski: Pls import IconInfo from lucide directly, icons in macaw are deprecated

src/orders/components/OrderTransaction/components/TransactionEvents/components/EventType.tsx
```diff
@@ -2,48 +2,48 @@
...
+    <Box display="flex" alignItems="center" gap={1}>
+      <Text
+        size={2}
+        style={
+          {
+            display: "-webkit-box",
+            WebkitLineClamp: 2,
+            WebkitBoxOrient: "vertical",
+            overflow: "hidden",
+          } as React.CSSProperties
```
lkostrowski: this assertions quits typescript but is not ensuring type safety

instead

```
const textStyles: React.CSSProperties = {
  display: "-webkit-box",
  WebkitLineClamp: 2,
  WebkitBoxOrient: "vertical",
  overflow: "hidden",
}

...

style={textStyles}
```

src/orders/components/OrderTransaction/components/TransactionEvents/components/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from "./EventAvatar";
```
lkostrowski: Don't add things to index files, we are actually removing them (this is common anti pattern)

src/orders/components/OrderTransaction/components/TransactionEvents/components/PspReference.tsx
```diff
@@ -1,51 +1,26 @@
...
-          duration: theme.transitions.duration.short,
-        }),
-      },
-
-      "&&.copied": {
-        background: theme.palette.saleor.success.mid,
-        color: theme.palette.saleor.main[1],
+    pillHovered: {
+      "& $externalIcon": {
+        color: vars.colors.text.default1,
```
lkostrowski: can you use sprinkles like in other file?

src/orders/components/OrderTransaction/components/TransactionEvents/components/PspReferenceLink.tsx
```diff
@@ -2,14 +2,18 @@ import { makeStyles } from "@saleor/macaw-ui";
...
+  () => ({
     link: {
-      color: theme.palette.saleor.active[1],
+      color: "inherit",
+      textDecoration: "none",
+      "&:hover": {
+        textDecoration: "underline",
+      },
     },
   }),
```
lkostrowski: lets use sprinkles

src/orders/components/OrderTransaction/OrderTransaction.module.css
```diff
@@ -0,0 +1,13 @@
...
+  align-items: center;
+  justify-content: center;
+  transition: transform 0.2s ease;
+  transform: rotate(-90deg);
+}
+
+button[data-state="open"] .transaction-chevron,
+[data-state="open"] .transaction-chevron {
+  transform: rotate(0deg);
+}
```
lkostrowski, line 13: we should use css modules because they scope styles per-import, now you polluted a global css scope

src/orders/components/OrderTransaction/OrderTransaction.tsx
```diff
@@ -1,8 +1,11 @@
 // @ts-strict-ignore
-import { DashboardCard } from "@dashboard/components/Card";
+import "./OrderTransaction.css";
+
 import { TransactionActionEnum } from "@dashboard/graphql";
 import { TransactionFakeEvent } from "@dashboard/orders/types";
+import { Accordion, Box, ChervonDownIcon } from "@saleor/macaw-ui-next";
```
lkostrowski: use icons from lucide

src/orders/components/OrderTransaction/components/TransactionEvents/components/PspReference.tsx
mirekm: Do you mean to the ones to communicate changes?

src/orders/fixtures/OrderFixture.ts
```diff
@@ -862,6 +863,227 @@ export class OrderFixture {
...
+          id: "mock-transaction-1",
+          name: "Stripe",
+          pspReference: "pi_3ABC123",
+          externalUrl: "https://dashboard.stripe.com/payments/pi_3ABC123",
+          createdAt: "2024-12-01T10:00:00Z",
+          actions: [TransactionActionEnum.REFUND],
+          chargedAmount: { __typename: "Money", amount: 75.5, currency: "USD" },
+          refundedAmount: { __typename: "Money", amount: 25.0, currency: "USD" },
+          authorizedAmount: { __typename: "Money", amount: 0, currency: "USD" },
+          events: [
```
witoszekdev: issue: let's not create too large issue, because it can get hard to maintain them
I think we could instead add more fixtures like `OrderFixture.transaction` - e.g. different Transaction types, different events (or factories - aka functions to create such events based on parameters). Right now it's hard to read, due to all the required fields.

Example:

```ts
...this.order
transactions: [
  OrderFixture.stripeTransaction({
events: [
OrderFixture.stripeTransactionEvent.chargeSuccess({createdBy: OrderFixture.stripeApp}),
 OrderFixture.stripeTransactionEvent.refundRequest({createdBy: null}),
/* etc... */
]}),
  OrderFixture.adyenTransaction({})
]

```

src/orders/components/OrderTransaction/OrderTransaction.module.css
```diff
@@ -0,0 +1,13 @@
...
+  align-items: center;
+  justify-content: center;
+  transition: transform 0.2s ease;
+  transform: rotate(-90deg);
+}
+
+button[data-state="open"] .chevron,
+[data-state="open"] .chevron {
+  transform: rotate(0deg);
+}
```
witoszekdev, line 13: issue: I think we should instead create a new component in macaw or if it's very specific to Dashboard, a component like `<DashbordAccordeon>` with this style override.

Adding too many overrides can cause style differences really fast.

Perhaps we should replace old MaterialUI components, this way we won't need overrides

src/orders/components/OrderTransaction/components/TransactionEvents/TransactionEvents.tsx
```diff
@@ -5,7 +5,7 @@ import { TransactionFakeEvent } from "@dashboard/orders/types";
 import { TableCell, TableRow } from "@material-ui/core";
 import { makeStyles, ResponsiveTable } from "@saleor/macaw-ui";
```
witoszekdev, line 6: suggestion: Since we're already rewriting these components, maybe we could get rid of material-ui? Especially `makeStyles` to use only one way to style components (e.g. by using `<Box>` from macaw-ui-next, or by CSS modules)

src/orders/components/OrderTransaction/components/TransactionEvents/components/PspReferenceLink.tsx
```diff
@@ -1,26 +1,16 @@
...
-export const PspReferenceLink = ({ href, children }: PspRerefenceLinkProps) => {
-  const classes = useStyles();
-
+export const PspReferenceLink = ({ href, children }: PspReferenceLinkProps) => {
   if (href) {
     return (
-      <a href={href} target="_blank" rel="noopener noreferrer" className={classes.link}>
+      <a href={href} target="_blank" rel="noopener noreferrer" className={styles.link}>
         {children}
       </a>
```
witoszekdev: issue: We don't need overrides via CSS modules, this can be achieved by using `<Box as="a" __textDecoration="underline" /* ... */>`, let's avoid overwritting CSS like this when possible, this should be easy enough using macaw

src/orders/components/OrderTransaction/components/TransactionEvents/components/PspReference.tsx
```diff
@@ -56,32 +16,98 @@ interface PspReferenceProps {
...
+              <CheckIcon size={14} className={iconClassName} />
+            ) : (
+              <CopyIcon size={14} className={iconClassName} />
+            )
+          }
+          onClick={() => copy(reference)}
+          aria-label={intl.formatMessage({
+            defaultMessage: "Copy to clipboard",
+            id: "aCdAsI",
+          })}
```
witoszekdev: suggestion: Let's use existing messages in dashbaord which should be defined in other files and import them here. This will create another "Copy to clipboard" message in our translations list, we should re-use them if possible. If we don't have one we can create it, I'm sure we'll re-use it since it's a common text.

src/orders/components/OrderTransaction/components/TransactionEvents/components/PspReference.module.css
```diff
@@ -0,0 +1,17 @@
...
+}
+
+.externalIcon {
+  color: var(--mu-colors-text-default2);
+  transition: color 0.15s ease-in-out;
+}
+
+.pillHovered .externalIcon {
+  color: var(--mu-colors-text-default1);
+}
```
witoszekdev, line 17: issue: I'm pretty sure everything here (except `transition`) could be written using macaw's `<Box>` component, e.g. `<Box display="flex" alignItems="center" ... >`

This way we don't need custom styles, and that `transition` could be just an inline style.

In the future, we should probably add transitions to macaw anyway to keep them consistent.

src/orders/components/OrderTransaction/components/TransactionEvents/components/EventType.tsx
```diff
@@ -2,49 +2,47 @@
...
-      display: "flex",
-      cursor: "pointer",
-    },
-  }),
-  { name: "EventType" },
-);
+const lineClampStyles: React.CSSProperties = {
+  display: "-webkit-box",
+  WebkitLineClamp: 2,
+  WebkitBoxOrient: "vertical",
```
witoszekdev: suggestion: We should use `line-clamp` directly here and use autoprefixer which is enabled by default in Vite to add `-webkit` prefixes.

Browsers might finally add normal `line-clamp` support in the future, without doing `-webkit-box` schenanigas ;)

For this it should be used in CSS modules, not inline styles like here.

src/orders/components/OrderTransaction/components/TransactionEvents/components/EventTime.tsx
```diff
@@ -11,5 +12,9 @@ export const EventTime = ({ date }: { date: string }) => {
     minute: "2-digit",
   });
 
-  return <time dateTime={date}>{intl.format(new Date(date))}</time>;
+  return (
+    <Text size={2} whiteSpace="nowrap">
+      <time dateTime={date}>{intl.format(new Date(date))}</time>
```
witoszekdev, line 25: issue: We should check if date exists before passing it to `new Date` otherwise we might get unexpected results, like Jan 01 1970 ;)

Image

src/orders/components/OrderTransaction/components/TransactionEvents/components/EventItem.tsx
```diff
@@ -16,80 +17,62 @@ interface EventItemProps {
   event: TransactionEventFragment | TransactionFakeEvent;
   onHover: (pspReference: string) => void;
   hoveredPspReference: string;
-  hasCreatedBy: boolean;
 }
 
 const useStyles = makeStyles(
```
witoszekdev, line 22: suggestion: Since we're already rewriting this component, let's get rid of legacy Material-UI components and `makeStyles`, and use `<Box>` or CSS modules instead.

src/orders/components/OrderTransaction/components/CardTitle/CardTitle.tsx
```diff
@@ -70,88 +90,88 @@ export const OrderTransactionCardTitle = ({
...
+      label: messages.refundPending,
+      money: refundPendingAmount,
+      show: refundPendingAmount.amount > 0,
+    },
+    {
+      label: messages.cancelPending,
+      money: cancelPendingAmount,
+      show: cancelPendingAmount.amount > 0,
+    },
+  ].filter(item => item.show);
```
witoszekdev: suggestion: Let's extract this to a separate function to keep component tidy.

This should also use `useMemo` otherwise we're be iterating over array on each render.

src/orders/components/OrderTransaction/components/CardTitle/CardTitle.tsx
```diff
@@ -70,88 +90,88 @@ export const OrderTransactionCardTitle = ({
...
-        )}
-
-        {refundPendingAmount.amount > 0 && (
-          <MoneyDisplay
-            label={intl.formatMessage(messages.refundPending)}
-            money={refundPendingAmount}
-          />
-        )}
+  const isDestructiveAction = (action: TransactionActionEnum) =>
+    action === TransactionActionEnum.CANCEL || action === TransactionActionEnum.REFUND;
```
witoszekdev: issue: This should be declared outside of component, to avoid re-creating function on each render. We can move it to separate file or above this component definition.

src/orders/components/OrderTransaction/components/CardTitle/CardTitle.tsx
```diff
@@ -27,22 +26,43 @@ const TransactionTitle = ({
...
+          {intl.formatMessage(
+            {
+              defaultMessage: "Transaction #{index} on {date}",
+              id: "nYD7NT",
+            },
+            {
+              date: <EventTime date={transaction.createdAt} />,
+              index: index + 1,
+            },
+          )}
```
witoszekdev: suggestion: let's revert this change so that it's a spearate variable

src/orders/components/OrderTransaction/OrderTransaction.module.css
mirekm: I don't want to block anything with now refactoring things to Macaw tbh. But it may be a good follow up issue.

src/orders/components/OrderTransaction/components/TransactionEvents/components/EventItem.tsx
mirekm: This is a good suggestion, but it has to be another PR. That would extend the scope significantly and delay release of these changes.

src/orders/components/OrderTransaction/components/TransactionEvents/TransactionEvents.tsx
mirekm: This is a good suggestion, but it has to be another PR. That would extend the scope significantly and delay release of these changes.

src/orders/components/OrderTransaction/components/TransactionEvents/components/EventType.tsx
mirekm: I've refactored it to CSS modules but the autoprefixing didn't seem to work locally so left it with both `line-clamp` and `-webkit-line-clamp` for now.

src/orders/components/OrderTransaction/components/TransactionEvents/components/PspReference.module.css
mirekm: But what is the problem with keeping them in the module? To me that's a way cleaner way.

src/orders/components/OrderTransaction/components/TransactionEvents/components/PspReferenceLink.tsx
mirekm: But tbh I'm not sure what's the value of doing it via `<Box />` element.

mirekm (commented):
All changes pushed upstream.

src/orders/components/OrderTransaction/components/TransactionEvents/components/PspReference.module.css
witoszekdev: We can discuss this, but since we already wrote the majority of UI this way, imho it would be best to stick to one way in order to not create yet-another-way-of-styling. Previously we used `makeStyles` from material-ui (CSS-in-JS), and now we would add 3rd way: CSS modules 😅. Right now it was reserved to last-resort escape hatch to override styles.

That being said... personally I think we could just use `vanilla-extract` in dashboard as well (adding 4th way of styling 🙈), instead of CSS modules since it integrates nicely with macaw and has strict typing, unlike CSS modules where you can have typos in `var()` and it won't get detected.

`Box` component has the same nice thing as `vanilla-extract`: we have clear design tokens, and it works a bit like Tailwind CSS. In my experience, it's faster to write this way since you get autocomplete and you don't have to move between files, but that's just my opinion ;)

src/orders/components/OrderTransaction/components/TransactionEvents/components/PspReferenceLink.tsx
witoszekdev: Please see my previous comment with more reasoning, TL;DR - we already use it, it's kind of like tailwind css - so it allows you to write styles more quickly, we have definitions for our design tokens instead and using overwrites is very clear (you have `__` prefix for overrides outside of design system, the rest uses regular props).

mirekm:
All change requests addressed.

lkostrowski (approved):

PR #6179, opened 2025-12-05
Overhaul of icons

lkostrowski:
- Update lucide
- remove icons provided by macaw-next
- remove direct use of mui icons

still todo in other PRs: replace icons re-exported by legacy macaw


witoszekdev (approved):
lgtm, it would be nice if we had screenshots before & after to make sure these icons don't break layout  and have colors correctly setup 

(I had issues when replacing icons from macaw)

PR #6181, opened 2025-12-05
Upgrade prettier, fix files

witoszekdev:
- **Upgrade prettier**
- **Lock prettier version**
- **Fix files with prettier**


witoszekdev:
Due to large amount of open PRs we will merge this later to avoid conflicts, low priority.

PR #6184, opened 2025-12-05
Improve order summary

mirekm:
* Improve empty state for _Payments summary_
* Add _tabular nums_ to amounts to match decimal point display
* Fix section buttons stacking up instead of rendeing next to each other

## New
image

## Old
image
image


mirekm:
I merged this changes into the https://github.com/saleor/saleor-dashboard/pull/6189

PR #6194, opened 2025-12-10
put exact version on pr

lkostrowski:
## Scope of the change


.github/workflows/deploy-staging-and-prepare-release.yaml
```diff
@@ -151,7 +158,7 @@ jobs:
           CDN_DISTRIBUTIONS: ${{ secrets.AWS_CLOUD_STAGING_CDN_DISTRIBUTIONS }}
 
       - name: Update check
-        if: always() && ${{ steps.create-check.outputs.CHECK_ID }}
```
lkostrowski, line 154: it was syntax error

PR #6199, opened 2025-12-11
Add workflow that creates draft release on tag

lkostrowski:
## Scope of the change


src/graphql/schemaVersion.ts
```diff
@@ -14,6 +14,7 @@ export type SchemaVersion = "main" | "staging";
  * Get the current schema version based on the FF_USE_STAGING_SCHEMA feature flag
  */
 export const getSchemaVersion = (): SchemaVersion => {
+  // Flags is Vite-specific global, it's resolved build time from env variables
```
lkostrowski, line 17: added because it was confusing to me

PR #6200, opened 2025-12-12
Add new unified OrderCaptureDialog with amount options

mirekm:
## Summary

- **New unified `OrderCaptureDialog`** for both Legacy Payments and Transactions APIs
- Shows order context (total, captured so far, balance due) and transaction context (available to capture, already captured)
- Handles authorization states: full, partial, none, and fully captured
- **Outcome prediction** shows resulting order status before confirming (e.g., "This will result in a Fully captured order")
- Supports currency-aware decimal input with comma/dot normalization
- Displays server errors from both API types
- **New shared utilities** in `PriceField/utils.ts`: `normalizeDecimalSeparator`, `parseDecimalValue`, `limitDecimalPlaces`
- **Fixed `Callout` component** dark mode colors using theme-aware `getStatusColor`

## Test plan

- [ ] Legacy API: Capture works on orders with `totalAuthorized > 0`
- [ ] Transactions API: Charge action works on transactions with `authorizedAmount > 0`
- [ ] Dialog correctly shows partial authorization warning when auth < balance
- [ ] Outcome prediction updates based on selected amount
- [ ] Decimal input respects currency precision (e.g., JPY = 0 decimals)
- [ ] Error messages display correctly on failed capture

### DIFF
--- a/.changeset/breeze-sets-revives.md
+++ b/.changeset/breeze-sets-revives.md
@@ -1,0 +1,7 @@
+---
+"saleor-dashboard": minor
+---
+
+Added support for [devcontainers] to facilitate local development.
+
+[devcontainers]: https://containers.dev/
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,37 @@
+FROM mcr.microsoft.com/devcontainers/typescript-node:22 AS pnpm
+
+RUN corepack enable
+
+USER 1000:1000
+WORKDIR /app
+
+RUN corepack prepare pnpm@10 --activate
+
+FROM pnpm AS deps
+
+COPY --chown=1000:1000 ./package.json ./pnpm-lock.yaml ./.npmrc /app/
+RUN --mount=type=cache,mode=0700,uid=1000,gid=1000,target=/pnpm/store \
+		pnpm install --frozen-lockfile
+
+FROM pnpm AS source
+
+# Copy source files (ordered from least likely to be updated, up to the most likely)
+COPY --chown=1000:1000 ./assets/ assets/
+COPY --chown=1000:1000 ./locale/ locale/
+COPY --chown=1000:1000 ./scripts/ scripts/
+COPY --chown=1000:1000 .featureFlags/ .featureFlags/
+COPY --chown=1000:1000 \
+	./vite.config.js \
+	./tsconfig.json \
+	./graphql.config.ts \
+	./*.d.ts \
+	./schema-main.graphql \
+	./
+COPY src/ src/
+
+COPY --chown=1000:1000 --from=deps /app/node_modules/ ./node_modules/
+COPY --chown=1000:1000 ./package.json ./pnpm-lock.yaml /app/
+
+FROM source AS dev
+CMD ["pnpm", "exec", "vite", "--host=0.0.0.0", "--port=5005"]
+
--- a/.devcontainer/dashboard.env
+++ b/.devcontainer/dashboard.env
@@ -1,0 +1,18 @@
+# URL to the Saleor GraphQL backend, can be for example a Saleor Cloud environment,
+# a remote self-hosted env, or a local one (e.g., another docker container)
+API_URL=http://localhost:8000/graphql/
+
+# URI where the Saleor dashboard is mounted
+# This is the path where the Saleor dashboard is accessed from the web server.,
+# useful when deploying to a subdirectory, e.g., `/dashboard/`, `/admin/`
+APP_MOUNT_URI=/dashboard/
+STATIC_URL=/dashboard/
+
+# Marketplace API URL to get Saleor app listings, used to retrieve information about
+# apps from the Saleor App Marketplace.
+APPS_MARKETPLACE_API_URL=https://apps.saleor.io/api/v2/saleor-apps
+EXTENSIONS_API_URL=https://apps.saleor.io/api/v1/extensions
+
+# Language for the Dashboard (ISO 639-1 format, e.g., `EN` for English,
+# `DE` for German, `FR` for French).
+LOCALE_CODE=EN
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,12 @@
+{
+  "name": "Saleor Dashboard",
+  "dockerComposeFile": "docker-compose.yaml",
+  "service": "dashboard",
+  "workspaceFolder": "/app",
+  "forwardPorts": [5005],
+  "portsAttributes": {
+    "5005": {
+       "label": "Saleor Dashboard"
+    }
+  }
+}
--- a/.devcontainer/docker-compose.yaml
+++ b/.devcontainer/docker-compose.yaml
@@ -1,0 +1,18 @@
+services:
+  dashboard:
+    build:
+      context: ..
+      dockerfile: .devcontainer/Dockerfile
+    restart: unless-stopped
+    ports:
+      - 127.0.0.1:5005:5005
+    env_file:
+      - ./dashboard.env
+    volumes:
+      - ../src:/app/src/
+      - ../assets/:/app/assets/
+      - ../locale/:/app/locale/
+      - ../scripts/:/app/scripts/
+      - ../.featureFlags/:/app/.featureFlags/
+      - ../vite.config.js:/app/vite.config.js
+      - ../schema-main.graphql:/app/schema-main.graphql
--- a/src/components/Sidebar/MountingPoint.tsx
+++ b/src/components/Sidebar/MountingPoint.tsx
@@ -2,12 +2,14 @@ import sideBarDefaultLogoDarkMode from "@assets/images/sidebar-deafult-logo-dark
 import sideBarDefaultLogo from "@assets/images/sidebar-default-logo.png";
 import { useCloud } from "@dashboard/auth/hooks/useCloud";
 import { useLegacyThemeHandler } from "@dashboard/components/Sidebar/user/Controls";
+import { Ripple } from "@dashboard/ripples/components/Ripple";
 import { Avatar, Box, Text, Tooltip } from "@saleor/macaw-ui-next";
 import { CloudIcon } from "lucide-react";
 import { useState } from "react";
 import { FormattedMessage } from "react-intl";
 
 import { useEnvLink } from "./menu/hooks/useEnvLink";
+import { rippleCloudEnvLink } from "./ripples/cloudEnvLink";
 
 export const MountingPoint = () => {
   const { theme } = useLegacyThemeHandler();
@@ -31,33 +33,36 @@ export const MountingPoint = () => {
         Saleor Dashboard
       </Text>
       {isAuthenticatedViaCloud && (
-        <Tooltip>
-          <Tooltip.Trigger>
-            <Box
-              as="a"
-              href={envLink}
-              target="_blank"
-              rel="noopener noreferrer"
-              display="flex"
-              alignItems="center"
-              color="default2"
-              paddingRight={1}
-              style={{
-                opacity: isHovered ? 1 : 0,
-                transition: "opacity 0.15s ease-in-out",
-              }}
-              data-test-id="cloud-environment-link"
-            >
-              <CloudIcon size={16} />
-            </Box>
-          </Tooltip.Trigger>
-          <Tooltip.Content side="bottom">
-            <Tooltip.Arrow />
-            <Text size={2}>
-              <FormattedMessage defaultMessage="Go to Saleor Cloud" id="EXqb2l" />
-            </Text>
-          </Tooltip.Content>
-        </Tooltip>
+        <>
+          <Ripple model={rippleCloudEnvLink} />
+          <Tooltip>
+            <Tooltip.Trigger>
+              <Box
+                as="a"
+                href={envLink}
+                target="_blank"
+                rel="noopener noreferrer"
+                display="flex"
+                alignItems="center"
+                color="default2"
+                paddingRight={1}
+                style={{
+                  opacity: isHovered ? 1 : 0,
+                  transition: "opacity 0.15s ease-in-out",
+                }}
+                data-test-id="cloud-environment-link"
+              >
+                <CloudIcon size={16} />
+              </Box>
+            </Tooltip.Trigger>
+            <Tooltip.Content side="bottom">
+              <Tooltip.Arrow />
+              <Text size={2}>
+                <FormattedMessage defaultMessage="Go to Saleor Cloud" id="EXqb2l" />
+              </Text>
+            </Tooltip.Content>
+          </Tooltip>
+        </>
       )}
     </Box>
   );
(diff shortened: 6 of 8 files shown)


src/misc.ts
```diff
@@ -293,7 +293,16 @@ export function getMutationStatus<TData extends Record<string, SaleorMutationRes
 ): ConfirmButtonTransitionState {
   const errors = getMutationErrors(opts);
```
witoszekdev, line 294: question: could you check what `getMutationErrors` does? I think it actually checks `opts.errors`, so change here is unnecessary. If it doesn't then we probably should add error handling there

src/orders/views/OrderDetails/OrderUnconfirmedDetails/index.tsx
```diff
@@ -398,15 +435,16 @@ export const OrderUnconfirmedDetails = ({
...
         onConfirm={() => orderVoid.mutate({ id })}
       />
-      <OrderPaymentDialog
+      <OrderCaptureDialog
         confirmButtonState={orderPaymentCapture.opts.status}
-        errors={orderPaymentCapture.opts.data?.orderCapture.errors || []}
-        initial={order?.total.gross.amount}
+        errors={orderPaymentCapture.opts.data?.orderCapture?.errors ?? []}
+        orderTotal={order?.total.gross ?? { amount: 0, currency: "USD" }}
+        authorizedAmount={order?.totalAuthorized ?? { amount: 0, currency: "USD" }}
```
witoszekdev: suggestion: we should have `{amount: 0, currency: "USD"}` as variable

src/orders/views/OrderDetails/OrderUnconfirmedDetails/index.tsx
```diff
@@ -366,10 +367,46 @@ export const OrderUnconfirmedDetails = ({
...
+              amount: 0,
+              currency: "USD",
+            }
+          }
+          chargedAmount={
+            order?.transactions?.find(t => t.id === params.id)?.chargedAmount ?? {
+              amount: 0,
+              currency: "USD",
+            }
+          }
```
witoszekdev: suggestion: this should be in a `useMemo` in the component, not inlined like here

src/orders/views/OrderDetails/OrderNormalDetails/index.tsx
```diff
@@ -303,10 +304,46 @@ export const OrderNormalDetails = ({
...
+        />
+      )}
+      {/* Transaction Action Dialog - for other actions like CANCEL */}
       <OrderTransactionActionDialog
         confirmButtonState={orderTransactionAction.opts.status}
         onClose={closeModal}
-        open={params.action === "transaction-action"}
+        open={
+          params.action === "transaction-action" && params.type !== TransactionActionEnum.CHARGE
+        }
```
witoszekdev: issue: if we need to make such comparison then this new actions should have separate URL param, otherwise we can't re-hydrate state after page is refreshed

e.g. add new action `transaction-charge-action`

src/orders/components/OrderCaptureDialog/messages.ts
```diff
@@ -0,0 +1,146 @@
+import { defineMessages } from "react-intl";
+
+export const messages = defineMessages({
+  title: {
+    id: "0f6YvV",
+    defaultMessage: "Capture Payment",
```
witoszekdev, line 6: question: are we okay with using `"Payment"`? afaik everywhere else we use "Transaction", "Payment" is reserved for the legacy API.

src/orders/components/OrderCaptureDialog/messages.ts
```diff
@@ -0,0 +1,146 @@
...
+    description: "dialog title",
+  },
+  statusFullyAuthorized: {
+    id: "HwIhau",
+    defaultMessage: "Fully Authorized",
+    description: "status pill for fully authorized payment",
+  },
+  statusPartial: {
+    id: "ZUYQ+C",
+    defaultMessage: "Partial authorisation",
```
witoszekdev:
```suggestion
    defaultMessage: "Partial Authorisation",
```

src/orders/components/OrderCaptureDialog/messages.ts
```diff
@@ -0,0 +1,146 @@
...
+    description: "label for amount already captured from this transaction",
+  },
+  remainingBalance: {
+    id: "OUMqG1",
+    defaultMessage: "Remaining balance",
+    description: "label for remaining balance to capture",
+  },
+  remainingMax: {
+    id: "jhyt3I",
+    defaultMessage: "Remaining max (authorized)",
```
witoszekdev, line 66: suggestion: This is a bit confusing to me, maybe we should use "Remaining to capture"?

src/orders/components/OrderCaptureDialog/messages.ts
```diff
@@ -0,0 +1,146 @@
...
+  },
+  captureButton: {
+    id: "bRXgSC",
+    defaultMessage: "Capture {amount}",
+    description: "capture button with amount",
+  },
+  warningPartialAuthorization: {
+    id: "8JEG80",
+    defaultMessage:
+      "The remaining authorization doesn't cover the balance. {shortfall} will need a separate payment.",
```
witoszekdev, line 102: question: what is remaining authorization?

src/orders/components/OrderCaptureDialog/messages.ts
```diff
@@ -0,0 +1,146 @@
...
+    description: "warning when authorized is less than total",
+  },
+  errorNoAuthorization: {
+    id: "SnV3LR",
+    defaultMessage:
+      "No payment has been authorized for this order. The full amount of {amount} cannot be captured.",
```
witoszekdev, line 108: question: will this be ever used? if we don't have transactions on order, then we don't show the button to capture...

src/orders/components/OrderCaptureDialog/messages.ts
```diff
@@ -0,0 +1,146 @@
...
+  outcomePartiallyCharged: {
+    id: "ycg2RR",
+    defaultMessage: "This will result in a {status} order",
+    description: "outcome prediction when order will be partially charged",
+  },
+  outcomeOvercharged: {
+    id: "DXaxpH",
+    defaultMessage: "This will result in an {status} order",
+    description: "outcome prediction when order will be overcharged",
+  },
```
witoszekdev: issue: these strings are the same, we should either:

1. remove `{status}` and replace it with the status we want to display
2. keep status, but add separate translation strings for each status

src/orders/components/OrderCaptureDialog/messages.ts
```diff
@@ -0,0 +1,146 @@
...
+  statusFullyCapturedPill: {
+    id: "G9y5Ze",
+    defaultMessage: "Fully captured",
```
witoszekdev, line 118:
```suggestion
    defaultMessage: "Fully Captured",
```

src/extensions/views/AddCustomExtension/AddCustomExtension.test.tsx
```diff
@@ -8,6 +9,10 @@ import { usePermissions } from "./hooks/usePermissions";
 import { useUserAppCreationPermissions } from "./hooks/useUserAppCreationPermissions";
 import { useUserPermissionSet } from "./hooks/useUserPermissionMap";
 
+const Wrapper = ({ children }: React.PropsWithChildren<{}>) => (
+  <ThemeProvider>{children}</ThemeProvider>
+);
```
witoszekdev: question: why do we need theme provider here?

If some components do need it for tests to run, we should rather mock their dependencies - this makes tests faster to run

src/components/PriceField/utils.ts
```diff
@@ -32,3 +32,40 @@ export const getCurrencyDecimalPoints = (currency?: string) => {
 
 export const findPriceSeparator = (input: string) =>
   SEPARATOR_CHARACTERS.find(separator => input.includes(separator));
+
+/**
+ * Normalizes decimal separator to JavaScript standard (dot).
+ * Converts comma to dot for locales that use comma as decimal separator.
+ */
+export const normalizeDecimalSeparator = (value: string): string => value.replace(",", ".");
```
witoszekdev: question: why do we want to override this if localization uses its own syntax for separation? we should probably leave it up to the localization used by the user

witoszekdev (changes requested):

lkostrowski:
Please change target to main / merge unify-order-values first

src/misc.ts
mirekm: https://saleorcommerce.slack.com/archives/C06DU38BB39/p1765631446124399

mirekm: Currently the Apollo-level errors are ignored. There's a toast message that shows the Apollo-level error but the dialog would assume that it was a successful operation as of now. This fixes it system wide.

src/orders/components/OrderCaptureDialog/messages.ts
mirekm: I've added it as I think the modal should be able to show that state. But as of now it's correct, our UI does not allow intentionally showing that state. I would still leave it as part of this component.

src/orders/components/OrderCaptureDialog/messages.ts
mirekm: I think we should aim to unify that as from the user perspective the underlying API should make no difference. The action is initialised already from the Transactions section which is the section covering the new API. The dialog itself should be unified in terms of naming.

src/orders/components/OrderCaptureDialog/messages.ts
mirekm: We don't use Title Case in most of the UI as far as I can tell, correct?

src/extensions/views/AddCustomExtension/AddCustomExtension.test.tsx
mirekm: It seems like an established pattern in the codeboase to use ThemeProvider. For that reason for this PR, I'd suggest keeping it as-is.

src/components/PriceField/utils.ts
mirekm: Treating both as decimal separators is reasonable I think because:

* Users don't type thousands separators - they're for display, not input. People type 1234.56, not 1,234.56.
* Price amounts are usually unambiguous - 10,50 clearly means "ten and a half" (no one writes one thousand fifty as 10,50)
* Large amounts are rare edge cases - how often does someone type 1,234 in a price field where the ambiguity matters?

src/components/PriceField/utils.ts
```diff
@@ -32,3 +32,63 @@ export const getCurrencyDecimalPoints = (currency?: string) => {
...
   SEPARATOR_CHARACTERS.find(separator => input.includes(separator));
+
+/**
+ * Normalizes decimal separator to JavaScript standard (dot).
+ * Handles different locale formats:
+ * - European: "1.234,56" → "1234.56" (comma decimal, dot thousand)
+ * - US: "1,234.56" → "1234.56" (dot decimal, comma thousand)
+ * - Simple: "10,50" or "10.50" → "10.50"
+ */
+export const normalizeDecimalSeparator = (value: string): string => {
```
lkostrowski, line 45: why do we need such logic on the frontend?

src/orders/components/OrderCaptureDialog/OrderCaptureDialog.tsx
```diff
@@ -0,0 +1,552 @@
...
+      // Default to max capturable (remaining auth)
+      return String(availableToCapture);
+    }
+
+    // Default to remaining amount to pay
+    return String(remainingToPay);
+  };
+
+  const [selectedOption, setSelectedOption] = useState<CaptureAmountOption>(getDefaultOption);
+  const [customAmount, setCustomAmount] = useState<string>(getDefaultCustomAmount);
```
lkostrowski: It looks like you set in state localized values instead of decimals, then you add money-parsing logic on frontend to make it decimal again

src/orders/components/OrderSummary/LegacyPaymentsApiButtons.tsx
```diff
@@ -26,7 +26,7 @@ export const LegacyPaymentsApiButtons = ({
   onLegacyPaymentsApiCapture,
   onLegacyPaymentsApiRefund,
   onLegacyPaymentsApiVoid,
-}: Props) => {
```
lkostrowski, line 29: Probably it should return ReactNode

src/orders/views/OrderDetails/OrderNormalDetails/index.tsx
```diff
@@ -174,6 +175,12 @@ export const OrderNormalDetails = ({
 
   const errors = orderUpdate.opts.data?.orderUpdate.errors || [];
 
+  const defaultZeroMoney = { amount: 0, currency: "USD" };
```
lkostrowski: do you need to provide fallback/default value? It's dangerous, not only code is harder to debug, but also it put actually correct business value (0) instead of programatically correct `null` vaue

src/components/PriceField/utils.ts
mirekm: I've introduced it in order to make the input more user friendly. People don't use thousands separator when entering data, they only separate decimal part (one way or the other). Thousands separator is mainly used for displaying purposes. So, this util accepts both comma and dot as a decimal separator - very handy from the UX perspective.

src/orders/components/OrderCaptureDialog/OrderCaptureDialog.tsx
mirekm: Yes, numbers from API can be safely stored as numbers. Fixing.

src/orders/components/OrderSummary/LegacyPaymentsApiButtons.tsx
mirekm: Fixed.

src/orders/views/OrderDetails/OrderNormalDetails/index.tsx
mirekm: Thanks, multiple problems with that one including currency. Fixing.

mirekm: If `order` or `selectedTransaction` is missing, the dialog simply won't render (instead of showing incorrect $0 USD).

src/components/PriceField/utils.ts
lkostrowski: Please check if this library will not do the job
https://currency.js.org/

Its already used in the dashboard and probably cover all cases

mirekm: `currency.js` solves a bit different problem as far as I can see (precision in arithmetic). It takes the decimal separator in parameters upfront `{ decimal: "," }` so it doesn't really solve my case.

lkostrowski (approved):

src/components/PriceField/utils.ts
mirekm: Also discussed with @lkostrowski

PR #6213, opened 2025-12-16
Fix customer->order filter redirection

lkostrowski:
Fixed redirection between Customer Details -> See all orders to Orders List (which this customer selected). Now filter is properly set on the URL and only relevant orders are displayed

### DIFF
--- a/src/orders/urls.ts
+++ b/src/orders/urls.ts
@@ -78,6 +78,11 @@ export type OrderListUrlQueryParams = BulkAction &
   OrderListUrlSort &
   Pagination &
   ActiveTab;
+
+/**
+ * @deprecated
+ * This helper is likely broken, at least filters don't work. Either construct url manually or fix it
+ */
 export const orderListUrl = (params?: OrderListUrlQueryParams): string => {
   const orderList = orderListPath;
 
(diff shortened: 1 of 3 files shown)


src/orders/urls.ts
```diff
@@ -78,6 +78,11 @@ export type OrderListUrlQueryParams = BulkAction &
   OrderListUrlSort &
   Pagination &
   ActiveTab;
+
+/**
+ * @deprecated
+ * This helper is likely broken, at least filters don't work. Either construct url manually or fix it
+ */
```
witoszekdev, line 85: it's not deprecated it's just that customer email filtering requires us to build filters from `ConditionalFilter` to create proper URL query

but we can still use that function to open modals, set active tab, change sorting, etc.: see. `OrderListUrlQueryParams`

witoszekdev (approved):

src/orders/urls.ts
lkostrowski: I will keep it deprecated as long as we fix it. For now, it accepts "customer" but fails to render it. We should eventualyl fix that, but I'd rather be informed that I try to use something broken, even partially

PR #6226, opened 2025-12-17
Link to the variant's metadata instead of letting users change it in order line metadata

mirekm:


src/components/Metadata/MetadataCard.tsx
```diff
@@ -25,9 +26,11 @@ export const MetadataCard = ({
   readonly = false,
   disabled,
   error,
+  defaultExpanded,
 }: MetadataCardProps) => {
   const intl = useIntl();
-  const [expanded, setExpanded] = useState(readonly ? "metadata-accordion" : undefined);
+  const isExpanded = defaultExpanded ?? readonly;
+  const [expanded, setExpanded] = useState(isExpanded ? "metadata-accordion" : undefined);
```
lkostrowski: this looks strange, first you duplicate some interim state and create another state from it, but in react "expanded" field will be created with first render/mount, when your props (defaultExpanded / readonly) change, it will not reflect that (should it?)

if this is expected, at least do not create 2 "identitcal" variables like "expanded" and "isExpanded"

you can call it "initiallyExpanded" and "expanded"

lkostrowski: why do you store string in state? can't it be `true` / `false`?

mirekm: I refactored it also because it conflates two different concerns: `isExpanded` and `readonly`.

mirekm: It's driven by how the `Accordion` component uses string identifiers to track which items are expanded.

src/components/Metadata/MetadataCard.tsx
```diff
@@ -16,24 +16,29 @@ export interface MetadataCardProps {
...
+const ACCORDION_VALUE = "metadata-accordion";
+
 export const MetadataCard = ({
   data,
   isPrivate,
   onChange,
   readonly = false,
   disabled,
   error,
+  defaultExpanded,
```
witoszekdev, line 31: nitpick: 
```suggestion
  defaultExpanded = false,
```

src/components/Metadata/MetadataCard.tsx
```diff
@@ -16,24 +16,29 @@ export interface MetadataCardProps {
...
 }: MetadataCardProps) => {
   const intl = useIntl();
-  const [expanded, setExpanded] = useState(readonly ? "metadata-accordion" : undefined);
+  const initiallyExpanded = defaultExpanded ?? false;
```
witoszekdev, line 34: nitpick: we can get rid of this variable if we add `defaultExpanded = false`

```suggestion
```

lkostrowski (approved):

PR #6207, opened 2025-12-15
Replace custom toasts with sonner

mirekm:
## Scope
* Replaces Saleor Dashboard toasts design and behaviour globally
* Drops previous `apiMessage` handling via _See more_ (we have to improve the error massages, as logs are not helping any of the Dashboard users)
* Falls back to _apiMessage_ if there's no description
* Respects Saleor's Dark and Light themes

PR #6241, opened 2025-12-26
Introduce devtools panel for debugging and testing purposes

mirekm:
A developer tools panel that allows registering debug tools from anywhere in the app.

For now it includes a notifications debug panel for testing toast messages (trigger different notification types, test auto-dismiss behavior, simulate long descriptions, etc.). Useful for development and QA.

image

### DIFF
--- a/src/components/notifications/NotificationProvider.test.tsx
+++ b/src/components/notifications/NotificationProvider.test.tsx
@@ -2,6 +2,7 @@ import { ThemeProvider } from "@saleor/macaw-ui-next";
 import { renderHook } from "@testing-library/react-hooks";
 import { PropsWithChildren, useContext } from "react";
 
+import { DevToolsProvider } from "../DevTools";
 import { NotificationContext } from ".";
 import NotificationProvider from "./NotificationProvider";
 
@@ -17,7 +18,9 @@ jest.mock("sonner", () => ({
 describe("NotificationProvider", () => {
   const wrapper = ({ children }: PropsWithChildren) => (
     <ThemeProvider>
-      <NotificationProvider>{children}</NotificationProvider>
+      <DevToolsProvider>
+        <NotificationProvider>{children}</NotificationProvider>
+      </DevToolsProvider>
     </ThemeProvider>
   );
 


.changeset/mighty-toes-march.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Introduce new, unified order capture dialog
```
witoszekdev, line 5: suggestion: let's add more description here what's new about this capture dialog


```suggestion
Introduced new, unified order capture dialog
```

src/components/Callout/messages.ts
```diff
@@ -1,14 +1,7 @@
...
-    description: "callout, title",
-    id: "BnB/7Y",
-  },
-  warning: {
-    defaultMessage: "Warning",
-    description: "callout, title",
-    id: "UUVUyy",
-  },
-});
+// Re-export common messages for backward compatibility
```
witoszekdev, line 3: nitpick: let's add a todo to remove this altogether

```suggestion
// Re-export common messages for backward compatibility
// TODO: Remove and use `commonMessages` from @dashboard/intl
```

src/auth/utils.ts
```diff
@@ -60,7 +60,7 @@ export const handleNestedMutationErrors = ({
 }: {
   data: any;
   intl: IntlShape;
-  notify: (message: IMessage) => void;
+  notify: (notification: INotification) => void;
```
witoszekdev, line 63: question: Shouldn't we use `INotificationCallback` here? isn't it the same thing?


```suggestion
  notify: INotificationCallback;
```

src/components/DevTools/DevToolsPanel.tsx
witoszekdev, line 1: question: Do we really need this? I think we can test notifications inside Storybook instead 😄 

This will be shipped in our production code and it feels like adding a Conami code, rather than proper devtools 😅

src/components/notifications/index.ts
```diff
@@ -0,0 +1,29 @@
+import { createContext } from "react";
+
+type Status = "success" | "error" | "info" | "warning";
+
+export interface INotification {
+  actionBtn?: {
```
witoszekdev, line 6: nitpick: imho it's better to use verbose names whenever possible

```suggestion
  actionButton?: {
```

src/components/notifications/index.ts
```diff
@@ -0,0 +1,29 @@
+import { createContext } from "react";
+
+type Status = "success" | "error" | "info" | "warning";
```
witoszekdev, line 3: suggestion: This is not exported, but I think having a full description what status is this would be better, because it's also visible in language server (as type completion in editor)

```suggestion
type NotificationStatus = "success" | "error" | "info" | "warning";
```

src/components/notifications/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  };
+  autohide?: number;
+  title?: string;
+  text?: React.ReactNode;
+  status?: Status;
+  apiMessage?: string;
+}
+
+export interface INotificationContext {
+  show: (notification: INotification, timeout?: number | null) => void;
```
witoszekdev, line 18: question: shouldn't this function return `notificationId` so that it can be later removed using `remove` method?

src/components/notifications/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  remove: (notificationId: number) => void;
+  clearErrorNotifications: () => void;
+}
+
+export type INotificationCallback = (notification: INotification) => void;
+export const NotificationContext = createContext<INotificationContext | null>(null);
```
witoszekdev, line 24: issue: Since we're already using jotai for global state management, we should stick with it, also for performance reasons because changes in context cause each component that has `useContext` to rerender, it also helps us avoid this:

```
    <ThemeProvider>
      <DevToolsProvider>
        <NotificationProvider>{children}</NotificationProvider>
```

https://jotai.org/docs/core/atom

src/components/notifications/NotificationsDebugPanel.tsx
witoszekdev, line 1: issue: again we're shipping a lot of code to prod just for us to debug notifications :/

src/components/notifications/Toast.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+      expect(screen.queryByText("description")).not.toBeInTheDocument();
+    });
+
+    it("renders action button when action is provided", () => {
+      // Arrange & Act
+      renderToast({
+        action: {
+          label: "Undo",
+          onClick: jest.fn(),
+        },
```
witoszekdev, line 68: suggestion: maybe we should rethink this API so that we can provide multiple actions?

src/components/notifications/Toast.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+      ["success", "Success"],
+      ["error", "Error"],
+      ["warning", "Warning"],
+      ["info", "Info"],
+    ] as const)("renders %s toast with title", (type, title) => {
+      // Arrange & Act
+      renderToast({ type, title });
+
+      // Assert
+      expect(screen.getByText(title)).toBeInTheDocument();
```
witoszekdev, line 95: nitpick: this test doesn't actually test each toast type, in Storybook we could have defined each toast and have snapshots for them to see if they render properly

src/components/notifications/Toast.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+const defaultProps: ToastProps = {
+  id: "test-toast-1",
+  type: "info",
+  title: "Test Title",
+};
+
+const renderToast = (props: Partial<ToastProps> = {}) =>
+  render(
+    <ThemeProvider>
```
witoszekdev, line 23: suggestion: we could replace ThemeProvider with `jest.mock` - since Dashboard has _a lot_ of tests it will speed things up, since we don't need to setup the actual theme provider.

afaik we only need to mock this to get the same behavior for themes:

```ts
jest.mock("@saleor/macaw-ui-next", () => ({
  useTheme: () => ({ theme: "defaultLight" }),
}));
```

for old macaw we can use:

```ts
jest.mock("@saleor/macaw-ui", () => ({
  useTheme: jest.fn(() => () => ({})),
  useStyles: jest.fn(() => () => ({})),
  makeStyles: jest.fn(() => () => ({})),
}));

```

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+  const containerRef = useRef<HTMLDivElement>(null);
+  const textRef = useRef<HTMLSpanElement>(null);
+
+  const isDark = theme === "defaultDark";
+
+  const backgroundColor = useMemo(() => {
+    if (type === "success") {
+      return isDark ? "oklch(25% 0.04 145)" : "oklch(97% 0.04 145)";
+    }
+
```
witoszekdev, line 54: suggestion: instead of using magic values here, let's add `darkColor` or something like this to our `TOAST_STYLES`

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+}
+
+export const Toast = ({ id, type, title, description, action }: ToastProps) => {
+  const { theme, themeValues } = useTheme();
+  const { iconColor, Icon } = TOAST_STYLES[type];
+
+  const [isExpanded, setIsExpanded] = useState(false);
+  const [isTruncated, setIsTruncated] = useState(false);
+  const [isHovering, setIsHovering] = useState(false);
+  const [isToggleHovering, setIsToggleHovering] = useState(false);
```
witoszekdev, line 43: issue: This is a primitive obsession antipattern, we're having a lot of boolean values for managing state and a bunch of handlers that interact with them, which can produce invalid state.

We should refactor this by doing some of these:

1. separate components (e.g. expand button, truncated text, etc.) and:
  a) a separate hook for all the state so that our UI is purely for rendering UI, not calculating state - this would be easiest to change and makes this much easier to test
  b) jotai atoms that are imported in each separate component
2. using `useReducer` to change the state (might be an overkill for this specific component, hard to tell from my perspective)

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+            setIsExpanded(false);
+          }
+        }
+      }
+    });
+
+    observer.observe(toastWrapper, { attributes: true });
+
+    return () => observer.disconnect();
+  }, []);
```
witoszekdev, line 119: suggestion: this could be its own hook to make this component more readable

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+      borderRadius={3}
+      boxShadow="defaultModal"
+      borderColor="default1"
+      borderWidth={1}
+      borderStyle="solid"
+      __backgroundColor={backgroundColor}
+      __width="380px"
+      __minHeight="94px"
+    >
+      {/* Icon */}
```
witoszekdev, line 138: suggestion: we don't need such comments, it's already very readable just by looking at the code itself

```suggestion
```

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+      <Box color={iconColor} __lineHeight="0" flexShrink="0" paddingTop={0.5}>
+        <Icon size={20} strokeWidth={2} />
+      </Box>
+
+      {/* Content */}
```
witoszekdev, line 143:
```suggestion
```

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+                  display: "block",
+                }}
+              >
+                <Text size={2} color="default2">
+                  {isExpanded ? "Show less" : "Show more"}
+                </Text>
+              </Box>
+            )}
+          </Box>
+        )}
```
witoszekdev, line 219: to my previous comments: by making this separate component we would only need to expose `isExpanded` and `handleExpand` as props, the rest would be encapsulated in the toggle itself

src/components/notifications/Toast.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+      );
+
+      // Act
+      renderToast({ description: jsxDescription });
+
+      // Assert
+      expect(screen.getByText("Bold")).toBeInTheDocument();
+      expect(screen.getByText("italic")).toBeInTheDocument();
+    });
+  });
```
witoszekdev, line 225: suggestion: we don't really need this test it's basically testing if React works correctly 😄 




```suggestion
```

src/components/notifications/Toast.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+    it("responds to mouse enter events on description container", async () => {
+      // Arrange
+      const description = "A description that can be hovered";
+
+      renderToast({ description });
+
+      // Act
+      const descriptionElement = screen.getByText(description);
+      const descriptionContainer = descriptionElement.closest('[class*="_18fs8ps"]');
```
witoszekdev, line 195: issue: what's this magic class id? is this from external library or our own styles? this could make the test fail if we change styles, so we should change this selector

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+    handleDismiss();
+  }, [action, handleDismiss]);
+
+  const handleMouseEnter = useCallback(() => setIsHovering(true), []);
+  const handleMouseLeave = useCallback(() => setIsHovering(false), []);
+  const handleToggleMouseEnter = useCallback(() => setIsToggleHovering(true), []);
+  const handleToggleMouseLeave = useCallback(() => setIsToggleHovering(false), []);
+
+  // Check if description text is truncated
+  useEffect(() => {
```
witoszekdev, line 84: issue: this should use `useLayoutEffect` - this way we won't even need `requestAnimationFrame`. Without this, we can have FLUC (flash of unstyled content)

Example:

```ts
useLayoutEffect(() => {
  const textEl = textRef.current;
  if (!textEl) return;

  const isOverflowing = textEl.scrollHeight > textEl.clientHeight + 1;
  if (isOverflowing) {
    setIsTruncated(true);
  }
}, [description]);
```

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+  const handleActionClick = useCallback(() => {
+    action?.onClick();
+    handleDismiss();
+  }, [action, handleDismiss]);
+
+  const handleMouseEnter = useCallback(() => setIsHovering(true), []);
+  const handleMouseLeave = useCallback(() => setIsHovering(false), []);
+  const handleToggleMouseEnter = useCallback(() => setIsToggleHovering(true), []);
+  const handleToggleMouseLeave = useCallback(() => setIsToggleHovering(false), []);
```
witoszekdev, line 81: issue: we don't need to use `useCallback` here, since these methods already don't have any dependencies

this way we can make it easier to read in our component, e.g.:

```
onMouseEnter={() => setIsHovering(true)}
```

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+  const handleToggleExpand = useCallback((e: React.MouseEvent) => {
+    e.stopPropagation();
+    setIsExpanded(prev => !prev);
+  }, []);
+
+  const handleActionClick = useCallback(() => {
+    action?.onClick();
+    handleDismiss();
+  }, [action, handleDismiss]);
```
witoszekdev, line 76: same here let's drop useCallback, since we don't get any optimization from this

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+      if (isOverflowing) {
+        setIsTruncated(true);
+      }
+    });
+  }, [description]);
+
+  // Auto-collapse text when Sonner stack collapses (peek mode)
+  useEffect(() => {
+    const toastWrapper = containerRef.current?.closest("[data-sonner-toast]");
```
witoszekdev, line 100: question: won't this break if `sonner` changes it's internals? could we use different selector instead? we might check the DOM tree but maybe it's just first child from our `containerRef`?

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+                onClick={handleToggleExpand}
+                onMouseEnter={handleToggleMouseEnter}
+                onMouseLeave={handleToggleMouseLeave}
+                style={{
+                  cursor: "pointer",
+                  border: "none",
+                  background: "none",
+                  padding: 0,
+                  marginTop: "4px",
+                  outline: "none",
```
witoszekdev, line 207: suggestion: let's move these styles to `Box` component props, rather than `style`

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+                  opacity: isToggleHovering ? 1 : 0.7,
+                  transition: `opacity ${ANIMATION_TIMING}`,
+                  display: "block",
+                }}
+              >
+                <Text size={2} color="default2">
+                  {isExpanded ? "Show less" : "Show more"}
```
witoszekdev, line 214: issue: Missing translations, we should use `<FormattedMessage />` or `intl.formatMessage`

src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,253 @@
...
+        borderRadius={2}
+        display="flex"
+        alignItems="center"
+        justifyContent="center"
+        color="default2"
+        borderWidth={0}
+        __backgroundColor="transparent"
+        style={{
+          outline: "none",
+          opacity: 0.6,
```
witoszekdev, line 245: suggestion: we should move these to `Box` component props (either `outline="none"` or `__outline="none"`)

src/components/PriceField/utils.ts
```diff
@@ -32,3 +32,40 @@ export const getCurrencyDecimalPoints = (currency?: string) => {
...
+  if (maxDecimalPlaces === 0) {
+    return integerPart;
+  }
+
+  if (decimalPart.length > maxDecimalPlaces) {
+    return `${integerPart}${separator}${decimalPart.slice(0, maxDecimalPlaces)}`;
+  }
+
+  return value;
+};
```
witoszekdev, line 71: suggestion: this could use a simple `step=".01"` on the HTML input element, no need to run JavaScript to limit user input and it's more accessible

We also already have `PriceInputField` component so maybe we could re-use it?

src/components/PriceField/utils.ts
```diff
@@ -32,3 +32,40 @@ export const getCurrencyDecimalPoints = (currency?: string) => {
 
 export const findPriceSeparator = (input: string) =>
   SEPARATOR_CHARACTERS.find(separator => input.includes(separator));
+
+/**
+ * Normalizes decimal separator to JavaScript standard (dot).
+ * Converts comma to dot for locales that use comma as decimal separator.
+ */
+export const normalizeDecimalSeparator = (value: string): string => value.replace(",", ".");
```
witoszekdev, line 40: issue: This will replace only first dot, some locales might use formatting like this: `1,234,567` this code will replace it to `1.234,567`

src/components/PriceField/utils.ts
```diff
@@ -32,3 +32,40 @@ export const getCurrencyDecimalPoints = (currency?: string) => {
...
+
+/**
+ * Parses a decimal string value to a number, handling locale-specific separators.
+ * Returns 0 if the value cannot be parsed.
+ */
+export const parseDecimalValue = (value: string): number =>
+  parseFloat(normalizeDecimalSeparator(value)) || 0;
+
```
witoszekdev, line 48: question: I'm wondering why do we even need this? Is this for `input`? If so then browsers already handle different locales via `valueAsNumber`: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/valueAsNumber

src/extensions/views/AddCustomExtension/AddCustomExtension.test.tsx
```diff
@@ -8,6 +9,10 @@ import { usePermissions } from "./hooks/usePermissions";
 import { useUserAppCreationPermissions } from "./hooks/useUserAppCreationPermissions";
 import { useUserPermissionSet } from "./hooks/useUserPermissionMap";
 
+const Wrapper = ({ children }: React.PropsWithChildren<{}>) => (
+  <ThemeProvider>{children}</ThemeProvider>
```
witoszekdev, line 13: suggestion: let's replace wrapper with `jest.mock` for better speed (see me prev comments)

src/hooks/useNotifier/useNotifier.tsx
```diff
@@ -0,0 +1,70 @@
...
+        },
+      );
+    },
+    [intl],
+  );
+
+  return notify;
+}
+
+export default useNotifier;
```
witoszekdev, line 70: suggestion: let's use named exports instead; if this is for compatibility reasons let's add a comment why we need this

src/orders/views/OrderDetails/OrderUnconfirmedDetails/index.tsx
```diff
@@ -168,6 +169,11 @@ export const OrderUnconfirmedDetails = ({
   const intl = useIntl();
   const [transactionReference, setTransactionReference] = useState("");
   const errors = orderUpdate.opts.data?.orderUpdate.errors || [];
+  const defaultZeroMoney = { amount: 0, currency: "USD" };
```
witoszekdev, line 172: question: is this correct? what if order has different currency?

.changeset/whole-groups-happen.md
witoszekdev, line 1: please let's add this as a separate PR if we want this 🙏🏻 

this makes such PR harder to review because we have a bunch of different changes - in notifications, in order capture dialog

.changeset/mighty-toes-march.md
witoszekdev, line 1: afaik we are also missing changelog for new notifications?

src/orders/components/OrderCaptureDialog/OrderCaptureDialog.tsx
```diff
@@ -0,0 +1,555 @@
...
+  confirmButtonState: ConfirmButtonTransitionState;
+  orderTotal: IMoney;
+  authorizedAmount: IMoney;
+  /** Amount already charged/captured - used for display */
+  chargedAmount?: IMoney;
+  /**
+   * Order's total balance (for multi-transaction orders).
+   * Negative = customer owes money, Positive = overpaid.
+   * When provided, used instead of (orderTotal - chargedAmount) for remaining calculation.
+   */
```
witoszekdev, line 48: I don't understand why is this optional? When we should provide this and when not? Maybe we should create two separate components for different flows (can be internally inside `OrderCaptureDialog` since for me it looks like we're mixing 2 different use-cases

src/orders/components/OrderCaptureDialog/OrderCaptureDialog.tsx
```diff
@@ -0,0 +1,555 @@
...
+
+  const handleSubmit = (): void => {
+    if (canSubmit) {
+      onSubmit(selectedAmount);
+    }
+  };
+
+  const formatMoney = (amount: number): JSX.Element => (
+    <span style={{ fontVariantNumeric: "tabular-nums" }}>
+      <Money money={{ amount, currency }} />
```
witoszekdev, line 187: suggestion: let's add option to `Money` to have different numeric value options instead of wrapping it with yet another span, this makes it harder to read, than just `<Money ... />` in a component

src/orders/components/OrderCaptureDialog/OrderCaptureDialog.tsx
```diff
@@ -0,0 +1,555 @@
...
+  const customAmountValue = parseDecimalValue(customAmount);
+  const isCustomAmountInRange = customAmountValue > 0 && customAmountValue <= maxCapturable;
+  const isCustomAmountValid = selectedOption !== "custom" || isCustomAmountInRange;
+  const showCustomAmountError =
+    selectedOption === "custom" &&
+    authStatus !== "none" &&
+    authStatus !== "charged" &&
+    !isCustomAmountInRange;
+  const canSubmit =
+    authStatus !== "none" && authStatus !== "charged" && isCustomAmountValid && selectedAmount > 0;
```
witoszekdev, line 177: issue: This is very hard to read what's going on in this component, we should extract this to a separate hook, example:

```ts
const captureState = useMemo(() => {
  const availableToCapture = authorizedAmount.amount;
  const alreadyCharged = chargedAmount?.amount ?? 0;
  const remainingToPay = orderBalance 
    ? Math.max(0, -orderBalance.amount) 
    : orderTotal.amount - alreadyCharged;
  
  const status: AuthorizationStatus = 
    remainingToPay <= 0 ? "charged" :
    availableToCapture <= 0 ? "none" :
    availableToCapture >= remainingToPay ? "full" : "partial";
  
  return {
    availableToCapture,
    alreadyCharged,
    remainingToPay,
    status,
    maxCapturable: Math.max(0, availableToCapture),
    shortfall: remainingToPay - availableToCapture,
    canCaptureOrderTotal: availableToCapture >= remainingToPay && remainingToPay > 0,
    orderTotalCaptured: orderTotal.amount - remainingToPay,
  };
}, [orderTotal, authorizedAmount, chargedAmount, orderBalance]);
```

src/orders/components/OrderCaptureDialog/OrderCaptureDialog.tsx
```diff
@@ -0,0 +1,555 @@
...
+  useEffect(() => {
+    if (open) {
+      setSelectedOption(getDefaultOption());
+      setCustomAmount(getDefaultCustomAmount());
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [open]);
+
+  // Get max decimal places for this currency (e.g., 2 for USD, 0 for JPY, 3 for KWD)
+  const maxDecimalPlaces = useMemo(() => getCurrencyDecimalPoints(currency), [currency]);
```
witoszekdev, line 149: afaik this is already handled in Dashboard by existing components for entering money, let's check this so that we don't reapeat ourselves

src/orders/components/OrderCaptureDialog/OrderCaptureDialog.tsx
```diff
@@ -0,0 +1,555 @@
...
+    }
+
+    if (authStatus === "partial") {
+      // Default to max capturable (remaining auth)
+      return String(availableToCapture);
+    }
+
+    // Default to remaining amount to pay
+    return String(remainingToPay);
+  };
```
witoszekdev, line 134: suggestion: these should be separate components to make it easier to read

src/orders/components/OrderCaptureDialog/OrderCaptureDialog.tsx
```diff
@@ -0,0 +1,555 @@
...
+    if (availableToCapture <= 0) {
+      return "none";
+    }
+
+    if (availableToCapture >= remainingToPay) {
+      return "full";
+    }
+
+    return "partial";
+  };
```
witoszekdev, line 108: suggestion: instead of creating a function which gets it's own closure from our state anyway we should re-write this to a hook

src/orders/components/OrderCaptureDialog/OrderCaptureDialog.tsx
```diff
@@ -0,0 +1,555 @@
...
+  const [selectedOption, setSelectedOption] = useState<CaptureAmountOption>(getDefaultOption);
+  const [customAmount, setCustomAmount] = useState<string>(getDefaultCustomAmount);
+
+  // Reset state when dialog opens to ensure correct defaults based on current props
+  useEffect(() => {
+    if (open) {
+      setSelectedOption(getDefaultOption());
+      setCustomAmount(getDefaultCustomAmount());
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
witoszekdev, line 145: issue: this warning is important here, these functions have stale closure when we call them in useEffect which means they will reset state to invalid values, instead of using useEffect we can rely on React to do cleanup for us, if we render this component conditionally wherever is used:

```ts
{params.action === "transaction-charge-action" && (
  <OrderCaptureDialog ... />
)}
```

and we can remove `open` as prop.

src/orders/views/OrderDetails/OrderNormalDetails/index.tsx
```diff
@@ -303,6 +312,40 @@ export const OrderNormalDetails = ({
...
+          }
+          chargedAmount={
+            order?.transactions?.find(t => t.id === params.id)?.chargedAmount ?? {
+              amount: 0,
+              currency: "USD",
+            }
+          }
+          orderBalance={order?.totalBalance ?? { amount: 0, currency: "USD" }}
+          isTransaction
+          open={true}
```
witoszekdev, line 335: we should remove `open` since we're already conditionally rendering this component

src/orders/views/OrderDetails/OrderNormalDetails/index.tsx
```diff
@@ -354,15 +397,16 @@ export const OrderNormalDetails = ({
...
       />
-      <OrderPaymentDialog
+      <OrderCaptureDialog
         confirmButtonState={orderPaymentCapture.opts.status}
-        errors={orderPaymentCapture.opts.data?.orderCapture.errors || []}
-        initial={order?.total.gross.amount}
+        errors={orderPaymentCapture.opts.data?.orderCapture?.errors ?? []}
+        orderTotal={order?.total.gross ?? { amount: 0, currency: "USD" }}
+        authorizedAmount={order?.totalAuthorized ?? { amount: 0, currency: "USD" }}
         open={params.action === "capture"}
```
witoszekdev, line 405: issue: we should replace this with conditionally rendering component, and remove `open` parameter

src/components/notifications/NotificationProvider.tsx
```diff
@@ -0,0 +1,36 @@
...
+const NotificationProvider = ({ children }: PropsWithChildren) => {
+  useRegisterDevTool({
+    id: "notifications",
+    label: "Notifications",
+    component: NotificationsDebugPanel,
+  });
+
+  const context = useMemo<INotificationContext>(
+    () => ({
+      show: () => {},
```
witoszekdev, line 17: issue: if this is a drop-in replacement we should mark `show` as deprecated since we don't provide any implementation for it

witoszekdev (changes requested):
it looks to me like there are 3 PRs in one here 😅 

we have:
- redesign of notifications
- new order capture dialog
- devtools

To keep this simple let's focus on notifications and capture dialog and _if_ we device to keep devtools instead of using Storybook we can create separate PR for that.

There are issues with the code that will create bugs + some of the logic for currency input is duplicated to what we already have in Dashboard

witoszekdev:
ok, I see that these _were_ indeed 3 separate PRs 🙈 

- https://github.com/saleor/saleor-dashboard/pull/6241
- https://github.com/saleor/saleor-dashboard/pull/6200
- https://github.com/saleor/saleor-dashboard/pull/6207

@mirekm please make sure you look at the commits, if they were created from the same branch we will have all unmerged changes in the diff, and later we'll have conflicts.

This can be fixed by:

```
git rebase --onto main previous-branch feature/devtools
```

where previous-branch is the name of the branch from those 2 other PRs 😄

PR #6242, opened 2025-12-29
Redesign "What's New" modal with timeline layout and type badges

mirekm:
## New
image

## Old
image

## Data Model:
* Added type field to Ripple type (feature | improvement | bugfix)
* Updated all existing ripples with appropriate type values

## UI Improvements:
* Timeline layout
* Color-coded type badges
* Action links with accent color and animated chevron on hover
* "Last updated" footer with computed date
* Simplified ripple content (removed duplicate inline links when actions exist)


src/orders/ripples/newRefundReasons.tsx
```diff
@@ -1,29 +1,15 @@
...
-        <Paragraph>
-          Read more in{" "}
-          <Link href="https://saleor.io/blog/refund-reasons" target="_blank">
-            the article
-          </Link>
-        </Paragraph>
-      </>
-    ),
+    global:
+      'Refund reasons are now more powerful. You can add reason to the "manual refund". Additionally, you can configure refunds to use structured data to enforce specific reasons to be provided when refunds are issued.',
```
witoszekdev, line 12: can we rethink this? we are dropping a link to our blog with more details, thus making the description worse - maybe we could add a separate interface for providing link to docs if we want to get rid of React components in the description

witoszekdev (approved):

src/orders/ripples/newRefundReasons.tsx
mirekm: I agree, we should turn it into a separate links list. Let me take another swipe at this.

mirekm: You are right the links (especially to docs) should be restored. At the same time:
* The `global` still allows `ReactNode` to be provided.
* I think it's probably better if the global and popup versions come with similar actions. And docs are a good example of just `actions`, not something we inline only in the content of the bigger description. 

So, my proposal is just to move the important links, like docs, to actions atm.

src/ripples/components/Ripple.tsx
```diff
@@ -31,6 +31,24 @@ export const Ripple = (props: { model: RippleModel }) => {
...
+                    e.stopPropagation();
+                    setManuallyHidden(props.model);
+                    window.open(rippleAction.href, "_blank", "noopener,noreferrer");
+                  }}
+                >
+                  {intl.formatMessage(rippleAction.label)}
+                </Button>
+              );
+            }
+
```
lkostrowski, line 51: instead of hacky rendering button that emulates anchor, you can just render anchor

like `<a><Button><Button></a>`

lkostrowski (approved):

PR #6243, opened 2026-01-05
feat: add devcontainers

NyanKiyoshi:
This adds support for [devcontainers](https://containers.dev/) in order to make it easier & quicker to develop, and in order to isolate the project from the host machine.

Can be for example used in VSCode (https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers), or in other editors (e.g., JetBrains), or even using CLI (`docker-compose up`)

## Scope of the change

### DIFF
--- a/.changeset/calm-paws-dig.md
+++ b/.changeset/calm-paws-dig.md
@@ -1,0 +1,9 @@
+---
+"saleor-dashboard": patch
+---
+
+Prevent accidental variant metadata edits from order context
+
+Previously, users could edit product variant metadata directly from the Order Line Metadata dialog. This could be misleading because variant metadata is shared across all orders—changes made here would affect the variant globally, not just this order.
+
+Variant metadata is now displayed as read-only in the order context, with a direct link to the variant page for intentional edits.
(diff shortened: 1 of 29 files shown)


krzysztofwolski (changes requested):
@NyanKiyoshi When trying to use it with VSCode, I'm getting error:
```
Error: Service 'saleor-dashboard' configured in devcontainer.json not found in Docker Compose configuration.
```

To rule out issue with my machine, tried it on two separate ones with the same result. Any idea what might be happening?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "Saleor Dashboard",
+  "dockerComposeFile": "docker-compose.yaml",
+  "service": "saleor-dashboard",
```
krzysztofwolski: Got the fix: you can change this line to:
```
"service": "dashboard",
```

After that container started successfully and could be accessed under the port 5005. Alternative could be changing the service name in the compose file

NyanKiyoshi: Fixed at f9dd0248fe63cc3a4f9af898c7eda3ca1e22cae3, thanks!

krzysztofwolski (commented):
All good! Tested and was able to run it 👍

PR #6250, opened 2026-01-12
Refactor Link component from makeStyles to sprinkles

lkostrowski:
## Summary

This PR refactors the Link component to use the modern sprinkles styling system instead of the legacy makeStyles API from Material-UI. This modernization improves the component's maintainability and aligns it with the current design system architecture.

## Changes

- **Replaced makeStyles with sprinkles**: Migrated from Material-UI's `makeStyles` to `@saleor/macaw-ui-next`'s `sprinkles` utility for styling
- **Improved font size inheritance**: Links now properly inherit the font size from their parent element (usually paragraphs), fixing visual inconsistencies within text blocks
- **Removed TypeScript strict ignore**: Cleaned up type definitions and removed `@ts-strict-ignore` directive
- **Added comprehensive tests**: Implemented full test coverage for the Link component (481 lines of new tests)
- **Fixed color theming**: Updated color handling to use the new design system tokens (`accent1`, `default1`, `default2`)
- **Simplified props interface**: Removed deprecated `typographyProps` parameter

## Files Changed

- `src/components/Link.tsx` - Complete refactor of styling implementation
- `src/components/Link.test.tsx` - New comprehensive test suite
- `.changeset/quick-parrots-heal.md` - Changeset documenting user-facing improvements

## Testing

- Added unit tests covering all component variants and states
- Verified visual consistency across different contexts
- Tested disabled states, color variants, and underline options

## Checklist

### DIFF
--- a/src/components/Link.test.tsx
+++ b/src/components/Link.test.tsx
@@ -28,7 +28,6 @@ describe("Link component", () => {
       expect(link).toHaveAttribute("href", href);
     });
 
-    // todo should assert if state is passed on click
     it("should pass state to RouterLink", () => {
       // Arrange
       const state = { from: "/previous-page" };
@@ -335,9 +334,7 @@ describe("Link component", () => {
   });
 
   describe("onClick handler", () => {
-    // Note: There's a bug in Link.tsx where RouterLink's inline onClick overrides handleClick
-    // causing onClick prop to not be called for internal links when not disabled
-    it("should call onClick for internal RouterLink (bug)", async () => {
+    it("should call onClick for internal RouterLink", async () => {
       // Arrange
       const onClick = jest.fn();
       const user = userEvent.setup();


src/components/Link.test.tsx
```diff
@@ -0,0 +1,481 @@
...
+const Wrapper = ({ children }: { children: ReactNode }) => {
+  return (
+    // @ts-expect-error - legacy provider
+    <BrowserRouter>{children}</BrowserRouter>
+  );
+};
+
+const renderWithRouter = (component: React.ReactElement) => {
+  return render(component, { wrapper: Wrapper });
+};
```
witoszekdev, line 14: suggestion: we could make it simple and run faster if we just mock the `Link` component from react-router-dom; if for some reason we need it then we should probably use memory router instead

src/components/Link.tsx
```diff
@@ -64,62 +34,90 @@ const Link = (props: LinkProps) => {
...
+    display: inline ? "inline" : undefined,
+    ...style,
+  } as const;
+
+  const commonLinkProps = {
+    className: linkClassName,
+    onClick: handleClick,
     target,
-    rel: (rel ?? (opensNewTab && isExternalURL(href))) ? "noopener noreferer" : "",
+    rel: (rel ?? (opensNewTab && href && isExternalURL(href))) ? "noopener noreferrer" : "",
```
witoszekdev: suggestion: let's extract `(opensNewTab && href && isExternalURL(href)` to a variable so we can make it easier to read

src/components/Link.tsx
```diff
@@ -64,62 +34,90 @@ const Link = (props: LinkProps) => {
...
+
+    event.preventDefault();
+    onClick(event);
+  };
+
+  // Sprinkles doesn't support display: "inline", so we use inline styles for this
+  const inlineStyle = {
+    display: inline ? "inline" : undefined,
+    ...style,
+  } as const;
```
witoszekdev, line 74: suggestion: we could fix it in macaw in `sprinkles.css.ts`: https://github.com/saleor/macaw-ui/blob/746f4870e2be275936c28c70e9e9add7b840e5e7/src/theme/sprinkles.css.ts?plain=1#L64

```
    display: [
      "none",
      "flex",
      "inline-flex",
      "grid",
      "inline-grid",
      "block",
      "inline-block",
      "contents",
    ],

```

src/components/Link.tsx
```diff
@@ -64,62 +34,90 @@ const Link = (props: LinkProps) => {
...
-      onClick(event);
-    },
+  const colorTheme = color === "primary" ? "accent1" : "default1";
+  const textColor = disabled ? "default2" : colorTheme;
+
+  const linkClassName = clsx(
+    sprinkles({
+      cursor: disabled ? "not-allowed" : "pointer",
+      fontSize: "inherit",
+      textDecoration: underline ? "underline" : "none",
```
witoszekdev, line 49: question: maybe we should show underline by default for better accessibility?

src/components/Link.tsx
```diff
@@ -64,62 +34,90 @@ const Link = (props: LinkProps) => {
...
-              ? undefined
-              : {
-                  pathname: urlObject.pathname,
-                  search: urlObject.search,
-                  hash: urlObject.hash,
-                  state,
-                }
+    return (
+      <RouterLink<LinkState>
+        to={disabled ? "#" : routerLinkToParams}
```
witoszekdev: issue: we shouldn't use `#` for disabled state, it triggers rerender (if we have hooks for router), pushes change to history stack, scrolls to top


```suggestion
        to={routerLinkToParams}
```

src/components/Link.tsx
```diff
@@ -64,62 +34,90 @@ const Link = (props: LinkProps) => {
...
+  // Sprinkles doesn't support display: "inline", so we use inline styles for this
+  const inlineStyle = {
+    display: inline ? "inline" : undefined,
+    ...style,
+  } as const;
+
+  const commonLinkProps = {
+    className: linkClassName,
+    onClick: handleClick,
     target,
```
witoszekdev, line 81: suggestion: for disabled state we catch it by `onClick`, we should also add this so that it's clear for a11y:


```suggestion
    target,
    ["aria-disabled"]: disabled ? "true" : undefined,
```

src/components/Link.test.tsx
```diff
@@ -0,0 +1,481 @@
...
+        <Link href="/products" disabled>
+          Disabled Link
+        </Link>,
+      );
+
+      // Assert
+      const link = screen.getByRole("link", { name: "Disabled Link" });
+
+      // Verify disabled link renders (sprinkles cursor styles applied via CSS classes)
+      expect(link).toBeInTheDocument();
```
witoszekdev, line 332: suggestion: we should also use assertions from react-testing-library to make sure it's marked as disabled via HTML props (e.g. aria-disabled)

src/components/Link.tsx
lkostrowski: It's disabled link - in `onClick` event is being killed. AFAIK it's only workaround in current react router version. Previously it received `undefined` but it was invalid prop

src/components/Link.test.tsx
lkostrowski: I used MemoryRouter, I think it works better for this case. I think its good we preserve routing behavior in tests

src/components/Link.tsx
lkostrowski: https://github.com/saleor/macaw-ui/pull/942

src/components/Link.tsx
lkostrowski: This PR is a refactor, we can do that separately if we like

src/components/Link.tsx
lkostrowski: I changed it to empty string

PR #6257, opened 2026-01-19
Variants generator

mirekm:
image

Adds a "Generate variants" tool to the Product Variants section, accessible via the header menu `[⋮]`.

## Features:
- Select attribute values via chips (≤12 values) or searchable multiselect (>12 values)
- Live preview: list view for any selection, matrix view for exactly 2 attributes
- Skips existing variant combinations automatically
- Optional SKU prefix generation (defaults to product name)
- Optional initial stock quantity for all warehouses
- Confirmation dialog for 30+ variants, hard limit at 100
- Bulk creation via productVariantBulkCreate with partial success handling

## UX:
- Color swatches displayed for swatch-type attributes
- Visual indicators for new vs existing variants


src/products/components/ProductVariantGenerator/components/SelectableChip.tsx
```diff
@@ -0,0 +1,102 @@
...
+// Theme-aware color definitions
+const getChipColors = (isDark: boolean, selected: boolean, isHovered: boolean) => {
+  if (selected) {
+    // Selected: inverted colors
+    if (isDark) {
+      return {
+        bg: isHovered ? "#e5e7eb" : "#ffffff",
+        border: isHovered ? "#e5e7eb" : "#ffffff",
+        color: "#111827",
+      };
```
lkostrowski, line 31: are these colors magical? Can we use colors from macaw? if not maybe at least assign them to variables and name somehow

src/products/components/ProductVariantGenerator/useVariantGenerator.ts
```diff
@@ -0,0 +1,160 @@
...
+} from "./types";
+import { extractExistingCombinations, generateVariantPreviews } from "./utils";
+
+interface UseVariantGeneratorProps {
+  variantAttributes: VariantAttributeFragment[];
+  existingVariants: ExistingVariantData;
+  defaultSkuPrefix?: string;
+}
+
+export function useVariantGenerator({
```
lkostrowski, line 20: this contain main business logic, there should be units tests here

src/products/components/ProductVariantGenerator/components/SelectableChip.tsx
mirekm: They kind of are as the are the dark on dark double stroke technique so these are theme independent from my POV.

src/products/components/ProductVariantGenerator/useVariantGenerator.ts
mirekm: Yes, it's unit tested already.

PR #6275, opened 2026-01-23
Fix .github/workflows not using it's own pnpm-lockfile.yaml

witoszekdev:
Fixed e2e tests failing due to missing dependencies


witoszekdev:
e2e tests are failing but that's not due to the setup (like previously) but due to missing element (either slow env for PR tests, or we need to fix something in tests)

PR #6276, opened 2026-01-24
Improve table design consistency across the dashboard

mirekm:
- Unified styling for ResponsiveTable components
- Better loading states
- Improved scroll behavior for variant navigation
- Many small improvements and fixes met along the way


.changeset/bumpy-jobs-clean.md
```diff
@@ -2,4 +2,4 @@
 "saleor-dashboard": patch
 ---
 
-Update lodash version to `4.17.23`
+Improve table design consistency across the dashboard with unified styling for ResponsiveTable components, better loading states, and improved scroll behavior for variant navigation
```
witoszekdev, line 5: issue: is this expected? we should create new changesets for changes, not editing old ones because they won't be included in the changelog

src/attributes/components/AttributeValues/AttributeValues.tsx
```diff
@@ -115,114 +115,113 @@ const AttributeValues = ({
...
+            </TableHead>
+            <SortableTableBody onSortEnd={onValueReorder}>
+              {renderCollection(values, (value, valueIndex) => (
+                <SortableTableRow<"row">
+                  data-test-id="attributes-rows"
+                  className={value ? classes.link : undefined}
+                  hover={!!value}
+                  onClick={value ? () => onValueUpdate(value.id) : undefined}
+                  key={value?.id}
+                  index={valueIndex || 0}
```
witoszekdev, line 179: issue: this might break table if we have multiple same value indexes, do we even need to provide a fallback?

maybe this `renderCollection` method has invalid types (imho it should always provide an index for rendering)

src/channels/pages/ChannelsListPage/ChannelsListPage.tsx
```diff
@@ -77,73 +77,69 @@ const ChannelsListPage = ({ channelsList, limits, onRemove }: ChannelsListPagePr
...
+      <DashboardCard marginTop={6}>
+        <DashboardCard.Content>
+          <ResponsiveTable>
+            <TableHead>
+              <TableRowLink>
+                <TableCellHeader>
+                  <FormattedMessage
+                    id="j/vV0n"
+                    defaultMessage="Channel Name"
+                    description="channel name"
```
witoszekdev: suggestion: This is pretty self explanatory, we don't need duplication in the description :)

if we already have such string already in dashboard we might re-use it

```suggestion
```

src/components/AssignProductDialog/styles.ts
```diff
@@ -3,18 +3,12 @@ import { makeStyles } from "@saleor/macaw-ui";
...
       width: 72,
     },
     checkboxCell: {
-      paddingLeft: 0,
       width: 88,
     },
-    colName: {
-      paddingLeft: 0,
-    },
+    colName: {},
```
witoszekdev: question: if we don't use these styles anymore can we remove it? (see other `makeStyles in PR, they have the same issue)

```suggestion
```

src/components/ResponsiveTable/AGENTS.md
witoszekdev, line 1: I really like this :)

although do we need this basic usage? claude uses typescript lsp for type completion. We should be careful with `AGENTS.md` because if we have too many of them it can eat all the context very fast (been there 🙈)

witoszekdev, line 1: suggestion: create a symlink to `CLAUDE.md` afaik it still doesn't support this standard and will not read AGENTS.md automatically, unlike CLAUDE.md

src/attributes/components/AttributeValues/AttributeValues.tsx
```diff
@@ -115,114 +115,113 @@ const AttributeValues = ({
...
-              onUpdateListSettings={onUpdateListSettings}
+      <DashboardCard.Content>
+        {values === undefined ? (
+          <Skeleton />
+        ) : values.length === 0 ? (
+          <Placeholder>
+            <FormattedMessage
+              id="g5zIpS"
+              defaultMessage="No values found"
+              description="No attribute values found"
```
witoszekdev: nitpick: using the same style from other descriptions in this file

```suggestion
              description="attribute values list: no attribute values found"
```

src/components/CountryList/CountryList.tsx
```diff
@@ -101,66 +98,72 @@ const CountryList = (props: CountryListProps) => {
...
-              const countries = groupedCountries[letter];
+      <DashboardCard.Content>
+        {hasCountriesToRender ? (
+          <ResponsiveTable>
+            <TableBody>
+              <TableRowLink className={classes.pointer} onClick={toggleCollapse}>
+                <TableCell className={clsx(classes.wideColumn, classes.toLeft)}>
+                  <FormattedMessage
+                    id="62Ywh2"
+                    defaultMessage="{number} Countries"
```
witoszekdev: suggestion: let's use `count` feature in react-intl instead, this way we'll have correct English countability:
1: `1 Country`
2+: `2 Countires`, `10 Countries`

```suggestion
                    defaultMessage="{count, plural, one {# Country} other {# Countries}}"
```

src/components/ResponsiveTable/ResponsiveTable.tsx
```diff
@@ -1,45 +1,148 @@
...
+  };
+
+  const clearSearch = () => {
+    setSearchValue("");
+    search?.onSearchChange?.("");
+
+    if (debounceRef.current) {
+      clearTimeout(debounceRef.current);
+    }
+  };
```
witoszekdev: suggestion: debounce logic could be simplified by using helper we already have: https://github.com/saleor/saleor-dashboard/blob/57cef5304ea96f4d178a790833998dd5d2e2ba8e/src/hooks/useDebounce.ts?plain=1#L4

src/components/ResponsiveTable/ResponsiveTable.tsx
```diff
@@ -1,45 +1,148 @@
...
+
+  const handleSearchKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === "Escape") {
+      clearSearch();
```
witoszekdev, line 56: suggestion: I think it would be better UX if we would execute clear immediately when user presses `Escape`, since we don't need to wait for another value to come, this is explicitly clearing (same with using Cmd + Backspace in input: it's meant for clearing everything)

src/components/ResponsiveTable/ResponsiveTable.module.css
```diff
@@ -0,0 +1,81 @@
...
+.colActionWide {
+  width: 120px;
+  text-align: right;
+}
+
+@media (min-width: 960px) {
+  .table {
+    table-layout: fixed;
+  }
+}
```
witoszekdev, line 81: nitpick: imho this is fine for now, but maybe we should migrate to a new custom Table component in macaw instead of fighting with legacy macaw-ui styles which we want to get rid of anyway?

src/components/ResponsiveTable/ResponsiveTable.module.css
```diff
@@ -0,0 +1,81 @@
+.container {
+  width: 100%;
+}
```
witoszekdev, line 3: suggestion: this can be replaced with `<Box>` component

src/components/Savebar/SavebarActionButtons.tsx
```diff
@@ -17,20 +17,43 @@ export const DeleteButton = ({
...
+      {children || <FormattedMessage {...buttonMessages.save} />}
+    </ConfirmButtonComponent>
+  );
+
+  if (tooltip && disabled) {
+    return (
+      <Tooltip>
+        <Tooltip.Trigger>
+          {/* Box wrapper needed to capture hover events on disabled button */}
+          <Box display="inline-block">{button}</Box>
```
witoszekdev, line 45: question: do we really need this?

macaw-ui already creates a Box to wrap buttons:

https://github.com/saleor/macaw-ui/blob/746f4870e2be275936c28c70e9e9add7b840e5e7/src/components/Tooltip/Content.tsx?plain=1#L35

```ts
        <Box
          borderStyle={children ? "solid" : "none"}
          borderWidth={1}
          borderRadius={3}
          borderColor="default1"
          padding={children ? 4 : 0}
          fontSize={1}
          lineHeight={1}
          color="default1"
          backgroundColor="default1"
          boxShadow="defaultOverlay"
        >
          {children}
        </Box>

```

src/components/TableHead/TableHead.tsx
```diff
@@ -101,32 +81,25 @@ const TableHead = (props: TableHeadProps) => {
...
-                )}
-              </div>
-            </TableCell>
-          </>
+          <TableCell colSpan={getColSpan(colSpan, dragRows)}>
+            <div className={classes.container}>
+              <Text data-test-id="SelectedText">
+                <FormattedMessage
+                  id="qu/hXD"
+                  defaultMessage="Selected {number} items"
```
witoszekdev: suggestion: let's add a proper plural syntax (selected 1 item, selected 2 items)


```suggestion
                  defaultMessage="Selected {number, plural, one {# item} other {# items}}"
```

src/products/components/ProductVariantNavigation/components/ProductVariantItem.tsx
```diff
@@ -27,18 +28,31 @@ export const VariantItem = ({
...
+    (node: HTMLDivElement | null) => {
+      setNodeRef(node);
+
+      if (isActive && node && !scrolledRef.current) {
+        node.scrollIntoView({ block: "nearest" });
+        scrolledRef.current = true;
+      }
+    },
+    [isActive, setNodeRef],
+  );
```
witoszekdev, line 43: suggestion: let's add this as a separate hook (e.g. `useScrollIntoView`)

src/taxes/components/TaxMenu/index.ts
witoszekdev, line 1: issue: let's get rid of barell files (`index.ts`), we want to get rid of them in dashboard

src/taxes/pages/TaxChannelsPage/TaxChannelsMenu/TaxChannelsMenu.tsx
```diff
@@ -1,53 +1,33 @@
...
+    <TaxMenu
+      title={intl.formatMessage(taxesMessages.channelList)}
+      columnHeader={<FormattedMessage {...taxesMessages.channelNameHeader} />}
+      items={items}
+    />
   );
 };
 
-export default TaxChannelsMenu;
+export { TaxChannelsMenu };
```
witoszekdev: suggestion: let's use regular export syntax for better readability:

`export const TaxChannelsMenu`

instead of:

`export { TaxChannelsMenu }`

AGENTS.md
```diff
@@ -127,6 +127,21 @@ The codebase follows a feature-based architecture with shared components:
...
+If the list could be long:
+
+- Add scrollable container with `max-height`
+- Keep headers outside the scroll area
+- Auto-scroll to active item on load
+
+Other considerations:
+
+- Loading states: Show skeletons
+- Empty states: Handle zero items gracefully
```
witoszekdev, line 143: question: maybe in general we should migrate to skills for specific tasks like this one (writing / refactoring components that display list) - I think it works better in general: agents.md tends to get ignored + we pollute our context less.

mirekm: Yes - 100%. I internally use both rules and skills. Just had no time to pull it together yet as I initially did for our storefront on the Paper PR/branch.

.changeset/bumpy-jobs-clean.md
mirekm: No, it's not expected. I will issue a new change log and revert that one.

src/components/ResponsiveTable/AGENTS.md
mirekm: Yes, we need to be careful with those but most importantly we have to switch to rules and skills.

src/attributes/components/AttributeValues/AttributeValues.tsx
mirekm: That would have to be either type overload or a rewrite of many files again so they all follow the same pattern of rendering skeletons.

witoszekdev: I've checked renderCollection method and this is "by design" (although probably not very good one 😄)

```ts
export function renderCollection<T>(
  collection: T[] | undefined,
  renderItem: (item: T | undefined, index: number | undefined, collection: T[] | undefined) => any,
  renderEmpty?: (collection: T[]) => any,
) {
  if (collection === undefined) {
    return renderItem(undefined, undefined, collection);
  }

  if (collection.length === 0) {
    return renderEmpty ? renderEmpty(collection) : null;
  }

  return collection.map(renderItem);
}
```

If we have empty collections list, meaning it's still fetching we will call this inner function with `value=undefined` and `valueIndex=undefined`. Then we already handle this and render a skeleton, so using fallback for `index` is fine, because we can only render one item like this.

Mystery solved

PR #6277, opened 2026-01-26
New product channel availability and diagnostics

mirekm:
## Redesigned Product Availability Panel

Replaces the existing channels availability card with a new unified panel that combines availability settings with real-time diagnostics.

### Key Features

**Availability Diagnostics**
- Analyzes product configuration per channel to identify issues preventing purchase (missing prices, no stock, inactive channels, missing shipping zones)
- Shows warnings for configuration problems before they affect customers
- Gracefully handles permission restrictions - shows available diagnostics even with limited access

**Public API Verification**
- "Test" button that queries the public (unauthenticated) API to verify what storefronts/clients actually see
- Confirms if product is visible, purchasable, and has stock from the storefront perspective
- Useful for debugging "why can't customers see the product?" issues

**Scheduling Support**
- Clear visual distinction between "live now" vs "scheduled" states

## Product Availability Diagnostics

The Availability Doctor detects the following issues that could prevent a product from being purchasable:

### Errors (blocking issues)

| Issue | Description |
|-------|-------------|
| **Channel inactive** | The channel is not active. Products cannot be purchased until the channel is activated. |
| **No variants created** | Product has no variants. Create at least one variant before customers can purchase. |
| **No variant listed in channel** | Variants exist but none are available in this channel. Add variant channel listings. |
| **No variant has a price** | Set a price for at least one variant in this channel. Price can be 0 for free products. |

### Warnings (potential issues)

| Issue | Description |
|-------|-------------|
| **No warehouses assigned** | Products in this channel cannot be fulfilled without warehouses. |
| **No stock in warehouses** | Add stock to at least one warehouse assigned to this channel. |
| **No shipping zones** | Customers cannot checkout without shipping methods. Add shipping zones to this channel. |
| **Stock unreachable** | Warehouses with stock are not assigned to any shipping zone in this channel. |

### Permission-aware diagnostics

- **Core checks** (variants, pricing, channel status) always run
- **Warehouse checks** require access to channel warehouse data
- **Shipping checks** require access to shipping zone data
- When permissions are limited, the UI shows "Limited diagnostics" with missing permissions listed

### API Impact

Adds one new query on product page load (`ChannelDiagnostics`) to fetch channel/warehouse/shipping zone data for diagnostics. Public API verification calls are user-initiated and debounced.

### Screenshots

<table>
  <tr>
    <td>aval1</td>
    <td>aval2</td>
    <td>aval3</td>
  </tr>
  <tr>
    <td>aval4</td>
    <td>aval5</td>
    <td></td>
  </tr>
</table>

### DIFF
--- a/src/products/components/ProductDoctor/AvailabilityCard.tsx
+++ b/src/products/components/ProductDoctor/AvailabilityCard.tsx
@@ -6,7 +6,6 @@ import {
   ProductChannelListingAddInput,
   ProductChannelListingErrorFragment,
 } from "@dashboard/graphql";
-import { areChannelFieldsDifferent } from "@dashboard/products/components/ProductUpdatePage/formChannels";
 import { Accordion, Box, Button, Skeleton, Spinner, Text, Tooltip } from "@saleor/macaw-ui-next";
 import { CheckCircle, ChevronLeft, ChevronRight, Info, Search, X, XCircle } from "lucide-react";
 import React, { useEffect, useMemo, useState } from "react";
@@ -19,11 +18,14 @@ import {
 } from "./hooks/usePublicApiVerification";
 import { messages } from "./messages";
 import {
-  AvailabilityIssue,
-  ChannelSummary,
-  DiagnosticsPermissions,
-  DiagnosticsResult,
-} from "./utils/types";
+  countIssuesBySeverity,
+  filterChannelsBySearch,
+  getDirtyChannelIds,
+  groupIssuesByChannel,
+  paginateItems,
+} from "./utils/channelUtils";
+import { mergeFormDataWithChannelSummaries } from "./utils/mergeChannelSummaries";
+import { DiagnosticsPermissions, DiagnosticsResult } from "./utils/types";
 
 interface AvailabilityCardProps {
   diagnostics: DiagnosticsResult;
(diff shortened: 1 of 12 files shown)


src/products/components/ProductDoctor/AvailabilityChannelItem.tsx
```diff
@@ -0,0 +1,460 @@
...
+  disabled = false,
+  errors = [],
+  issues = [],
+  isExpanded = false,
+  verificationResult,
+  onVerify,
+}: AvailabilityChannelItemProps) => {
+  const intl = useIntl();
+  const dateNow = useCurrentDate();
+  const status = getAvailabilityStatus(originalSummary ?? summary, dateNow);
```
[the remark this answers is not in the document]
mirekm: The current behaviour is exactly as intended now.

src/products/components/ProductDoctor/AvailabilityChannelItem.tsx
```diff
@@ -0,0 +1,460 @@
...
+
+  const getStatusDescription = () => {
+    switch (status) {
+      case "live":
+        if (isPurchasable(originalSummary ?? summary, dateNow)) {
+          return intl.formatMessage(messages.statusDescription_live);
+        }
+
+        return intl.formatMessage(messages.statusDescription_liveNotPurchasable);
+      case "scheduled":
```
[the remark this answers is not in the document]
mirekm: As intended.

src/products/components/ProductDoctor/hooks/usePublicApiVerification.ts
```diff
@@ -0,0 +1,273 @@
...
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+      // Explicitly NOT including Authorization header
+    },
+    body: JSON.stringify({
+      query,
+      variables,
+    }),
+  });
```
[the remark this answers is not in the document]
mirekm: Addressed.

src/products/components/ProductDoctor/hooks/usePublicApiVerification.ts
```diff
@@ -0,0 +1,273 @@
...
+        ...prev,
+        isVerifying: true,
+        results: new Map(prev.results).set(channelId, {
+          channelId,
+          channelSlug,
+          status: "loading",
+          result: null,
+          error: null,
+        }),
+      }));
```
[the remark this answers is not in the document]
mirekm: Addressed.

src/products/views/ProductUpdate/handlers/utils.ts
```diff
@@ -147,8 +144,10 @@ export function getProductChannelsUpdateVariables(
 
       return {
         ...data,
+        // Only force isAvailableForPurchase=true when there's an actual date set
+        // Use loose equality (!=) to treat both null and undefined as "no date"
         isAvailableForPurchase:
-          data.availableForPurchaseAt !== null ? true : data.isAvailableForPurchase,
+          data.availableForPurchaseAt != null ? true : data.isAvailableForPurchase,
```
lkostrowski: I'd prefer if we dont ever use `!=`, can you just explicitly check null/undefined or somehow flip it?

src/products/components/ProductUpdatePage/formChannels.ts
lkostrowski, line 1: add tests, at least for your changes

src/products/components/ProductDoctor/utils/mapProductToDiagnosticData.ts
lkostrowski, line 1: add test

src/products/components/ProductDoctor/utils/dateUtils.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Check if a date string represents a future time.
+ */
+export const isFutureDate = (date: string | null | undefined): boolean => {
+  if (!date) {
+    return false;
+  }
```
lkostrowski, line 7: its not best idea, because if you pass wrong value (null, random string), it will gracefully fall back. If entire architecture is hard to fix, I can accept that, but in general we should create precise helpers with strict types, which should throw errors

For example
1. Parent wants to check date
2. `isFutureDate` throws error
3. Parent should try/catch (error boundary), call Sentry, show user error

This way
1. We know there is a bug and we can fix it
2. User knows there is an error, instead arbitrary value (what if date is actually future, but due to error we show "false" ?)

src/products/components/ProductDoctor/sections/index.ts
lkostrowski, line 1: avoid index files (lets add claude skill or something), it makes code hard to maintain due to circular dependencies

src/products/components/ProductDoctor/utils/availabilityChecks.ts
lkostrowski, line 1: heavy business logic, definitely need tests

src/products/components/ProductDoctor/primitives.tsx
```diff
@@ -0,0 +1,232 @@
...
+/**
+ * Reusable icon wrapper that provides consistent sizing and color based on state.
+ */
+interface IconWrapperProps {
+  isActive: boolean;
+  activeColor?: string;
+  children: React.ReactNode;
+}
+
+export const IconWrapper = ({
```
lkostrowski, line 212: maybe not now, but probably this should be global component (src/components) or even macaw?

src/products/components/ProductDoctor/index.ts
lkostrowski, line 1: remove index file

src/products/components/ProductDoctor/hooks/useProductAvailabilityDiagnostics.ts
lkostrowski, line 1: add test

src/products/components/ProductDoctor/hooks/usePublicApiVerification.ts
lkostrowski, line 1: add test

src/products/components/ProductDoctor/AvailabilityCard.tsx
lkostrowski, line 1: heavy inline logic, best if you extract to separate files as helper functions/classes, test them and make markup file cleaner and simpler. debugging view mixed with logic is difficult

src/products/views/ProductUpdate/handlers/utils.ts
mirekm: We may want to add it to ESLint I think too.

src/products/components/ProductDoctor/primitives.tsx
mirekm: Would be nice - the pattern with colorful icon that is "on" will be used in other places too.

src/products/components/ProductDoctor/utils/dateUtils.ts
mirekm: I absolutely see your point but I would prefer to accept a tradeoff for simplicity in this case as managing the state for switches and dates was already quite hard.

src/components/ChannelsAvailabilityCard/types.tsx
```diff
@@ -4,7 +4,8 @@ import {
 } from "@dashboard/graphql";
 
 export interface ChannelOpts {
-  availableForPurchase?: string;
+  // Can be string (date), null (clear date), or undefined (don't change)
+  availableForPurchase?: string | null;
```
lkostrowski, line 8: I can accept that, but I personally would introduce better data model, becuase null/undefined are different-but-similar and usually developers use it interchangeably, which may lead to a bug

Additionally (not your change), but `availableForPurchase?: string` doesnt really indicate its a "date", without a comment, meaning this is not good data model

what can be done for example:

## Example 1

This is domain modelling, more code, but explicit about business intent, especially valuable on the backend

```typescript
type AvailableForPurchaseStrategy = FutureDate | Clear | PreserveBehavior

class FutureDate {
  private  strategy: "future-date"
  private date: Date

  constructor(date...)

  toGraphqlVariable() {return this.date}
}

type Clear = {
  private strategy: "clear

  toGraphqlVariable(){ return null }
}

type PreserveBehavior = {
  private  strategy: "PreserveBehavior"

  toGraphqlVariable(){ return undefined }
}
```

## Example 2

Simpler one , probably for this PR

```availableForPurchase?: Date | null;``` or ```availableForPurchase: Date | "clear" | "preserve"```

It is much less bug-prone if you force value explicitly and map it to "null" on the network/graphql layer

mirekm: Thanks, I will address it in the follow up PR.

PR #6279, opened 2026-01-27
Add Claude skills for business logic separation and no-barrel-exports

lkostrowski:
## Summary

- Adds a **separate-business-logic** Claude skill that enforces extracting business logic from React components into standalone, testable utility files following a `logic → hook → component` pattern
- Adds a **no-barrel-exports** Claude skill that prevents creation of `index.ts`/`index.tsx` barrel files and enforces direct imports from source files

Both skills align with existing conventions documented in `CLAUDE.md` and provide structured guidance for Claude Code when writing or refactoring code in this repository.

## Changes

- `.claude/skills/separate-business-logic/SKILL.md` — New skill with rules for extracting data transformations, validation, formatting, and conditional logic out of components, along with file naming conventions, testing expectations, and examples
- `.claude/skills/no-barrel-exports/SKILL.md` — New skill enforcing direct imports instead of barrel re-exports, with clear exceptions for route entry points

## Test plan

- [ ] Verify skills are recognized by Claude Code in this repository
- [ ] Test that the `separate-business-logic` skill triggers when refactoring a component with inlined logic
- [ ] Test that the `no-barrel-exports` skill triggers when creating new components or modules


.claude/skills/no-barrel-exports/SKILL.md
witoszekdev, line 1: suggestion: for enforcement SKILLS are brittle, claude loads them dynamically - it might load or not and it can just straight up ignore it 😄. It's better to use them to perform certain tasks in a way we want (e.g. separate business logic)

Maybe we could rewrite this to a hook instead? This way we can check if Claude edits / creates `index.ts` / `index.tsx` and return an error there to **force** it to not create such files programatically

Example: 
https://github.com/saleor/saleor-dashboard/blob/57cef5304ea96f4d178a790833998dd5d2e2ba8e/.claude/hooks/check-as-any-pre.py?plain=1#L4

.claude/skills/separate-business-logic/SKILL.md
```diff
@@ -0,0 +1,127 @@
...
+
+# Separate Business Logic
+
+## Goal
+
+Extract business logic out of React components into standalone, testable files. Components should focus on rendering and user interaction, delegating data transformations, mapping, filtering, sorting, formatting, validation, and decision-making to separate modules.
+
+## What Counts as Business Logic
+
+- Data mapping and transformations (e.g., converting API responses to UI models)
```
witoszekdev, line 19: suggestion: let's add a description so that it doesn't overengineer data mapping - a simple `data.map()` in JSX is fine imho, if we don't calculate new values just render a list for example

.claude/skills/separate-business-logic/SKILL.md
```diff
@@ -0,0 +1,127 @@
...
+- Filtering, sorting, grouping collections
+- Validation rules and error message resolution
+- Conditional logic that determines what to display or which action to take
+- Formatting (dates, currencies, percentages, labels)
```
witoszekdev, line 23: suggestion: we should use `react-intl` for formatting most of the time, so maybe let's also add a note? otherwise i suspect we might end up with more and more utility functions that are built into libraries we already use

.claude/skills/separate-business-logic/SKILL.md
```diff
@@ -0,0 +1,127 @@
...
+
+Place logic in a separate file **next to the component** that uses it:
+
+```
+ComponentName/
+├── ComponentName.tsx          # UI only
+├── useComponentName.ts        # Hook (optional, bridges logic → component)
+├── componentNameUtils.ts      # Pure logic functions
+├── componentNameUtils.test.ts # Unit tests for the logic
+└── ComponentName.module.css
```
witoszekdev, line 37: question: do we want to move ahead with writing `.module.css`? I remember we were thinking about using `vanilla-extract` since macaw-ui is already using it and imho it's a pretty good library, less verbose than css modules and has nice type checking

.claude/skills/separate-business-logic/SKILL.md
```diff
@@ -0,0 +1,127 @@
...
+   };
+   ```
+
+5. **Test the logic, not the wiring.** Unit test `*Utils.ts` functions thoroughly — edge cases, empty inputs, error states. Hook tests and component tests can be lighter since they just verify integration.
+
+6. **Name functions by intent.** Use descriptive names: `getActiveFilters`, `formatPriceRange`, `resolveStatusBadgeColor` — not `processData` or `transform`.
+
+## Testing Expectations
+
+- Every `*Utils.ts` file must have a corresponding `*.test.ts` file.
```
witoszekdev, line 78: suggestion: let's also add explicitly hooks

```suggestion
- Every `*Utils.ts` and `use*.ts` file must have a corresponding `*.test.ts` file.
```

.claude/skills/separate-business-logic/SKILL.md
```diff
@@ -0,0 +1,127 @@
...
+    const items: Item[] = [];
+
+    // Act
+    const result = calculateTotal(items);
+
+    // Assert
+    expect(result).toBe(0);
+  });
+});
+```
```
witoszekdev, line 114: suggestion: maybe we should separate examples into a separate file in a skill directory? This is a good way to save context for specific use cases, e.g. writing test

https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview#level-3-resources-and-code-loaded-as-needed

.claude/skills/separate-business-logic/SKILL.md
witoszekdev, line 1: suggestion: in general maybe we should also add instructions about finding existing utilities? I think we have plenty of them (e.g. related to prices) - it would be great if we would re-use that instead of creating more and more code

PR #6283, opened 2026-01-27
App Problems implementation 

lkostrowski:
CleanShot 2026-01-29 at 10 42 47

https://github.com/saleor/saleor/pull/18724

### DIFF
--- a/.github/workflows/deploy-cloud.yaml
+++ b/.github/workflows/deploy-cloud.yaml
@@ -150,6 +150,7 @@ jobs:
       status: ${{ needs.build.result }}
       product: dashboard
       environment: ${{ needs.build.outputs.ENVIRONMENT }}
+      mention_on: failure
     secrets:
       mention_group_id: ${{ secrets.SLACK_DASHBOARD_GROUP_ID }}
       slack-webhook-url: ${{ secrets.SLACK_CLOUD_DEPLOYMENTS_WEBHOOK_URL }}
--- a/.github/workflows/deploy-master-staging.yaml
+++ b/.github/workflows/deploy-master-staging.yaml
@@ -99,6 +99,7 @@ jobs:
       status: ${{ needs.build.result }}
       product: dashboard
       environment: saleor-master-staging
+      mention_on: failure
     secrets:
       mention_group_id: ${{ secrets.SLACK_DASHBOARD_GROUP_ID }}
       slack-webhook-url: ${{ secrets.SLACK_CLOUD_DEPLOYMENTS_WEBHOOK_URL }}
--- a/.github/workflows/deploy-staging-and-prepare-release.yaml
+++ b/.github/workflows/deploy-staging-and-prepare-release.yaml
@@ -200,6 +200,7 @@ jobs:
       status: ${{ needs.build.result }}
       product: dashboard
       environment: ${{ needs.build.outputs.ENVIRONMENT }}
+      mention_on: failure
     secrets:
       mention_group_id: ${{ secrets.SLACK_DASHBOARD_GROUP_ID }}
       slack-webhook-url: ${{ secrets.SLACK_CLOUD_DEPLOYMENTS_WEBHOOK_URL }}


schema-main.graphql
lkostrowski, line 1: this will be 3.22 but ensure you add info to changelog which saleor release is required

locale/defaultMessages.json
```diff
@@ -823,12 +829,18 @@
...
+  },
   "2qHTHI": {
     "string": "Default weight unit updated"
   },
   "2qfU+C": {
     "string": "The extension identifier is already in use."
   },
+  "2rUVsU": {
+    "string": "App"
+  },
```
witoszekdev: suggestion: i'm pretty sure we already have translation string for "App", let's reuse what we already have

locale/defaultMessages.json
```diff
@@ -4640,6 +4673,9 @@
     "context": "input label",
     "string": "Model type"
   },
+  "OYWn0V": {
+    "string": "CircuitBreaker"
```
witoszekdev: question: typo?


```suggestion
    "string": "Circuit Breaker"
```

src/extensions/components/AppAlerts/useAppsAlert.ts
```diff
@@ -15,8 +17,33 @@ export const useAppsAlert = () => {
     useSidebarDotState();
   const { lastFailedWebhookDate, fetchAppsWebhooks } = useAppsFailedDeliveries();
 
+  const [fetchInstalledApps, { data: installedAppsData }] = useInstalledAppsListLazyQuery({
+    fetchPolicy: "no-cache",
+  });
```
witoszekdev: question: do we need to use this query? afaik it's meant for the apps list, which has more details than we need here (app problems)

src/extensions/components/AppAlerts/useAppsAlert.ts
```diff
@@ -15,8 +17,33 @@ export const useAppsAlert = () => {
...
+    if (hasManagedAppsPermission) {
+      fetchInstalledApps({ variables: { first: 100 } });
+    }
+  };
+
+  // Fetch immediately on mount
+  useEffect(() => {
+    if (hasManagedAppsPermission) {
+      fetchAll();
+    }
```
witoszekdev: issue: we don't need `if` here - since `fetchAll` already checks if user has permissions. We wouldn't fetch the list if user doesn't have manage apps permission

```suggestion
  fetchAll()
```

src/extensions/components/AppAlerts/useAppsAlert.ts
```diff
@@ -28,8 +55,14 @@ export const useAppsAlert = () => {
     }
   }, [lastFailedWebhookDate]);
 
+  useEffect(() => {
+    if (hasAppProblems) {
+      handleFailedAttempt(new Date().toISOString());
+    }
+  }, [hasAppProblems]);
```
witoszekdev: question: I don't quite get why do we need to do this? 🤔

src/extensions/views/InstalledExtensions/components/AppProblems/AppProblems.module.css
witoszekdev, line 1: Most of this file could be rewritten in <Box> so that we can use our design tokens from macaw-ui, if we need to use .module.css, we should use `var`'s to re-use our color and size definitions

src/extensions/views/InstalledExtensions/components/AppProblems/fixtures.ts
witoszekdev, line 1: tip: we can use our new codegen to build mocked data: https://github.com/saleor/saleor-dashboard/pull/6334

src/extensions/views/InstalledExtensions/components/AppProblems/ProblemCard.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+  dismissedByUserProblem,
+  warningAppProblem,
+  webhookDeliveryError,
+} from "./fixtures";
+import { ProblemCard } from "./ProblemCard";
+
+const withLocale: Decorator = (Story: React.FC) => (
+  <LocaleContext.Provider value={{ locale: "en" as any, setLocale: () => {} }}>
+    <Story />
+  </LocaleContext.Provider>
```
witoszekdev: suggestion: maybe we should add this to global decorator so we can re-use it everywhere? or if it doesn't need to be used everywhere into `storybookUtils`

src/extensions/views/InstalledExtensions/components/AppProblems/ProblemCard.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export default meta;
+
+type Story = StoryObj<typeof ProblemCard>;
+
+export const Critical: Story = {
+  args: {
+    problem: criticalAppProblem,
+    onForceClear: () => alert("Force clear clicked"),
```
witoszekdev: suggestion: instead of using `alert` we should probably rather use `fn` from storybook, so that actions pop-up in Actions tab in storybook: https://github.com/saleor/saleor-dashboard/blob/8b4d8b1cfdc77cfa57927b0f3574d8b1749c017d/src/components/AssignCategoryDialog/AssignCategoryDialog.stories.tsx

src/extensions/views/InstalledExtensions/components/AppProblems/ProblemCard.tsx
```diff
@@ -0,0 +1,175 @@
...
+  }, [date, locale]);
+};
+
+const TimestampWithTooltip = ({ date, label }: { date: string; label?: string }) => {
+  const intl = useIntl();
+  const { short, local, utc } = useFormattedDates(date);
+
+  return (
+    <>
+      {label && `${label} `}
```
witoszekdev: nitpick: could be simplified, so see it's used for space:


```suggestion
      {label}
      {" "}
```

src/extensions/views/InstalledExtensions/components/AppProblems/ProblemCard.tsx
```diff
@@ -0,0 +1,175 @@
...
+    </span>
+  );
+};
+
+const useDismissedByText = (problem: AppProblem): string | null => {
+  const intl = useIntl();
+
+  if (problem.__typename !== "AppProblem" || !problem.dismissed) {
+    return null;
+  }
```
witoszekdev: nitpick: not sure why we check for typename, since it can only accept `AppProblem`, if it's meant for loading state we should probably just check it like so:


```suggestion
  if (!problem || !problem.dismissed) {
    return null;
  }
```

Image

src/extensions/views/InstalledExtensions/components/AppProblems/ProblemCard.tsx
witoszekdev, line 1: suggestion: it's getting a bit hard to read from a single file, let's maybe separate all the inner components to a separate file (e.g. `ProblemCard/ProblemCard.tsx`, `ProblemCard/components/...`)

src/extensions/views/InstalledExtensions/components/AppProblems/ProblemsBadge.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const intl = useIntl();
+
+  if (totalCount === 0) {
+    return null;
+  }
+
+  const label =
+    criticalCount > 0
+      ? `${intl.formatMessage(problemMessages.problemCount, { count: totalCount })}, ${intl.formatMessage(problemMessages.includingCritical, { count: criticalCount })}`
+      : intl.formatMessage(problemMessages.problemCount, { count: totalCount });
```
witoszekdev: suggestion: we could simplify this by using ICU message syntax:

```
"problemSummary": "{total, plural, one {# problem} other {# problems}}{critical, plural, =0 {} other {, including {active, plural, =0 {{critical} critical} other {{active} active and {critical} critical}}}}"
```

```ts
const label = intl.formatMessage(problemMessages.problemSummary, {
  total: totalCount,
  active: activeCount,
  critical: criticalCount,
});
```

src/extensions/views/InstalledExtensions/components/AppProblems/ProblemsBadge/ProblemsBadge.tsx
```diff
@@ -0,0 +1,45 @@
...
+      className={criticalCount > 0 ? styles.problemsBadgeError : styles.problemsBadgeWarning}
+      aria-expanded={expanded}
+      onClick={e => {
+        e.preventDefault();
+        onToggle();
+      }}
+    >
+      {label}
+      {expanded ? <ChevronUp size={14} /> : <ChevronDown size={14} />}
+    </button>
```
witoszekdev, line 44: afaik we could re-use existing component from macaw-ui: https://macaw-ui.vercel.app/?path=/docs/components-accordion--docs

it would have better a11y, since it uses radix: https://www.radix-ui.com/primitives/docs/components/accordion

src/extensions/views/InstalledExtensions/components/AppProblems/ProblemsList.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const dismissed = isProblemDismissed(problem);
+  const actionLink = getActionLink(problem, appId, appType);
+  const canForceClear =
+    hasManagedAppsPermission && problem.__typename === "AppProblem" && !!onClearProblem;
+
+  const borderClass = dismissed
+    ? styles.severityDismissed
+    : critical
+      ? styles.severityError
+      : styles.severityWarning;
```
witoszekdev: nitpick: this could use a if block or switch statempt instead of turnary to make it easier to read

src/extensions/views/InstalledExtensions/components/AppProblems/ProblemsList/ProblemsList.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const [internalPopupOpen, setInternalPopupOpen] = useState(false);
+
+  const isPopupOpen = modalOpen ?? internalPopupOpen;
+  const handlePopupChange = (open: boolean) => {
+    if (onModalOpenChange) {
+      onModalOpenChange(open);
+    } else {
+      setInternalPopupOpen(open);
+    }
+  };
```
witoszekdev, line 55: question: i'm not sure what's going on here: why do we have modal prop and also internal state in `useState`?

src/extensions/views/InstalledExtensions/components/InstalledExtensionsList/InstalledExtensionsList.tsx
```diff
@@ -42,27 +53,138 @@ const ExtensionLink = ({
...
+}
+
+const ExtensionRow = ({
+  extension,
+  hasManagedAppsPermission,
+  onClearProblem,
+}: ExtensionRowProps) => {
+  const intl = useIntl();
+  const problems = extension.problems ?? [];
+  const activeProblems = problems.filter(p => !isProblemDismissed(p));
```
witoszekdev: suggestion: this should probably use `useMemo` since it will iterate over list on each render

src/extensions/views/InstalledExtensions/components/InstalledExtensionsList/InstalledExtensionsList.tsx
```diff
@@ -42,27 +53,138 @@ const ExtensionLink = ({
...
+  const totalCount = activeProblems.length;
+  const criticalCount = activeProblems.filter(p => isProblemCritical(p)).length;
+
+  const hasActiveProblems = totalCount > 0;
+  const hasAnyProblems = problems.length > 0;
+
+  const [problemsVisible, setProblemsVisible] = useState(hasActiveProblems);
+  const [modalOpen, setModalOpen] = useState(false);
```
witoszekdev: suggestion: let's put it into separate hook, so can be tested separately + readability

src/extensions/views/InstalledExtensions/hooks/useInstalledExtensions.tsx
```diff
@@ -178,9 +190,18 @@ export const useInstalledExtensions = () => {
     [installedPluginsData],
   );
 
+  const activeProblems = installedApps.flatMap(
+    app => app.problems?.filter(p => !isProblemDismissed(p)) ?? [],
+  );
+
+  const totalCount = activeProblems.length;
+  const criticalCount = activeProblems.filter(p => isProblemCritical(p)).length;
```
witoszekdev: i saw we calculate the same values somewhere else as well, maybe it should be calculated in once place and passed down as context / props?

src/extensions/messages.ts
```diff
@@ -338,6 +338,113 @@ export const messages = defineMessages({
   },
 });
 
+export const problemMessages = defineMessages({
```
witoszekdev, line 341: suggestion: let's add `context` for each message, so it's clear what are they used for (e.g. app problems / ...)

src/extensions/messages.ts
```diff
@@ -338,6 +338,113 @@ export const messages = defineMessages({
...
+    id: "1aAwu4",
+  },
+  showMoreIncludingActive: {
+    defaultMessage: ", including {active} active",
+    id: "CyYBH5",
+  },
+  showMoreIncludingCritical: {
+    defaultMessage: ", including {critical} critical",
+    id: "3OiSHe",
+  },
```
witoszekdev: duplicate of `includingCritical`

src/extensions/types.ts
```diff
@@ -1,11 +1,51 @@
...
+
+export type AppProblem = GraphQLAppProblem | WebhookDeliveryProblem;
+
+export const isProblemCritical = (problem: AppProblem): boolean => {
+  if (problem.__typename === "AppProblem") {
+    return problem.isCritical;
+  }
+
+  // WebhookDeliveryError is a warning, not critical
+  return false;
```
witoszekdev, line 30: idea: maybe we could mark it as critical if app had too many WebhookDeliveryError's (e.g. circuit breaker tripped)

src/fragments/apps.ts
```diff
@@ -202,6 +202,20 @@ export const InstalledAppDetails = gql`
...
+      createdAt
+      count
+      isCritical
+      dismissed {
+        by
+        userEmail
+      }
+      updatedAt
+      id
+    }
```
witoszekdev, line 218: suggestion: we should probably add some limit to problems we fetch on initial list render (we fetch all apps), e.g. only recent 3; rest should be fetched when user interacts with the list

src/extensions/views/InstalledExtensions/components/AppProblems/ProblemsBadge/ProblemsBadge.tsx
witoszekdev: bump - afaik we won't have proper a11y with current problems list because this button doesn't tell which element is expanded, it should have `aria-controls="id"` with id of the expanded div

src/extensions/types.ts
lkostrowski: we dont have circuit breaker integrated yet, this should happen via backend first

src/fragments/apps.ts
lkostrowski: on main extensions page we show number of problems and allow to expand. if we fetch 3, we lose that information. I would need to update API first to give me total number of problems

PR #6301, opened 2026-01-29
Feature/attribute values search

mirekm:
- Add search input to Attribute Values table for filtering by slug or name
- Search only appears when there are more than 5 values to avoid UI clutter

image

### DIFF
--- a/src/attributes/components/AttributeValues/AttributeValues.tsx
+++ b/src/attributes/components/AttributeValues/AttributeValues.tsx
@@ -136,7 +136,7 @@ const AttributeValues = ({
               <Box position="relative">
                 <SearchInput
                   value={searchQuery}
-                  onChange={onSearchChange}
+                  onChange={onSearchChange!}
                   placeholder={intl.formatMessage({
                     id: "9seX5T",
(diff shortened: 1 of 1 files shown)


src/attributes/components/AttributeValues/AttributeValues.tsx
```diff
@@ -118,113 +129,151 @@ const AttributeValues = ({
...
-                  />
-                </TableCell>
-                <TableCell className={classes.columnStore}>
-                  <FormattedMessage
-                    id="H60H6L"
-                    defaultMessage="Default Store View"
-                    description="attribute values list: name column header"
+          <Box display="flex" flexDirection="column" gap={4}>
+            {/* Search input - always visible when search is enabled */}
+            {showSearch && onSearchChange && (
```
lkostrowski: `const showSearch = Boolean(onSearchChange);`

so it should do with only `showSearch`

lkostrowski (approved):

src/attributes/components/AttributeValues/AttributeValues.tsx
mirekm: `true`

PR #6318, opened 2026-02-05
Fix for non shippable products in Product Doctor

mirekm:
Product availability diagnostics should skip shipping zone warnings for non-shippable products (digital goods, activation codes, etc.). Products with isShippingRequired: false on their product type will no longer see false positive warnings about missing shipping zones or unreachable warehouses via shipping.


src/products/components/ProductDoctor/utils/availabilityChecks.test.ts
```diff
@@ -0,0 +1,262 @@
...
+        Object.entries(values).forEach(([key, value]) => {
+          result = result.replace(new RegExp(`\\{${key}\\}`, "g"), String(value));
+        });
+      }
+
+      return result;
+    },
+  }) as unknown as IntlShape;
+
+const mockIntl = createMockIntl();
```
witoszekdev, line 37: suggestion: if we need to have checks in tests for proper formatting of placeholders, maybe we should just use the actual `react-intl` with a mocked provider?

i'm not sure if we need this though, since we are not making any checks for messages in the actual tests below if I understand correctly. In that case we can make this much simpler:


```suggestion
const mockIntl = {
  formatMessage: (descriptor: MessageDescriptor) => return descriptor?.defaultMessage ?? "";
} as unknown as IntlShape;
```

src/products/components/ProductDoctor/utils/availabilityChecks.ts
```diff
@@ -316,8 +321,12 @@ export function runAvailabilityChecks(
     }
   }
 
-  // Run shipping checks only if we have permission
-  if (!skipOptions?.skipShippingChecks) {
+  // Run shipping checks only if:
+  // 1. We have permission (skipShippingChecks is not set)
```
witoszekdev, line 333: suggestion: let's also add a comment that `skipShippingChecks` is used when user doesn't have permissions (in the function parameters definition)

src/products/components/ProductDoctor/utils/mapProductToDiagnosticData.test.ts
```diff
@@ -5,6 +5,58 @@ import { mapProductToDiagnosticData } from "./mapProductToDiagnosticData";
...
+    }>;
+  }>;
+}
+
+/**
+ * Factory function to create test product input data.
+ *
+ * Note: We use TestProductInput (a simplified type) because:
+ * 1. ProductInput is a complex generated GraphQL type with many required fields
+ * 2. mapProductToDiagnosticData only accesses specific fields we provide
```
witoszekdev: suggestion: if this function is accessing only specific fields we should change type there, instead of requiring entire GraphQL fragment.

This means we can re-use the `TestProductInput` declaration above for that function

src/products/components/ProductDoctor/utils/mapProductToDiagnosticData.test.ts
```diff
@@ -5,6 +5,58 @@ import { mapProductToDiagnosticData } from "./mapProductToDiagnosticData";
...
+ * 3. Tests intentionally provide partial data to verify edge case handling
```
witoszekdev: I'm not sure if I understand this - if Saleor types say that it **always** returns data, but then it doesn't it violates GraphQL schema; if we just don't provide values in test to keep it simple this is fine

src/products/components/ProductDoctor/utils/mapProductToDiagnosticData.test.ts
```diff
@@ -119,14 +188,47 @@ describe("mapProductToDiagnosticData", () => {
...
+      ...createTestProduct(),
+      productType: null,
+    } as unknown as ProductInput;
+
+    // Act
+    const result = mapProductToDiagnosticData(product);
+
+    // Assert - should default to true (safer for diagnostics - shows more warnings)
+    expect(result?.isShippingRequired).toBe(true);
+  });
```
witoszekdev: this is interesting - can this even happen that we have products with null `productType`? 🤔

src/products/components/ProductDoctor/utils/mapProductToDiagnosticData.ts
```diff
@@ -15,6 +15,8 @@ export function mapProductToDiagnosticData(
   return {
     id: product.id,
     name: product.name,
+    // Default to true (shippable) if not specified - safer for diagnostics
+    isShippingRequired: product.productType?.isShippingRequired ?? true,
```
witoszekdev: suggestion: maybe we should put a warning (e.g. "Cannot check if shipping is required") instead of having a silent failure? If this actually happens users might be confused why diagnostics show error if product is not shippable

mirekm: There is a warning if you have no required permissions to perform full diagnostics.

src/products/components/ProductDoctor/utils/mapProductToDiagnosticData.test.ts
mirekm: No, I don't think so. I removed the tests.

PR #6320, opened 2026-02-06
ENG-1161 - Update workflows to use shared slack notifications, instead of custom python script

witoszekdev:
Replaced old Python script to send Slack notifications with new custom shared github action.

Updated 1Password action to latest: 3.1.0

https://linear.app/saleor/issue/ENG-1161/refactor-workflows-in-saleor-dashboard-so-they-use-reusable-slack

### DIFF
--- a/.github/workflows/create-draft-release.yml
+++ b/.github/workflows/create-draft-release.yml
@@ -116,5 +116,6 @@ jobs:
     with:
       custom_title: "🚨 Draft Release Creation Failed for *${{ github.ref_name }}*"
       status: failure
+      mention_group_id: ${{ secrets.SLACK_DASHBOARD_GROUP_ID }}
     secrets:
       slack-webhook-url: ${{ needs.load-failure-secrets.outputs.slack-webhook-url }}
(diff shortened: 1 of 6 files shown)


.github/workflows/create-draft-release.yml
```diff
@@ -94,60 +94,27 @@ jobs:
...
-                    "text": "<${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View Failed Workflow Run>"
-                  }
-                }
-              ]
-            }
+  notify-failure:
+    if: failure()
+    needs: [create-draft-release, load-failure-secrets]
+    permissions: {}
+    uses: saleor/saleor-internal-actions/.github/workflows/notify-slack.yaml@dcebf56bd7d513dab52c0bb7c0a5cff98ecaadd1 # v1.8.0
```
NyanKiyoshi: The commit `dcebf56bd7d513dab52c0bb7c0a5cff98ecaadd1` doesn't exist (I get HTTP 404) and isn't associated with `v1.8.0`:

```suggestion
    uses: saleor/saleor-internal-actions/.github/workflows/notify-slack.yaml@b659ecfb3d2d7e6fc55823c8a0ddd8fc26fcc2a9 # v1.8.0
```

Where did that SHA come from? Is it some LLM hallucination or something else?

witoszekdev: I think I might have copy-pasted wrong SHA, not even from this repo 😅

PR #6327, opened 2026-02-10
Add filtering to assign modal dialog

witoszekdev:
Added filtering by "Model type" to Assign modal dialog.

Refactored `AssignContainerRows` so that it supports dialog with filters (generic ui used for all "Assign ..." dialogs except products) 

CleanShot 2026-02-11 at 16 01 36@2x
CleanShot 2026-02-11 at 16 01 32@2x
CleanShot 2026-02-11 at 16 01 27@2x


src/components/AssignAttributeValueDialog/AssignAttributeValueDialog.tsx
```diff
@@ -73,41 +63,35 @@ const getSingleOrMultipleDialogProps = (attribute: AttributeInput) => {
...
+    entityType,
+    pages,
+    products,
+    collections,
+    categories,
+    attribute,
+    initialConstraints,
+    onFilterChange,
+    ...rest
+  } = props;
```
lkostrowski, line 77: I guess not needed change, maybe we can add some static check or claude rule to pick one of the syntaxes so agents dont try to change that with no reason

PR #6337, opened 2026-02-12
add actionlint to check github actions

witoszekdev:
Added static checker for Github Actions definitions: https://github.com/rhysd/actionlint


NyanKiyoshi (changes requested):
Just use [pre-commit instead](https://github.com/rhysd/actionlint/blob/v1.7.10/docs/usage.md#pre-commit)? It removes all the need for custom installation scripts, checking hashes (`--freeze`), etc.

The `actionlint-docker` hook is the preferable one

We are re-inventing the wheel 😅

witoszekdev:
@NyanKiyoshi I know it's in pre-commit but... we also have TypeScript in pre-commit, and yet we also run it in CI to make sure code actually passes the test (after all you can use `--no-verify` 😄)

For now I marked it as optional (`|| true`) because we have linter issue to resolve, but ideally it should block merging a PR.

I can migrate it to Docker, but that would be a spearate step - we already use `pnpm run lint` which runs all linters in parallel (via special regex syntax which looks for all `lint:` scripts) so this approach should be faster imho

scripts/install-actionlint.bash
```diff
@@ -0,0 +1,48 @@
...
+case "$machine" in
+  x86_64)       arch=amd64 ;;
+  i?86)         arch=386 ;;
+  aarch64|arm64) arch=arm64 ;;
+  arm*)         arch=armv6 ;;
+  *) echo "Unsupported arch: $machine" >&2; exit 1 ;;
+esac
+
+file="actionlint_${VERSION}_${os}_${arch}.${ext}"
+url="https://github.com/rhysd/actionlint/releases/download/v${VERSION}/${file}"
```
NyanKiyoshi, line 34: Could we at least use the checksums to ensure the release doesn't mutate as they don't use immutable releases?

The checksums (they need to be stored raw in our repo in this file or file next to it):
```
16782c41f2af264db80f855ee5d09164ca98fc78edf3bcd0f46eecff279682ba  actionlint_1.7.10_darwin_amd64.tar.gz
004ca87b367b37f4d75c55ab6cf80f9b8c043adbfbd440f31c604d417939c442  actionlint_1.7.10_darwin_arm64.tar.gz
31ec5ee4deef04edf3b9eb3bac5243d785b5dccf70688e02a022b0723bfae80a  actionlint_1.7.10_freebsd_386.tar.gz
418461873e1f8eb7acc7524d70d4c33e99f9aab53da8625f4a3c50273612c930  actionlint_1.7.10_freebsd_amd64.tar.gz
b728a3c31f7bd37d3026a7cd87148da9ddf845b5d56213fcc410bb222523c7d0  actionlint_1.7.10_linux_386.tar.gz
f4c76b71db5755a713e6055cbb0857ed07e103e028bda117817660ebadb4386f  actionlint_1.7.10_linux_amd64.tar.gz
cd3dfe5f66887ec6b987752d8d9614e59fd22f39415c5ad9f28374623f41773a  actionlint_1.7.10_linux_arm64.tar.gz
7624bae3f6e41985f498de7b246a2334cd734b9f1a8f04a9748429cbcf99b750  actionlint_1.7.10_linux_armv6.tar.gz
1783f8a1dd59a67dae373e753a88592351e00128195bc9ddd2b836a0c31df71b  actionlint_1.7.10_windows_386.zip
283467f9d6202a8cb8c00ad8dd0ee4e685b71fb86a6a56c68fcbb9ae8ed91237  actionlint_1.7.10_windows_amd64.zip
0c7242bca2f0ee4672a00ce7fa3c7185311fb94cb9e1e0f6bb57d5456e421d80  actionlint_1.7.10_windows_arm64.zip
```

You can grep the correct line based on the file name that you generate (e.g., `actionlint_1.7.10_windows_amd64.zip`), then do `sha256sum -c -`

For example, let's say we have `checksums.txt` stored them, then we can do this:

```bash
( # (sub-shell as we do 'cd')
    cd "$tmpdir"
    if ! (grep -E "${file}\$" checksums.txt | sha256sum -c -); then
       echo "Downloaded file doesn't matched expected checksum. Aborting..." >&2
       exit 1
    fi
)
```

(assumes that the filename isn't changed, you could alter `curl` command with `-o "$file"` to make it consistent, it makes it a bit simpler as otherwise we would need intermediary variables)

scripts/install-actionlint.bash
```diff
@@ -0,0 +1,48 @@
...
+
+echo "Downloading actionlint v${VERSION} (${os}/${arch})..."
+mkdir -p "$TARGET_DIR"
+
+if [ "$os" = "windows" ]; then
+  tmpdir="$(mktemp -d)"
+  curl -sSL -o "$tmpdir/tmp.zip" "$url"
```
NyanKiyoshi, line 41: Missing the `--fail` flag (to fail on HTTP ≥ 400)
```suggestion
  curl -fsSL -o "$tmpdir/tmp.zip" "$url"
```

scripts/install-actionlint.bash
```diff
@@ -0,0 +1,48 @@
...
+  unzip -o "$tmpdir/tmp.zip" actionlint.exe -d "$TARGET_DIR"
+  rm -rf "$tmpdir"
+else
+  curl -sSL "$url" | tar xz -C "$TARGET_DIR" actionlint
```
NyanKiyoshi, line 45:
```suggestion
  curl -fsSL "$url" | tar xz -C "$TARGET_DIR" actionlint
```

PR #6338, opened 2026-02-13
Add missing required secrets to run app-smoke-test workflow from other repos

witoszekdev:
Continuation to make smoke tests run when called from this repo and from other repos (apps, adyen app):

- Added required secrets to be passed by external called (when called by external repository, e.g. apps monorepo, it cannot access secrets defined in saleor-dashbaord repo)
- Added loading for secrets when called inside saleor-dashboard repo from 1Password (previously these were in saleor-dashboard secrets, now we have single source of truth)
- Added ephemeral encryption key to pass JWT between runners (previously used a predefined static key from repo secrets, won't work with calling workflow from other repos)
- Hidden `ACCOUNTS` encrypted variable from being shown in logs

Workflow test run: [App smoke tests](https://github.com/saleor/saleor-dashboard/actions/runs/21998897995)

Linked PRs:
- https://github.com/saleor/saleor-app-payment-adyen/pull/1562
- https://github.com/saleor/apps/pull/2234


.github/workflows/app-smoke-tests.yml
```diff
@@ -117,10 +163,27 @@ jobs:
         with:
           repository: saleor/saleor-dashboard
 
+      - name: Load secrets from 1Password
+        # load secrets when NOT called from external repo
+        if: github.event_name == 'workflow_dispatch'
+        uses: 1password/load-secrets-action@581a835fb51b8e7ec56b71cf2ffddd7e68bb25e0 # v2.0.0
+        with:
+          export-env: true
```
NyanKiyoshi, line 171: Bad/risky practice: why not just use `outputs` instead of putting them in the environment variables? That way only intended steps are receiving the secrets rather than all of them

.github/workflows/app-smoke-tests.yml
```diff
@@ -59,13 +72,46 @@ jobs:
...
+          _E2E_PERMISSIONS_USERS_PASSWORD: ${{ secrets.E2E_PERMISSIONS_USERS_PASSWORD }}
+          _MAILPITURL: ${{ secrets.MAILPITURL }}
+        run: |
+          {
+            echo "STAGING_TOKEN=${_STAGING_TOKEN}"
+            echo "E2E_USER_NAME=${_E2E_USER_NAME}"
+            echo "E2E_USER_PASSWORD=${_E2E_USER_PASSWORD}"
+            echo "E2E_PERMISSIONS_USERS_PASSWORD=${_E2E_PERMISSIONS_USERS_PASSWORD}"
+            echo "MAILPITURL=${_MAILPITURL}"
+          } >> "$GITHUB_ENV"
```
NyanKiyoshi, line 106: Bad/risky practice: don't put secrets in the global environment variables, instead only share them to the steps that actually need the secrets (basically, undo the changes you made)

Otherwise the secrets are shared to all steps

.github/workflows/app-smoke-tests.yml
NyanKiyoshi: Discussed in messages, it's used as secrets cannot be easily shared on external invocations. We could do things like `${{ secrets.MAILPITURL || steps.1password.outputs.MAILPITURL }}` but it increases complexity and human error risk

PR #6423, opened 2026-03-16
Replace some moment.js usages

lkostrowski:


src/components/Date/Date.test.tsx
```diff
@@ -59,6 +68,116 @@ describe("Date", () => {
...
+
+    // Act
+    render(
+      // @ts-expect-error - legacy types
+      <ThemeProvider>
+        <TimezoneProvider value="UTC">
+          <Date date={testDate} />
+        </TimezoneProvider>
+      </ThemeProvider>,
+    );
```
witoszekdev, line 88: nitpick: we could use `wrapper` parameter for react-testing-library to re-use the providers

src/components/Date/getRelativeTime.test.ts
```diff
@@ -0,0 +1,239 @@
...
+      // Arrange
+      const date = "2024-01-15T12:00:00Z";
+      const now = toMs("2024-01-15T12:00:00Z");
+
+      // Act
+      const result = getRelativeTime(date, now, "en");
+
+      // Assert
+      expect(result).toBe("now");
+    });
```
witoszekdev: nitpick: we could use now in `0-3`s range, just to have a little bit nicer UI after you make changes there might be slight difference between fetching data after change and actually making change

src/components/Timeline/TimelineNote.test.tsx
```diff
@@ -76,7 +76,7 @@ describe("TimelineNote", () => {
     // Assert
     expect(screen.getByText("Test App")).toBeInTheDocument();
     expect(screen.getByText("Note")).toBeInTheDocument();
-    expect(screen.getByText("a few seconds ago")).toBeInTheDocument();
+    expect(screen.getByText("5 seconds ago")).toBeInTheDocument();
```
witoszekdev: nitpick: (from prev comment) we could change it to match exactly the previous implementation for _very_ recent changes

witoszekdev (dismissed):

src/components/Timeline/TimelineNote.test.tsx
lkostrowski: done

PR #6453, opened 2026-03-21
Promotions design update

mirekm:
First leg of unification and improvements in the discounting UI in Saleor Dashboard.

* Status added to the main Promotions listing 
* Redesign of the Promotion page
* Redesign of the Rules editor/modal

## After
image

image

image

## Before
image
image

image


src/discounts/components/DiscountDetailsPage/Title.tsx
```diff
@@ -0,0 +1,94 @@
+import { Pill } from "@dashboard/components/Pill";
+import { getPromotionStatus, type PromotionStatus } from "@dashboard/discounts/utils";
+import { type PromotionDetailsFragment, PromotionTypeEnum } from "@dashboard/graphql";
+import { type PillStatusType } from "@dashboard/misc";
+import { Box, Skeleton, Text } from "@saleor/macaw-ui-next";
+import { type IntlShape, useIntl } from "react-intl";
+
+function getRelativeTimeHint(
```
lkostrowski: add tests

src/discounts/components/DiscountRules/componenets/RuleForm/components/DiscountTypeSwitch/DiscountTypeSwitch.tsx
```diff
@@ -15,43 +18,74 @@ export const DiscountTypeSwitch = ({
...
+        return;
+      }
+
+      event.preventDefault();
+
+      if (selected === RewardValueTypeEnum.FIXED) {
+        setPercentage();
+      } else {
+        setFixed();
+      }
```
lkostrowski, line 47: I dont undersand
1. You ignore non-arrow events
2. On other type of event it sets values based on something else than keyboard

How is it suppose to work?

src/discounts/components/DiscountRules/componenets/RulesList/components/RuleSummary/utils.ts
```diff
@@ -55,3 +47,25 @@ export const conditionTypeToHue = (type: string, theme: DefaultTheme) => {
 export const hasNoRuleConditions = (rule: Rule) => {
   return !rule.conditions.length || rule.conditions.every(condition => !condition?.value?.length);
 };
+
+export const getConditionEntityUrl = (
```
lkostrowski, line 51: every util should be tested

src/discounts/utils.ts
```diff
@@ -104,3 +105,23 @@ export function sortRules(rules: Rule[]) {
...
 }
+
+export function getPromotionStatus(
+  startDate: string | null | undefined,
+  endDate: string | null | undefined,
+  now = new Date(),
+): PromotionStatus {
+  const nowTimestamp = now.getTime();
+  const startTimestamp = startDate ? new Date(startDate).getTime() : NaN;
+  const endTimestamp = endDate ? new Date(endDate).getTime() : NaN;
```
lkostrowski: you should try-catch `new Date()`, because in case of invalid string, it will not return `NaN` but throw in the runtime. Or best if you accept `Date` instead of `string` so this function is "protected" by typescript

lkostrowski (commented):
+ fix conflicts

src/discounts/utils.ts
```diff
@@ -104,3 +105,88 @@ export function sortRules(rules: Rule[]) {
 export function sortAPIRules(rules: PromotionRuleDetailsFragment[]) {
   return rules.sort(sortAlphabetically("name"));
 }
+
+/** ISO / API date strings or pre-parsed {@link Date}; invalid values yield NaN (never throws). */
+function toUtcMillis(value: string | Date | null | undefined): number {
```
lkostrowski: This is not best design choice because in js/ts `NaN` is type of number. So this function looks like return number but return NaN  in certain cases. On type/compiler layer you can't detect that and it can lead to silent throws

possible choices:
1. return `number | null` and replace`NaN` with null. Clients will have to add handling to null which is safe
2. throw and handle error

Also its bad design to accept `... | null | undefined`. You allow empty value just to throw. Instead, better to check these values in parent and function should only parse valid value

When function is called `toUtcMillis` it can technically accept `Date` only. But this is private function for this file, so you can also call it `normalizeToMillis` which will be more truthy (get date and near-date values, normalize, return in proper format or fallback/error)

src/discounts/utils.ts
```diff
@@ -104,3 +105,88 @@ export function sortRules(rules: Rule[]) {
...
+  } catch {
+    return NaN;
+  }
+}
+
+export function getPromotionStatus(
+  startDate: string | Date | null | undefined,
+  endDate: string | Date | null | undefined,
+  now = new Date(),
+): PromotionStatus {
```
lkostrowski: replace with named arguments (object destructuring). It's poor DX now because you have 3 positional arguments with the same types, so its easy to mix arguments order (eg `getPromotionStatus(now, start ,end)`, all `Date`).

src/discounts/utils.ts
```diff
@@ -104,3 +105,88 @@ export function sortRules(rules: Rule[]) {
...
+
+/**
+ * Relative time value/unit for scheduled (vs start) or finished (vs end) promotions.
+ * Active promotions have no reference date for this hint.
+ */
+export function getRelativePromotionTimeParts(
+  status: PromotionStatus,
+  startDate: string | Date | null | undefined,
+  endDate: string | Date | null | undefined,
+  now = new Date(),
```
lkostrowski: same, use destructuring in arguments

lkostrowski (approved):

PR #6490, opened 2026-04-08
strict-typed Scalars: JSON and JSONString.

lkostrowski:
Set strict-typed Scalars: JSON and JSONString. Previously Codegen generated `any` types, making them insecure in the codebase. Now they are `unknown` and `string`. Now it's explicit that JSON must be narrowed (e.g. with Zod schema) and JSONString must be first parsed.

For example - if type is `any`, Typescript allows any operation on it (eg `data.foo` will not be guarded). This leads to insecure data accessor and runtime errors

Now, JSON will be `unknown` - typescript requires narrowing type first (eg. `if "foo" in data then data.foo)
For JSONString it's `string` - typescript requires `JSON.parse(stringdata)` first


src/products/components/ProductUpdatePage/ProductUpdatePage.tsx
```diff
@@ -320,7 +320,8 @@ const ProductUpdatePage = ({
       const newProductDescription = productDescriptionField.value;
 
       // cache may be empty if editor was not used before sending event to app
-      const productDescriptionWithFallback = descriptionCache.current ?? product.description;
+      const productDescriptionWithFallback =
+        descriptionCache.current ?? (JSON.parse(product.description) as OutputData);
```
lkostrowski, line 324: it was a bug, but it's private/WIP api so it never surfaced

peelar (dismissed):

peelar (dismissed):

witoszekdev (dismissed):

wcislo-saleor (approved):

PR #6492, opened 2026-04-08
Update React Testing Library to 13.4 to support React@18

lkostrowski:
Prior to this change, we hade bunch of errors in tests with incompatible setup between RTL and React. 13.4 is only 1 version bump, but it adds support to React 18

I decided not to bump to newest versions, because its safer in general to update with small increments (peer deps changes → more risk)


testUtils/setup.ts
```diff
@@ -2,6 +2,8 @@ import "@testing-library/jest-dom";
 
 import { configure } from "@testing-library/react";
 
+(globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
```
witoszekdev: question: why do we need this?

witoszekdev (dismissed):

testUtils/setup.ts
lkostrowski: I think thats the intermediate step when upgrading, but I have tried without it and I see no difference, so I removed that

witoszekdev (approved):

PR #6493, opened 2026-04-08
[3.22] Fix bug with gifctard sending to customer

lkostrowski:
This is another fix of broken comboboxes, it adds extra handling around values to allow custom option


src/giftCards/GiftCardCreateDialog/GiftCardCustomerSelectField.tsx
```diff
@@ -20,44 +22,96 @@ export const GiftCardCustomerSelectField = ({
...
+  }, 500);
+
+  const options = useMemo(() => {
+    const opts = (customers || []).map(({ email, firstName, lastName }) => ({
+      value: email,
+      label: getFullName({ firstName, lastName }) || email,
+    }));
+    const trimmed = inputValue.trim();
+
+    if (trimmed && trimmed.includes("@")) {
```
witoszekdev: question: what are we supposed to check here? if this is a valid email? then perhaps we should use a better validation, with regex, not just checking `@` - since users cannot select `Use email: test@test` since it's not a valid domain

https://stackoverflow.com/questions/201323/how-can-i-validate-an-email-address-using-a-regular-expression

src/giftCards/GiftCardCreateDialog/GiftCardCustomerSelectField.tsx
```diff
@@ -20,44 +22,96 @@ export const GiftCardCustomerSelectField = ({
...
+        : null,
+    [selectedCustomer.email, selectedCustomer.name],
+  );
+
+  const handleSelect = (option: Option | null): void => {
+    if (!option) {
+      setSelectedCustomer({
+        email: "",
+        name: "",
+      });
```
witoszekdev, line 76: question: shouldn't this be a `null` if user didn't select a valid option?

lkostrowski: no, existing logic relies on empty strings from form

witoszekdev (approved):

PR #6504, opened 2026-04-14
Fix deleting server-side voucher codes

mirekm:
### Summary
* The dashboard incorrectly blocked deletion of saved voucher codes with a "Can't delete saved codes" tooltip, even though the backend supports it via voucherCodeBulkDelete
* Wired up the voucherCodeBulkDelete mutation so saved codes are actually deleted server-side, while draft (unsaved) codes continue to be removed from local state
* Added id to the VoucherCode GraphQL fragment (required by the mutation)
* Delete confirmation dialog now shows loading state during the mutation, displays error/success notifications with count of deleted codes, and uses red button styling consistent with other destructive actions in the dashboard


witoszekdev (dismissed):
lgtm but it looks like tests are failing due to import issue

mirekm:
> lgtm but it looks like tests are failing due to import issue

Fixed.

witoszekdev (approved):

PR #6514, opened 2026-04-15
build: split deploy to unstable main workflow

NyanKiyoshi:
Splits the build & deploy steps into two distinct jobs to reduce the blast radius and risk of malicious dependencies obtaining access to secrets.

## Scope of the change


lkostrowski (dismissed):

.github/workflows/deploy-master-staging.yaml
```diff
@@ -32,19 +33,28 @@ jobs:
...
+          # Explicitly disables cache to ensure cache is always disabled
+          cache: false
+
       - name: Setup Node
-        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
         with:
           node-version-file: "package.json"
-          cache: "pnpm"
+          # Cache disabled to prevent poisonings due to lack of cryptographic signatures
```
NyanKiyoshi: @lkostrowski what do you think?

```suggestion
          # Cache is intentionally disabled for security.
          #
          # GitHub Actions cache doesn't provide any trust boundary:
          # - Any workflow that can run in this repository (e.g., pull_request
          #   or push events) can populate cache entries (using any cache keys)
          # - A trusted workflow (like this build job) can then later restore
          #   that cache and execute or deploy malicious content
          #
          # This enables cache poisoning, meaning someone could inject
          # malicious NPM dependencies or malicious binaries (by replacing
          # the node or npm executable with a malicious one) then leading to
          # code execution during build or compromised deployments (e.g.,
          # implanting backdoors inside the JS or HTML files generated
          # during `pnpm run build`)
          #
          # Since this workflow deploys to AWS, it must not consume artifacts
          # from shared cache.
```

witoszekdev (approved):

.github/workflows/deploy-master-staging.yaml
```diff
@@ -3,13 +3,14 @@ on:
   push:
     branches:
       - main
```
stmpn, line 5: you are removing `- ci/staging/**` from `on.push.branches` - approving, with the assumption it's intentional.

stmpn (approved):

.github/workflows/deploy-master-staging.yaml
NyanKiyoshi: @stmpn yes, otherwise it would require to enable branch protections for `ci/staging/**` which would defeat the goal of having `ci/staging` 😄

PR #6530, opened 2026-04-20
Update dependency cruiser

lkostrowski:
dependency cruiser package was very old and `pnpm audit` reported several vulnerabilities in it's packages

This PR updates it


.dependency-cruiser.mjs
```diff
@@ -1,5 +1,5 @@
 /** @type {import('dependency-cruiser').IConfiguration} */
-module.exports = {
+export default {
```
lkostrowski, line 2: removed legacy module syntax. This also require extension to be `mjs`

pnpm-workspace.yaml
```diff
@@ -9,6 +9,10 @@ minimumReleaseAgeExclude:
 
 trustPolicy: no-downgrade
 
+trustPolicyExclude:
+  # Semver packages are not signed before 7.x. Check pnpm-lock which package is still using 6.x and plan upgrade entire tree.
+  - "semver@6.3.1"
```
lkostrowski, line 14: Must be added to ignored policies, because semver introduced signing after 7.x

witoszekdev (approved):

PR #6531, opened 2026-04-20
Update vulnerable protobuf and flatted packages

lkostrowski:


pnpm-lock.yaml
```diff
@@ -11397,10 +11399,10 @@ packages:
         integrity: sha512-KXUM8UEV+IK4JYWHNyxkPGDGbxeTEUHQv3POApfyTRN5eMcPFbY4cB0mDJr0LPelVvYPghmZDOCqfCIm9mYHtQ==,
       }
 
-  protobufjs@7.5.4:
+  protobufjs@8.0.0:
```
NyanKiyoshi: Should be either `>=7.5.5, <8`, or `>= 8.0.1` (per https://github.com/advisories/GHSA-xq3m-2v4x-88gg)

pnpm-workspace.yaml
```diff
@@ -44,6 +44,8 @@ overrides:
   "immutable@<3.8.3": ">=3.8.3"
   "flatted@<3.4.0": ">=3.4.0"
   "@tootallnate/once@<3.0.1": ">=3.0.1"
+  "protobufjs@<7.5.5": ">=7.5.5"
+  "flatted@<=3.4.1": ">=3.4.2"
```
witoszekdev, line 53: We already have override for flatted flatted@<3.4.0, so we should get rid of the old one

NyanKiyoshi (approved):

witoszekdev (approved):

PR #6537, opened 2026-04-24
pt-br update missing keys

nerijunior:
## Scope of the change

Added the missing translation keys from `defaultMessages` in comparison to `pt-BR.json`

1700 strings in total.

- [X] I confirm I added ripples for changes (see src/ripples) or my feature doesn't contain any user-facing changes

### DIFF
--- a/locale/pt-BR.json
+++ b/locale/pt-BR.json
@@ -1,7697 +1,13865 @@
 {
-    "PRXpBm": {
-        "context": "dialog header",
-        "string": "Cancelar Pedido"
-    },
-    "ll2dE6": {
-        "context": "PageTypeDeleteWarningDialog multiple assigned items description",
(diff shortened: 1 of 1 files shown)


lkostrowski:
While I understand this is generated by LLM, I also used LLM to verify that. Even not knowing Portugese, it's easy to recognize these strings are not correct and can't be accepted.

Maybe LLM hallucinated and confused keys? You can try to do it in smaller batches or parallel agents.

For example:
key `a/JR9Y`

English (default)

```
  "a/JR9Y": {
    "context": "sale type order discount",
    "string": "Sale"
  },
```

Contribution

```
  "a/JR9Y": {
    "context": "sale type order discount",
    "string": "{cardsAmount} cartões-presente emitidos"
  },
```

There are *many* cases like this, but this is easy to detect - the original value does not inject any variable into the string

PR #6542, opened 2026-04-28
Disallow to set MANAGE_APPS on apps in ui

lkostrowski:
In 3.24 we will hard-block `MANAGE_APPS` permission for Apps

However dashboard still shows this permission as available form "checkbox". As a part of deprecation of this behavior, let's hide id from the UI

Core change https://github.com/saleor/saleor/pull/19166


src/extensions/domain/app-manifest.test.ts
```diff
@@ -345,6 +345,98 @@ describe("App Manifest Schema", () => {
...
+        ],
+      });
+
+      // Act
+      const result = appManifestSchema.safeParse(invalidData);
+
+      // Assert
+      expect(result.success).toBe(false);
+
+      if (!result.success) {
```
wcislo-saleor, line 453: I'd image that if `expect(result.success).toBe(false);` passed then we don't need any `if` statements here. Just have another `except` directly.

Same for the other test(s) in this file.

wcislo-saleor (dismissed):

wcislo-saleor (dismissed):

wcislo-saleor (approved):

PR #6545, opened 2026-04-29
ProductDoctor: Add support for the new availability mode

mirekm:
Saleor 3.23 introduced `Shop.useLegacyShippingZoneStockAvailability`. New installations default to the new direct warehouse↔channel stock model (false); existing installations keep the legacy shipping-zone-intersection model (true). The Product Doctor was so far only supporting the legacy semantics, so on direct-mode shops it raised false positives ("warehouse not in shipping zone") and labelled shipping-zone misconfiguration as a purchase blocker - when in direct mode it only blocks shipping.

This PR makes the doctor mode-aware and gives users explicit context about which mode is active.

https://github.com/user-attachments/assets/964d38db-f7c0-4a00-9fc6-c6ec10db08b0

### DIFF
--- a/src/products/components/ProductDoctor/AvailabilityCard.test.tsx
+++ b/src/products/components/ProductDoctor/AvailabilityCard.test.tsx
@@ -1,0 +1,208 @@
+import Wrapper from "@test/wrapper";
+import { fireEvent, render, screen } from "@testing-library/react";
+
+import { AvailabilityCard } from "./AvailabilityCard";
+import { type AvailabilityIssue, type DiagnosticsResult } from "./utils/types";
+
+// usePublicApiVerification hits the live API; stub it for isolated UI tests.
+jest.mock("./hooks/usePublicApiVerification", () => ({
+  usePublicApiVerification: () => ({
+    verifyChannel: jest.fn(),
+    getChannelResult: () => undefined,
+    isVerifying: false,
+    results: new Map(),
+    lastVerified: null,
+  }),
+}));
+
+const baseDiagnostics = (overrides: Partial<DiagnosticsResult> = {}): DiagnosticsResult => ({
+  issues: [],
+  channelSummaries: [
+    {
+      id: "channel-1",
+      name: "Default Channel",
+      slug: "default-channel",
+      currencyCode: "USD",
+      isActive: true,
+      isPublished: true,
+      publishedAt: "2024-01-01T00:00:00Z",
+      isAvailableForPurchase: true,
+      availableForPurchaseAt: "2024-01-01T00:00:00Z",
+      visibleInListings: true,
+      warehouseCount: 1,
+      warehouseNames: ["Main Warehouse"],
+      shippingZoneCount: 1,
+      shippingZoneNames: ["US"],
+      countryCount: 1,
+    },
+  ],
+  hasErrors: false,
+  hasWarnings: false,
+  isLoading: false,
+  permissions: {
+    canViewChannelWarehouses: true,
+    canViewShippingZones: true,
+    missingPermissions: [],
+  },
+  useLegacyShippingZoneStockAvailability: true,
+  ...overrides,
+});
+
+describe("AvailabilityCard / StockAvailabilityModeIndicator", () => {
+  it("renders the legacy mode indicator when shop uses shipping-zone stock availability", () => {
+    // Arrange
+    const diagnostics = baseDiagnostics({ useLegacyShippingZoneStockAvailability: true });
+
+    // Act
+    render(<AvailabilityCard diagnostics={diagnostics} totalChannelsCount={1} />, {
+      wrapper: Wrapper,
+    });
+
+    // Assert
+    const indicator = screen.getByTestId("stock-availability-mode-indicator");
+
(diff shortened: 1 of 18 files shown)


src/products/components/ProductDoctor/primitives.tsx
```diff
@@ -160,7 +160,15 @@ export const IssueBadge = ({ count, type }: IssueBadgeProps) => {
...
   return (
-    <Box display="flex" alignItems="center" title={title} position="relative">
+    <Box
+      display="flex"
+      alignItems="center"
+      title={title}
+      position="relative"
+      data-test-id="channel-issue-badge"
+      data-test-count={count}
+      data-test-type={type}
```
lkostrowski: This is hacky, because this is not visible to the user. So you build false assumption that test is testing some UI, but actually is not. 

IMO you should add integration test (which renders react), can be e.g. in Storybook, or normally using react-testing-library. Test should look at the screen and look for the expected values. You can add `<span data-test-id="product-doctor-issue-badge-count">{count}</span>` etc, for easier query in test

And if count/type computation is not trivial, additionally you can add unit test for sole calculation logic, but this should not include react

src/products/components/ProductDoctor/AvailabilityChannelItem.tsx
```diff
@@ -412,14 +428,41 @@ interface IssueCalloutProps {
...
 
   return (
-    <Box display="flex" gap={2} alignItems="flex-start">
+    <Box
+      display="flex"
+      gap={2}
+      alignItems="flex-start"
+      data-test-id="availability-issue-callout"
+      data-test-issue-id={issue.id}
+      data-test-severity={issue.severity}
```
lkostrowski: The convention (which has build it support in testing libs) is `data-testid` and you build value accordingly:
`data-testid="product-doctor-issue-callout-issue-id"`. And in test you read what is rendered inside (see the other comment)

testing libs have dedicated methods for this

https://testing-library.com/docs/queries/bytestid/

src/products/components/ProductDoctor/AvailabilityChannelItem.tsx
```diff
@@ -64,8 +64,18 @@ export const AvailabilityChannelItem = ({
...
+  // "Blocking" issues (errors/warnings) drive the channel header state — they
+  // change the status label, dot color, and surface the issue badge. Info
+  // severity issues are advisories: they are still rendered in the body so
+  // users see the recommendation, but they should not promote the channel
+  // into a "has issues" visual state. This matches the top-level summary
+  // banner which counts only errors and warnings.
+  const blockingIssues = React.useMemo(
+    () => issues.filter(i => i.severity === "error" || i.severity === "warning"),
+    [issues],
+  );
```
lkostrowski: We can accept that as internal semantics, but I would expect "warning" *not* to be blocking. This is usually what warning means

src/products/components/ProductDoctor/hooks/useProductAvailabilityDiagnostics.ts
```diff
@@ -13,6 +13,14 @@ import {
   type ProductDiagnosticData,
 } from "../utils/types";
 
+/**
+ * Defaults to `true` (legacy mode) when the shop fragment in the diagnostics
+ * query hasn't loaded yet so we don't downgrade legacy behaviors prematurely
+ * on a fresh page load. Once the query resolves, the real value takes over
+ * and re-renders the diagnostics.
+ */
+const LEGACY_MODE_FALLBACK = true;
```
lkostrowski: you can read about optimistic rendering. Probably Apollo has it built-in. Instead of fallback on this level, you can provide default value to apollo so it will use it like it was cached

(non blocking, also not sure how its done in old apollo)

src/products/components/ProductDoctor/utils/availabilityChecks.ts
```diff
@@ -299,13 +428,17 @@ export function runAvailabilityChecks(
   channelData: ChannelDiagnosticData,
   channelListing: ProductDiagnosticData["channelListings"][0],
   intl: IntlShape,
-  skipOptions?: CheckSkipOptions,
+  options?: CheckOptions,
 ): AvailabilityIssue[] {
+  const useLegacyShippingZoneStockAvailability =
+    options?.useLegacyShippingZoneStockAvailability ?? true;
```
lkostrowski: let's re-use fallback value you assigned to variable in the other place, so fallback is the same and we can also update it to "false" one day (in one place)

src/products/components/ProductDoctor/AvailabilityCard.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+    // Act
+    render(<AvailabilityCard diagnostics={diagnostics} totalChannelsCount={1} />, {
+      wrapper: Wrapper,
+    });
+
+    // Assert
+    const indicator = screen.getByTestId("stock-availability-mode-indicator");
+
+    expect(indicator).toHaveAttribute("data-test-mode", "legacy");
```
lkostrowski: as other comments, you should default to assert what user sees, not hidden data on html tag. Its possible that inner component fails to render and your tests pass

lkostrowski (approved):

PR #6556, opened 2026-05-06
Improve models main view for easier access based on model type

mirekm:
Let's fix the generic lists UX to match the individual views and use-cases better.

Scope for the Models list view:
- Split models into separate tabs based on the type.
- Make the Create model button context/tab aware
- Hide Filters since there is only Model Type to filter against (we may bring it back for custom tabs with mode model types per single tab)

## After 
image

## Before
image


lkostrowski:
replaced by https://github.com/saleor/saleor-dashboard/pull/6568

PR #6564, opened 2026-05-12
Show line-level discounts for the confirmed orders

mirekm:
Adds a read-only price breakdown for order lines on confirmed orders. Click a discounted unit price or line total to open a modal explaining how the price was calculated - promotions, vouchers, manual discounts, order-wide discounts, and free-gift lines. Voucher rows deep-link to the voucher detail page.

The waterfall is derived from backend truth: `OrderLine.totalPrice` and `OrderLine.discounts[]` are anchors, and the order-level slice is recovered as `start − Σ line.discounts − end`. When multiple `Order.discounts[]` records contribute to a single line, the breakdown collapses them into one combined row with named contributors - Saleor doesn't surface the decomposition. Free-gift lines (`OrderLine.isGift = true`) get a dedicated "Free gift" row.

# New
https://github.com/user-attachments/assets/abf3e28f-f302-4e41-849d-2015874d94a5

# Old
orders-no-discount-details


src/products/components/OrderDiscountProviders/utils.test.ts
```diff
@@ -33,12 +33,14 @@ const lineBase = (
     __typename: "TaxedMoney" as const,
     gross: { __typename: "Money" as const, amount: unitGross, currency: "USD" },
     net: { __typename: "Money" as const, amount: unitGross, currency: "USD" },
+    tax: { __typename: "Money" as const, amount: 0, currency: "USD" },
```
lkostrowski, line 36: It's better to set mocks to be realistic. Price of line is not likely to be zero (unless you test specific case like gift).

Better set mocks to be real values, you will have more representative data set

src/orders/fixtures.ts
```diff
@@ -2292,6 +2423,11 @@ export const draftOrder = (placeholder: string): OrderDetailsWithMetadataFragmen
           amount: 68.2,
           currency: "USD",
         },
+        tax: {
+          __typename: "Money" as const,
+          amount: 0,
```
lkostrowski, line 2434: Its more likely that products ARE taxed, so better set mocks to include tax

src/orders/ripples/orderLinePriceBreakdown.tsx
```diff
@@ -0,0 +1,17 @@
+import { type Ripple } from "@dashboard/ripples/types";
+
+export const rippleOrderLinePriceBreakdown: Ripple = {
+  type: "feature",
+  // Persisted in users' localStorage to track dismissal — keep as-is even if
+  // the surrounding feature is renamed, so previously dismissed notifications
+  // stay dismissed.
```
lkostrowski: do we need this comment? isn't it in general what ripple is?

src/orders/components/OrderUnfulfilledProductsCard/OrderUnfulfilledProductsCard.tsx
```diff
@@ -66,6 +68,7 @@ const OrderUnfulfilledProductsCard = ({
             lines={toLineWithUnfulfilledQuantity(lines)}
             loading={loading}
             onOrderLineShowMetadata={onOrderLineShowMetadata}
+            onShowLinePricing={onShowLinePricing}
```
lkostrowski: Nitpick: Is callback name valid? On UI it looks like just a price. There is no button "show pricing". So in terms of the code/component, its `onPriceClicked`, isnt it?

src/components/Money/index.ts
```diff
@@ -9,6 +10,19 @@ export function subtractMoney(init: IMoney, ...args: IMoney[]): IMoney {
...
 
+/**
+ * Round a numerical amount to the minor-unit precision of its currency.
+ * USD/EUR → 2 decimals, JPY/KRW → 0 decimals, BHD/KWD → 3 decimals. Falls
+ * back to 2 decimals when the currency is unknown. Use this for any money
+ * arithmetic that needs to land on a representable amount; for display use
+ * `formatMoneyAmount` / `formatMoney` instead.
+ */
+export const roundMoneyAmount = (amount: number, currency: string): number => {
+  const factor = 10 ** getCurrencyDecimalPoints(currency);
```
lkostrowski: I think cents amount and fraction digits are available in saleor api for some time

lkostrowski: Also please check https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

itsn't it something that will work for you? (if you can use built-in api, you should)

src/orders/components/OrderLinePriceBreakdown/components/LinePriceWaterfallModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+import { AlertTriangle, X } from "lucide-react";
+import { FormattedMessage, useIntl } from "react-intl";
+
+import { messages } from "../messages";
+import { type LinePriceWaterfall, type PriceWarning } from "../utils/types";
+import { PriceWaterfallList } from "./PriceWaterfallList";
+
+interface LinePriceWaterfallModalProps {
+  open: boolean;
+  waterfall: LinePriceWaterfall | null;
```
lkostrowski: It's poor design. This component should render waterfall, so it doesnt make sense to accept MISSING waterfall, just to render null.

instead
1. require waterfall (remove null)
2. check null in parent

src/orders/components/OrderLinePriceBreakdown/components/PriceWaterfallList.tsx
```diff
@@ -0,0 +1,244 @@
...
+      to={voucherUrl(link.voucherId)}
+      title={intl.formatMessage(messages.linkVoucherTitle)}
+      className={styles.subtleLink}
+    >
+      {name}
+    </Link>
+  );
+}
+
+function getFactorDetail(factor: PriceFactor, intl: IntlShape): ReactNode {
```
lkostrowski, line 130: this and other utility function in this file contain a lot of logic. Should be tested. You can export them in this file, or move to the other one, but test.

to avoid mixing logic and ui, refator it to avoid rendering and return only pure calculations. If its hard to detach, test result ui

src/orders/fixtures.ts
mirekm: Agreed in principle. tax: 0 was the pre-existing convention in this fixture (gross == net everywhere), and the additions in this PR just complete the schema shape for the new fragment fields, they don't introduce the zero-tax assumption.

src/orders/ripples/orderLinePriceBreakdown.tsx
mirekm: No, removing.

src/orders/components/OrderUnfulfilledProductsCard/OrderUnfulfilledProductsCard.tsx
mirekm: A user pressing Enter on a focused price cell triggers it too. Renamed to `onShowLinePriceBreakdown`

lkostrowski: Ok thats make sense

src/components/Money/index.ts
mirekm: I introduced `fragmentMoneyWithFractionDigits` fragment.

src/orders/components/OrderLinePriceBreakdown/components/LinePriceWaterfallModal.tsx
mirekm: Fixed.

lkostrowski (approved):

PR #6565, opened 2026-05-13
fix: include port in external host check for redirect actions

### LINKED ISSUE #5549: [Bug]: Unable to redirect to app it self using appBridge Redirect action
### Description of the issue

When I tried to use the `Redirect` action on `appBridge` with the address `http://localhost:3000/page` while developing locally, the action failed, and the following message was displayed in the console:

```
Couldn't handle Redirect action properly, this should not happen
```

After checking the source code, I found that when the dashboard handles a redirect, it determines the correct way to execute the redirect. If it cannot find an appropriate method, it logs this message and fails.

https://github.com/saleor/saleor-dashboard/blob/0b33b71cda3d1a5f7724ce5c8de5a4102ec5a869/src/apps/components/AppFrame/appActionsHandler.ts#L120-L137

The correct behavior in this case would be to treat the URL `http://localhost:3000/page` as external, since the dashboard is running at `http://localhost:9000`. However, when we examine the logic behind the `isExternalHost` function, we see that it only compares hostnames:

https://github.com/saleor/saleor-dashboard/blob/0b33b71cda3d1a5f7724ce5c8de5a4102ec5a869/src/apps/components/AppFrame/appActionsHandler.ts#L28

And since:

```js
new URL('http://localhost:3000/page').hostname
// localhost

window.location.hostname
// localhost
```

It incorrectly determines that this is **not** an external URL, which prevents the redirect action from being executed, resulting in the failure.

I encountered this issue on `localhost`, but I believe the flawed logic for determining whether an address is external could also affect some production setups.

1. Run the Saleor Dashboard locally at `http://localhost:9000`.
2. In your app, attempt to trigger a Redirect action via appBridge with a URL such as `http://localhost:3000/page`.
3. Observe the browser console.

### What did you expect to happen?

The app should treat `http://localhost:3000/page` as an external URL (since it's on a different port) and perform a redirect.

### Additional information

_No response_

### Environment

Browser and version: Chrome 135.0.7049.95
OS and version: Linux
Dashboard version: v3.20.6
Core version: v3.20.33

MD-Mushfiqur123:
The \isExternalHost\ function only compared \hostname\, causing \localhost:3000\ and \localhost:9000\ to be treated as the same host. Changed to compare \.host\ which includes the port number.

Closes #5549

### DIFF
--- a/.changeset/plenty-mangos-pull.md
+++ b/.changeset/plenty-mangos-pull.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Fixed `appBridge` redirects to treat different ports on the same domain as different origins (for example `localhost:3000` vs `localhost:8000`).
--- a/src/extensions/views/ViewManifestExtension/components/AppFrame/appActionsHandler.ts
+++ b/src/extensions/views/ViewManifestExtension/components/AppFrame/appActionsHandler.ts
@@ -29,7 +29,7 @@ const createResponseStatus = (actionId: string, ok: boolean): DispatchResponseEv
     ok,
   },
 });
-const isExternalHost = (host: string) => new URL(host).hostname !== window.location.hostname;
+const isExternalHost = (host: string) => new URL(host).host !== window.location.host;
 const isLocalPath = (path: string) => path.startsWith("/");
 const useHandleNotificationAction = () => {
   const notify = useNotifier();


NyanKiyoshi (approved):
Thanks @MD-Mushfiqur123!

lkostrowski (approved):

PR #6568, opened 2026-05-15
Extract Models/Pages filters to tabs

lkostrowski:
https://github.com/user-attachments/assets/8f42c28c-863a-4081-ac51-e0f7600694c1

- Use current filter navigation
- Extract tabs that set filter-by-type
- Pre-fetch available types
- Pre-fetch initial filters count
- Store count in cache, refetch on tab visit
- "Create" button navigates to the contextually selected type
- "Back" button navigates back to selected tab
- "Search" looks up current selected type
- Counting only counts page size, show "+" if more than one page, to avoid heavy counting on the DB

Prior work: https://github.com/saleor/saleor-dashboard/pull/6556


mirekm (dismissed):

src/hooks/useLastCreatedEntityTypeStorage.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export type LastCreatedEntityType = "MODEL";
+
+const storageKey = (entityType: LastCreatedEntityType) => `lastCreatedEntityType:${entityType}`;
+
+export const useLastCreatedEntityTypeStorage = (entityType: LastCreatedEntityType) => {
+  const [value, setValue] = useLocalStorage<string | null>(storageKey(entityType), null);
+
+  return [value, setValue] as const;
+};
```
lkostrowski, line 11: persist last created type by user. E.g "Create model" pop up will read from this to set default value. So the assumption is that you prefer to have last used selected, than empty value

src/modeling/views/PageCreate.tsx
```diff
@@ -42,12 +44,29 @@ interface PageCreateProps {
...
+
+  if (params.get("action") === "create-page") {
+    params.delete("action");
+  }
+
+  const result = params.toString();
+
+  return result ? `?${result}` : "";
+};
+
```
lkostrowski, line 61: when navigating back from "Create model" to "model list", we want to navigate to
1. the same list state (stored in url) like active tab and sorting
2. but, we don't want modal/popup "create" to be open (which is also in url), so we strip it here

src/modeling/index.tsx
```diff
@@ -28,11 +27,7 @@ const PageList = () => {
     PageListUrlSortField.title,
   );
 
-  return (
-    <ConditionalPageFilterProvider locationSearch={location.search}>
-      <PageListComponent params={params} />
-    </ConditionalPageFilterProvider>
```
lkostrowski, line 34: we remove previous filter logic, because only filter it provided is product type. now, its moved to tabs

src/modeling/urls.ts
```diff
@@ -2,26 +2,19 @@ import { stringifyQs } from "@dashboard/utils/urls";
...
 
 export const modelingSection = "/models/";
 
 export const pageListPath = modelingSection;
-export type PageListUrlDialog =
-  | "publish"
-  | "unpublish"
-  | "remove"
-  | "create-page"
-  | TabActionDialog;
```
lkostrowski, line 24: part of legacy fitlers

wcislo-saleor (approved):

PR #6579, opened 2026-05-24
Customer details: add Overview card with per-currency KPIs

mirekm:
Replaces the legacy CustomerStats sidebar block with a new Overview card at the top of the customer detail page. The card shows Total orders (with the most recent order's date as a secondary line), Last login in the operator's configured time zone (with a GMT+X label), and one (Spent in recent orders, Avg order value) card pair per currency present in the customer's last 10 orders — so multi-currency customers see self-scoped per-currency stats instead of having the metric hidden.

Money values render with the amount inheriting the main KPI weight and the currency code shown as a smaller secondary annotation, matching the Pulse/Analytics KPI design language.

Introduces a shared, reusable KpiCard component (src/components/KpiCard) with proper keyboard accessibility — role="button", tabIndex, Enter/Space activation, and a focus-visible outline — when an onSelect handler is provided. Pure grouping logic lives in CustomerOverview/utils.ts so it can be unit-tested in isolation.

Also drops the redundant lastPlacedOrder alias from the CustomerDetails query in favour of reading the most recent order from the existing orders connection, and fetches up to 10 recent orders (previously 5) to back the new stats.

Tests: 6 cases for the grouping util (empty / single bucket / multi bucket / count-desc sort / alphabetical tie-breaker / single order), 8 cases for KpiCard (rendering variants, loading, click activation, keyboard activation, missing onSelect → no interactive semantics, active → aria-pressed).

image


src/components/KpiCard/KpiCard.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+import { KpiCard } from "./KpiCard";
+
+const Wrapper = ({ children }: { children: ReactNode }): JSX.Element => (
+  <ThemeProvider>{children}</ThemeProvider>
+);
+
+const renderCard = (ui: ReactElement): RenderResult => render(ui, { wrapper: Wrapper });
+
+describe("KpiCard", () => {
```
lkostrowski, line 13: imo better to move these to Storybook. They have similar UI & interaction testing, but gives you unified frontend and replay feature to debug

src/components/KpiCard/KpiCard.tsx
```diff
@@ -0,0 +1,158 @@
...
+  active?: boolean;
+  onSelect?: () => void;
+  dataTestId?: string;
+}
+
+const deltaTextColor = {
+  up: "success1",
+  down: "critical1",
+  neutral: "default2",
+} as const;
```
lkostrowski, line 30: try to make types strict - take variables from Macaw. With current implementation you dont have any autocomplete, you must debug typescript errors  to find available vars

lkostrowski (approved):

PR #6580, opened 2026-05-24
Customer details: redesign sidebar and main column

mirekm:
Replaces the legacy detail layout with a consolidated "Customer details" card (name, email, note in one place), a moved address book in the main column with hover-reveal copy-to-clipboard, and a new sidebar trio of Account Status, External Reference, and Gift Cards cards.

Sidebar:
- Account Status card shows isConfirmed / isActive as right-aligned pills. The destructive Activate / Deactivate actions live in the "More actions" cogs menu, each behind a two-step confirmation dialog with action-specific success toasts.
- External Reference card renders a read-only value with copy-to-clipboard, gated on the new externalReference field added to customerDetailsFragment.
- Customer Gift Cards card gets a header + "Issue new card" toolbar and a dashed Placeholder for empty states. The button is disabled until the customer query resolves so the dialog can't open without a target user. The data source switched from the top-level giftCards query (filtered by usedBy) to user.giftCards so newly-redeemed cards appear here.

Main column:
- Consolidated "General Information", "Contact Information", and "Notes" into one "Customer details" card.
- Moved the address book out of the sidebar. Each default-address block gets a hover-reveal copy button that serialises the address via a shared formatAddressForClipboard helper extracted from OrderCustomer (now reused there).
- Added a Staff badge inline with the customer name and a "Member since {date}" subtitle in the TopNav header.

Form ergonomics: the address create/update mutations now use disableErrorHandling so that field-level validation errors surface inline rather than as one toast per failed field.

The redesigned cards opt into the new section-header typography explicitly via size={6} fontWeight="medium" overrides. The companion branch refactor/section-header-typography flips the DashboardCard.Title default and removes those overrides repo-wide; this PR is self-contained so it can land independently.

## New

https://github.com/user-attachments/assets/ac696461-6c6e-4c7e-bac5-9162dfb9004e

## Old

https://github.com/user-attachments/assets/08507897-0854-466c-a362-bc4044b38894


src/components/AddressFormatter/formatForClipboard.ts
```diff
@@ -0,0 +1,21 @@
...
+  const lines = [
+    [address.firstName, address.lastName].filter(Boolean).join(" "),
+    address.companyName,
+    address.streetAddress1,
+    address.streetAddress2,
+    [address.postalCode, address.city].filter(Boolean).join(" "),
+    address.countryArea,
+    address.country?.country,
+    address.phone,
+  ].filter(Boolean);
```
lkostrowski, line 18: add tests

src/customers/components/CustomerDetailsPage/CustomerDetailsPage.tsx
```diff
@@ -88,32 +90,97 @@ const CustomerDetailsPage = ({
...
+                description: "customer detail cogs menu, deactivates the customer account",
+                id: "zP3Rb6",
+              })
+            : intl.formatMessage({
+                defaultMessage: "Activate user",
+                description: "customer detail cogs menu, activates a deactivated customer account",
+                id: "62Rs/K",
+              }),
+          onSelect: onActivateToggle,
+          testId: customer.isActive ? "deactivate-user" : "activate-user",
```
lkostrowski, line 109: how about permissions? can you open customer page without MANAGE_USERS (read only)? If yes, these probably should be disabled

lkostrowski (dismissed):

stmpn (dismissed):

src/customers/components/CustomerDetailsPage/CustomerDetailsPage.tsx
mirekm: It's ok for now as the current Dashboard routing gates `/customers` behind `MANAGE_USERS`, and the schema says user can be queried by `MANAGE_STAFF`, `MANAGE_USERS`, or `MANAGE_ORDERS`.

While Dashboard’s customer section is now stricter than API we should fix this in another PR by giving access to `/customers` to roles that can query (and make the Customer read-only for them).

lkostrowski (approved):

PR #6583, opened 2026-05-25
Unify channel component

mirekm:
Introduce new sales channel component. It also now appears next to the order date in the page header instead of a separate sidebar card. Click the channel name to see all orders for that channel.

## New
image

## Old
image


stmpn (dismissed):

.changeset/clear-tables-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Orders: the sales channel now appears next to the order date in the page header instead of a separate sidebar card. Click the channel name to see all orders for that channel.
```
lkostrowski, line 5: worth to add ripple, this change introduces confusion for all users looking for channel in the previously place

src/components/Channel/Channel.tsx
```diff
@@ -0,0 +1,125 @@
+import { Box, Skeleton, Text } from "@saleor/macaw-ui-next";
+import { Radio } from "lucide-react";
+import { useIntl } from "react-intl";
+import { Link as RouterLink } from "react-router-dom";
+
+import styles from "./Channel.module.css";
+import { messages } from "./messages";
+
+type TextSize = 1 | 2 | 3 | 4 | 5 | 6 | 7;
+type TextColor = "default1" | "default2" | "critical1";
```
lkostrowski: these are vars from Macaw, why redeclaring them?

src/components/Channel/Channel.tsx
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Make the channel clickable by passing a target URL. The whole pill becomes
+   * a router link styled to inherit the secondary color (no underline, no blue).
+   * When omitted, the component renders as plain (non-interactive) text.
+   *
+   * Pick the destination per use-case:
+   * - `orderListUrlWithChannel(order.channel)` to scope the orders list
+   * - `channelUrl(channel.id)` to jump to channel settings (admin only)
+   */
+  href?: string;
```
lkostrowski: maybe make this logic inside this component? Every time you will use it, you must enter implementation, read comment to manually construct url... Abstraction breaks

I'd presonally to it this way

```
export const Channel = () => {} // existing component

export const ClickableChannel = (props: Omit<ChannelProps, "href">) => <Channel href={build url here} />
```

ClickableChannel can use "permissions" hook to get if user can access it and build the link

src/components/Channel/Channel.tsx
```diff
@@ -0,0 +1,125 @@
...
+  1: 12,
+  2: 14,
+  3: 14,
+  4: 16,
+  5: 18,
+  6: 20,
+  7: 24,
+};
+
+export const Channel = ({
```
lkostrowski: `ChannelDisplay` would be better name. Channel is very generic

lkostrowski (approved):

PR #6597, opened 2026-05-31
Fix highlighting Drafts in main nav when viewing draft order details

mirekm:
When viewing a draft order details page, the main navigation currently highlights the "Orders" section instead of "Drafts". This is because the logic for determining active menu items does not account for the fact that draft orders also use the same URL pattern as regular orders. This PR introduces a context to track which section of the orders menu should be considered active, and updates the menu item active state logic to use this context when determining if an order-related menu item should be highlighted.

An alternative would be introducing `/orders/drafts/:id` routing (breaking change).


lkostrowski:
@mirekm I think this is quite hacky. If we treat drafts separately, our resource URL should also show draft order as a separate navigation path. I would try this
1. Separate path `orders/ID`, `orders/drafts/ID`, new `drafts/ID` is canonical
2. `useEffect` on order page, add silent url `replace()` to resolve canonical path (this is only needed if someone has bookmarked old path, normally would be redirected from `drafts/` list to `/drafts/id`
3. Navigation should be fixed now automatically

lkostrowski (dismissed):

mirekm:
> @mirekm I think this is quite hacky. If we treat drafts separately, our resource URL should also show draft order as a separate navigation path. I would try this
> 
> 1. Separate path `orders/ID`, `orders/drafts/ID`, new `drafts/ID` is canonical
> 2. `useEffect` on order page, add silent url `replace()` to resolve canonical path (this is only needed if someone has bookmarked old path, normally would be redirected from `drafts/` list to `/drafts/id`
> 3. Navigation should be fixed now automatically

Yes, I think it's super hacky. I will try your approach.

lkostrowski (approved):

PR #6599, opened 2026-05-31
Let app widgets size their iframe to content height

mirekm:
Detail-page sidebar widgets previously rendered in a fixed 200px box. They now start at a sensible default and follow the height an embedded app reports via the `saleor:widget:resize` postMessage, so widgets flow naturally one after another.

- Add `useWidgetIframeAutoHeight` hook (with default height fallback and a max clamp to guard against runaway/hostile heights) plus unit tests
- Gate auto-height on an explicit AppFrame `autoHeight` prop instead of overloading target === "WIDGET" (which also covers the full-height Home page widget)
- Extract shared useNodeRef hook to avoid duplicating the ref + state pattern


lkostrowski (dismissed):

.changeset/widget-auto-height.md
```diff
@@ -0,0 +1,7 @@
+---
+"saleor-dashboard": patch
+---
+
+Detail-page app widgets in the sidebar no longer sit in a fixed-height box. Apps report content height via the `widgetResize` App Bridge action (`actions.WidgetResize` in `@saleor/app-sdk` ≥ 1.9.0); the Dashboard resizes the iframe and responds with `ok`. GET widgets use `useAppActions`; POST widgets handle the same action in `useWidgetIframeAutoHeight`. App developers can adopt `useWidgetAutoResize` / `reportWidgetHeightFromElement` ([saleor/app-sdk#507](https://github.com/saleor/app-sdk/pull/507)). Apps that have not adopted this yet keep the previous default height.
+
+Bumps `@saleor/app-sdk` to `1.9.0-rc.1` until `1.9.0` is published.
```
lkostrowski: remember to update once we release

lkostrowski (approved):

PR #6607, opened 2026-06-02
Deprecation notice on the frontend

lkostrowski:
Dashboard will show red warning 
CleanShot 2026-06-02 at 16 15 25


peelar (dismissed):

.nvmrc
wcislo-saleor, line 1: Intended change?

.changeset/deprecated-saleor-version-banner.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Added a hardcoded deprecation banner in the sidebar informing users that this Saleor version is deprecated and will be automatically upgraded on June 3, 2026.
```
wcislo-saleor: This date does not match date in src/components/Sidebar/DeprecationBanner/DeprecationBanner.tsx.

wcislo-saleor (dismissed):

.nvmrc
lkostrowski: Technically yes, without that we can't install dependencies (they require v18), but on the other hand, we may have looser rules on CICD. I will revert just in case of breaking

wcislo-saleor (approved):

peelar (approved):

PR #6647, opened 2026-06-17
Add refreshEntity app-bridge action to refresh host entity page

lkostrowski:
Apps embedded as widgets or popups on an entity detail page can now emit a `refreshEntity` action to request the host page to silently refetch its entity in the background, without a full page reload.

- Add a global entity-refresh registry (Jotai atom) with useRegisterEntityRefresh / useTriggerEntityRefresh hooks. Single active-page slot, ref-backed wrapper to avoid stale closures, and token-guarded cleanup to avoid clobbering during route transitions.
- Handle the action in useAppActions/appActionsHandler (covers popups and GET-AppFrame widgets). Fire-and-forget: ack immediately, no-op when no entity page is registered.
- Register the refresh function on every extension-bearing entity detail page: orders, draft orders, products, gift cards, customers, collections, vouchers, categories, promotions, models, model types and menus.

The wire type is handled via a raw string literal until it is published in @saleor/app-sdk. POST-method widgets are not wired yet.

## Scope of the change


przlada (approved):

xseignard:
nice! thanks! 🙏

PR #6670, opened 2026-06-24
ci: diff-based dependency audit + nightly main audit

lkostrowski:
`pnpm audit` on a PR was red whenever `main` already had a vulnerable (usually transitive) dependency, so every PR looked broken even when it changed nothing security-relevant.

- Replace the `check-lock` job with `audit-lock`, which audits both the PR and `main` and fails ONLY on advisories the PR introduces. Renders the result to the job summary and posts a sticky PR comment (same-repo PRs) when new vulnerabilities are added. Fails open on audit infra errors.
- Add `nightly-audit.yml` (04:30 UTC + workflow_dispatch) that audits `main` and fails red on any moderate+ advisory, so pre-existing vulnerabilities stay visible. Fails closed as the backstop.
- Add `scripts/audit-diff.ts` (pure functions + CLI, run via Node TS type stripping) and `node:assert` tests wired into `test:lint-rules`.

## Scope of the change

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,8 +13,14 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  check-lock:
+  # Runs `pnpm audit` on both the PR and `main`, then fails ONLY on advisories
+  # the PR introduces. Pre-existing vulnerabilities on `main` no longer turn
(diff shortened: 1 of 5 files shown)


lkostrowski:
installed socket.dev instead

PR #6680, opened 2026-06-28
Fix product numeric filter zero values

Max-Reisinger:
## Scope of the change

Fixes product numeric attribute filter parsing so zero values are preserved instead of being treated as empty.

Before this change, a numeric filter value of `0` could be converted to `undefined`. Now ranges starting or ending at 0 remain applied.

Added focused unit coverage for zero-only and zero-to-positive numeric ranges.

## Reproduction steps

1. Open the products list with a numeric attribute filter in the URL.
2. Use a range where one side is `0`, for example `numeric-attributes[<attribute-slug>]=0` or `numeric-attributes[<attribute-slug>]=0&numeric-attributes[<attribute-slug>]=10`.
3. Before this change, parsing the URL filter dropped the zero because `0 || undefined` resolved to `undefined`, so the generated `valuesRange` missed the `0` bound.
4. With this change, the same URL filter is parsed as `valuesRange: { gte: 0, lte: 0 }` for a zero-only filter, or `valuesRange: { gte: 0, lte: 10 }` for a zero-to-positive range.

## Testing

- `./node_modules/.bin/jest --silent src/products/views/ProductList/filters.test.ts`

- [x] I confirm I added ripples for changes or my feature doesn't contain any user-facing changes

### DIFF
--- a/.changeset/product-filter-zero-values.md
+++ b/.changeset/product-filter-zero-values.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Preserve zero values in product numeric attribute filters, so ranges starting or ending at 0 remain applied.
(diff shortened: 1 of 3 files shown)


mirekm (commented):
Thanks for the PR. Can you share reproduction steps?

Max-Reisinger:
> 感谢您的PR。能否分享一下复现步骤？
I’ve added the reproduction steps and verification commands below the description.

PR #6681, opened 2026-06-29
Fix numeric variant attributes in variants datagrid

Max-Reisinger:
## What

Fixes numeric variant attributes in the variants datagrid.

Numeric variant attributes are now:

- available as supported datagrid attribute columns
- rendered with the existing number cell
- submitted as the expected `numeric` attribute input

## Why

The variants datagrid previously only supported dropdown and plain text variant attributes. As a result, numeric variant attributes could not be edited from the grid, which prevented users from filling required numeric variant attributes in that flow.

## Testing

- `corepack pnpm run lint`
- `corepack pnpm run test:quiet products/components/ProductVariants/datagrid.test.ts`
- `corepack pnpm run test:quiet products/views/ProductUpdate/handlers/data/attributes.test.ts`
- `corepack pnpm run check-types`
- `corepack pnpm run knip`

### DIFF
--- a/src/products/components/ProductVariants/utils.tsx
+++ b/src/products/components/ProductVariants/utils.tsx
@@ -11,7 +11,11 @@ import { emptyDropdownCellValue } from "@dashboard/components/Datagrid/customCel
 import { numberCellEmptyValue } from "@dashboard/components/Datagrid/customCells/NumberCell";
 import { type DatagridChange } from "@dashboard/components/Datagrid/hooks/useDatagridChange";
 import { type AvailableColumn } from "@dashboard/components/Datagrid/types";
-import { type ProductDetailsVariantFragment } from "@dashboard/graphql";
+import {
+  AttributeInputTypeEnum,
+  type ProductDetailsVariantFragment,
+  type VariantAttributeFragment,
+} from "@dashboard/graphql";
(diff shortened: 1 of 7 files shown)


src/products/components/ProductVariants/utils.tsx
```diff
@@ -151,19 +157,35 @@ export function getData({
...
     });
   }
 }
+
+function getNumericAttributeValue(value: string | null | undefined) {
+  if (!value) {
+    return numberCellEmptyValue;
+  }
+
+  return Number(value);
```
mirekm: Minor: NaN can leak to the API.

``` typescript
const parsed = Number(value);
return Number.isFinite(parsed) ? parsed : numberCellEmptyValue;
```

mirekm (commented):
I left one minor comment and we will be able to merge this bug fix. Thanks for your first contribution @jiuyige!

src/products/components/ProductVariants/utils.tsx
Max-Reisinger: Thank you so much for letting my changes be included in your project.  @mirekm 
I updated the numeric parsing to return the empty number cell value when the parsed value is not finite.

PR #6750, opened 2026-07-17
Fix secondary text contrast

mirekm:
* Improve secondary text legibility by raising `text.default2` contrast from ~4.1:1 → ~5.7:1 on light backgrounds (helper copy and subtitles), clearing WCAG AA (4.5:1) for normal text.

* Dark theme also nudged (~5.3:1 → ~7.3:1).

### DIFF
--- a/.changeset/secondary-text-contrast.md
+++ b/.changeset/secondary-text-contrast.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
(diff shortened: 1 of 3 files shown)


lkostrowski (changes requested):
this should be done in Macaw repository, not only because we own it and it belongs there, but also because apps will not have this change reflected

PR #6787, opened 2026-07-30
Refactor: split legacy payments and modern transactions paths on order

lkostrowski:


lkostrowski:
@mirekm this PR contains what we have discussed - large refactor that moves "ifs" higher in Order view. Now Order page decides early if this is legacy payments view or Transactions and redners accordingly. Added storybook for visual regression. It can help us maintain the UI but we probably need some more manual testing

PR #6796, opened 2026-08-02
🌐 Channels Update

mirekm:


szczecha:
The helper text is inconsistently positioned across the dropdowns.image

szczecha:
Allow using 0 as the product price 
image

szczecha:
View published products / View unpublished products
1. Consider opening the product list in a new tab.
2. For a copied channel, after assigning products, "View products" opens the product list filtered by the original channel instead of the copied one.

Steps to reproduce:

Go to default-channel.
Duplicate the channel (slug: default-channel-copy).
Assign a product to the duplicated channel.
Click View unpublished products.

Actual result:
The product list opens with the default-channel filter applied instead of default-channel-copy.

Expected result:
The product list should open with the default-channel-copy filter applied.

https://github.com/user-attachments/assets/599d684c-53ec-455a-aa84-9ebc7c89f53e

mirekm:
Thanks, all great findings @szczecha! Fixed.

mirekm:
Merged with #6812

PR #6806, opened 2026-08-04
Collection and Categories update

mirekm:


szczecha:
**Unsaved changes state is not cleared after saving scheduled channel availability**

Steps to reproduce:

Open a Collection.
Go to Channel availability.
Select Schedule for later and configure the schedule.
Click Save.
Wait for the success toast to appear.
Try to navigate to another page (e.g. Products).

Actual result:
The "Leave without saving changes?" dialog is displayed, even though the scheduled channel availability was successfully saved. Additionally, the page still indicates unsaved changes—the "Unsaved changes: channel availability" message remains visible next to the Back and Save buttons.

Expected result:
After a successful save, the unsaved changes state should be cleared, and navigating away should not trigger the "Leave without saving changes?" dialog.

szczecha:
Screenshot 2026-08-06 at 08 02 02
Suggestion: Consider using "Live in 2 channels" instead of "Both channels" for consistency with labels like "Live in 3 channels" and "Live in 2 of 3 channels". Showing the actual number of channels makes the status more consistent and easier to scan.

mirekm:
Fixed.

mirekm:
Merged with #6812

PR #6810, opened 2026-08-05
Gift Card details update

mirekm:


szczecha:
**1. No success/error toast after saving Gift Card settings**

Steps to reproduce:

Go to the Gift Cards list.
Click the gear icon.
Open Settings.
Change the Expiration period.
Save the changes.

Actual result:

The GraphQL mutation is sent successfully.
No success or error toast is displayed.
After clicking Back, the "Leave without saving changes?" dialog appears, even though the changes were already saved.

Expected result:

A success (or error) toast should be displayed after saving.
After a successful save, navigating back should not trigger the "Leave without saving changes?" dialog.

**2. Gift Card settings are accessible from an individual Gift Card**

The Gift Card Settings page can be accessed from the details of a single Gift Card. This may be confusing because the settings (such as the expiration period) apply globally to all Gift Cards, not just the currently viewed one.

mirekm:
Fixed.

mirekm:
Merged with #6812

PR #6811, opened 2026-08-06
Vouchers update

mirekm:


szczecha:
Regression: Percentage voucher value cannot be set independently per channel

When configuring a percentage voucher, the same discount percentage is currently applied to all channel listings.

Actual result:
All channel listings use the same percentage discount value.

Expected result:
Each channel listing should allow its own percentage discount value, so different discount percentages can be configured per channel.

image

szczecha:
Suggestion: Clarify when the Redemptions section is applicable

The Redemptions section is only relevant when a voucher has a `usageLimit` configured, but this is not clear from the UI.

I would suggest either:

adding information in the Redemptions section that these statistics are available/applicable only when a usage limit is set, or
hiding the Redemptions section entirely when usageLimit is not set.


image

szczecha:
Voucher setup checklist is never shown – is this intentional?

When creating a new voucher, the setup checklist is not displayed at any point.

As a user creating a voucher for the first time, it's not clear which fields are required. For example, after entering the voucher name, the Save button remains disabled, but there is no indication of what else needs to be configured before the voucher can be saved.

This forces the user to figure out which required settings are missing.

Suggestion:
Consider displaying the setup checklist for a newly created voucher, or otherwise clearly indicate which required fields/settings still need to be completed before Save becomes available.

mirekm:
Thanks - great findings again. All fixed.

mirekm:
Merged with #6812

PR #6814, opened 2026-08-08
Fix fulfillment warehouse line scope

### LINKED ISSUE #6813: [Bug]: Selecting a fulfillment warehouse hides other order lines
## Description of the issue

When an order has multiple unfulfilled lines and a warehouse must be selected manually, choosing a warehouse for one line unexpectedly switches the whole fulfillment form into single-line mode.

The remaining order lines disappear and the page shows the hint that the fulfillment was opened from the line matrix, even though the user opened the regular order-level **Fulfill** flow.

This prevents staff from creating one fulfillment containing all order lines with one tracking number. The practical workaround is to create separate fulfillments, which can also result in multiple customer fulfillment notifications.

## Steps to reproduce the problem

Precondition: an order has at least two unfulfilled shippable lines, and at least one line does not have a warehouse preselected (for example, the variant has no allocation/default stock that Dashboard can use).

1. Open the order details page.
2. Click the regular order-level **Fulfill** action.
3. Verify that all unfulfilled order lines are visible.
4. Open **Change warehouse** for one of the lines.
5. Select a warehouse and close/confirm the picker.

Actual result:

- Only the selected order line remains in the fulfillment form.
- The page displays `Opened from the line matrix for ...`.
- Submitting creates a fulfillment for that one line instead of one fulfillment for all originally visible lines.

## What did you expect to happen?

Selecting a warehouse for a line should only update that line's warehouse. It should not change the scope of the fulfillment form.

All originally visible order lines should remain available so Dashboard can send one `orderFulfill` mutation containing multiple lines and one tracking number.

The intentional single-line flow opened from Line Matrix must continue to work independently.

## Additional information

The issue is caused by `lineId` having two unrelated responsibilities:

1. It is used as the intentional single-line focus from Line Matrix. `linesToFulfill` filters the form whenever `params.lineId` exists:
   https://github.com/saleor/saleor-dashboard/blob/main/src/orders/components/OrderFulfillPage/OrderFulfillPage.tsx#L94-L102
2. The warehouse picker also writes its target line into the same `lineId` URL parameter:
   https://github.com/saleor/saleor-dashboard/blob/main/src/orders/components/OrderFulfillPage/OrderFulfillPage.tsx#L265-L269
3. The warehouse dialog resolves and updates its line through that same parameter:
   https://github.com/saleor/saleor-dashboard/blob/main/src/orders/components/OrderFulfillPage/OrderFulfillPage.tsx#L179-L182
   https://github.com/saleor/saleor-dashboard/blob/main/src/orders/components/OrderFulfillPage/OrderFulfillPage.tsx#L348-L365
4. `OrderFulfill` does not pass any warehouse-related `fieldsToClear` to `createDialogActionHandlers`, so `lineId` and `warehouseId` remain in the URL after the modal closes:
   https://github.com/saleor/saleor-dashboard/blob/main/src/orders/views/OrderFulfill/OrderFulfill.tsx#L29-L36
   https://github.com/saleor/saleor-dashboard/blob/main/src/utils/handlers/dialogActionHandlers.ts#L25-L38

A robust fix is to give the warehouse picker its own URL parameter, for example `warehouseLineId`:

- Add `warehouseLineId` to `OrderFulfillUrlFiltersType`.
- Pass `warehouseLineId: line.id` when opening `change-warehouse`.
- Use `params.warehouseLineId` to resolve/update the warehouse dialog line.
- Clear `warehouseLineId` and `warehouseId` when closing the dialog.
- Keep `params.lineId` exclusively for the intentional Line Matrix single-line flow.

I verified this approach in a fork with regression tests covering both cases:

- opening the warehouse picker keeps all three fulfillment lines visible;
- a real `lineId` from Line Matrix still limits the form to the selected line.

## Environment

Browser and version: Chromium-based browser (Brave; exact version unavailable)
OS and version: Windows (exact version unavailable)
Dashboard version: reproduced on 3.23.19; the affected code is still present on `main` at 3.23.22
Core version: 3.23.22

offx366:
## Scope of the change

Fixes #6813.

Selecting a warehouse for a fulfillment line reused `lineId`, which is also the URL state for the intentional single-line flow opened from Line Matrix. As a result, opening the warehouse picker filtered the form down to that line and hid the remaining order lines.

This change:
- introduces `warehouseLineId` for warehouse-picker state;
- keeps `lineId` exclusively for intentional Line Matrix focus;
- clears `warehouseLineId` and `warehouseId` when the picker closes;
- adds regression coverage for multi-line fulfillment, Line Matrix focus, and dialog cleanup.

This preserves one order-level fulfillment containing multiple lines, one tracking number, and one customer notification.

## Verification

- `pnpm run lint` (0 errors)
- `pnpm run test:quiet src/orders/components/OrderFulfillPage/OrderFulfillPage.test.tsx src/utils/handlers/dialogActionHandlers.test.ts src/orders/urls.test.ts` (19/19)
- `pnpm run check-types` (3848 strict files passed)
- `pnpm run knip` (exit 0; existing repository warnings only)

- [x] This is a bug fix to existing behavior and does not require a ripple.
- [x] No new analytics event is required for this fix.

### DIFF
--- a/.changeset/fair-fulfillment-warehouse.md
+++ b/.changeset/fair-fulfillment-warehouse.md
@@ -1,0 +1,5 @@
+---
+"saleor-dashboard": patch
+---
+
+Keep every order line in a fulfillment when selecting warehouses, so one tracking number and one customer notification can cover the whole shipment.
--- a/src/orders/components/OrderFulfillPage/OrderFulfillPage.test.tsx
+++ b/src/orders/components/OrderFulfillPage/OrderFulfillPage.test.tsx
@@ -1,0 +1,96 @@
+import { type ConfirmButtonTransitionState } from "@dashboard/components/ConfirmButton";
+import { SavebarRefProvider } from "@dashboard/components/Savebar/SavebarRefContext";
+import Wrapper from "@test/wrapper";
+import { fireEvent, render, screen } from "@testing-library/react";
+import { MemoryRouter } from "react-router-dom";
+
+import { orderToFulfill } from "./fixtures";
+import OrderFulfillPage from "./OrderFulfillPage";
+
+jest.mock(
+  "@dashboard/orders/components/OrderChangeWarehouseDialog/OrderChangeWarehouseDialog",
+  () => ({
+    OrderChangeWarehouseDialog: ({ line }: { line: { productName: string } }) => (
+      <div data-test-id="change-warehouse-dialog">{line.productName}</div>
+    ),
+  }),
+);
+
+const defaultProps = {
+  loading: false,
+  errors: [],
+  order: orderToFulfill,
+  saveButtonBar: "default" as ConfirmButtonTransitionState,
+  onSubmit: jest.fn(),
+  openModal: jest.fn(),
+  closeModal: jest.fn(),
+};
+
+const renderPage = (params = {}) =>
+  render(
+    <MemoryRouter>
+      <Wrapper>
+        <SavebarRefProvider>
+          <OrderFulfillPage {...defaultProps} params={params} />
+        </SavebarRefProvider>
+      </Wrapper>
+    </MemoryRouter>,
+  );
+
+describe("OrderFulfillPage", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("keeps every order line visible while changing a warehouse", () => {
+    // Arrange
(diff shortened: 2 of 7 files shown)


mirekm (approved):

mirekm:
Thanks @offx366! The diagnosis is correct. I rebased it to the current main, and approved.

lkostrowski (approved):

PR #6858, opened 2026-08-22
Fix Navigator (Ctrl+K) combobox ARIA pattern

### LINKED ISSUE #6808: [Bug]: Navigator (Ctrl+K) search field ships an invalid/incomplete combobox ARIA pattern
## Description of the issue

The Navigator / command menu search field carries combobox-style ARIA attributes, but
the pattern is only half wired up. Four separate problems, all in
`src/components/NavigatorSearch/`:

### (a) `role="input"` — an abstract role

https://github.com/saleor/saleor-dashboard/blob/main/src/components/NavigatorSearch/NavigatorSearchInput.tsx#L42-L60

```tsx
// NavigatorSearchInput.tsx:42-60
<Box
  id="navigator-search-input"
  as="input"
  role="input"                          // <- line 45
  ...
  aria-activedescendant="/orders/"      // <- line 58
  aria-expanded                         // <- line 59
/>
```

`input` is an **abstract** role: *"Abstract roles are used for the ontology. Authors
MUST NOT use abstract roles in content."*
([WAI-ARIA 1.2 §5.3.1](https://www.w3.org/TR/wai-aria-1.2/#abstract_roles)). The
native `<input>` already exposes `textbox`; the explicit role only invalidates it.
axe-core reports this as `aria-roles`, impact `critical`.

Note this is a second, hand-written occurrence — `@saleor/macaw-ui`'s `Input`
component has the same default (`src/components/Input/Input.tsx:108`), reported
separately.

### (b) `aria-activedescendant` cannot satisfy either normative branch

WAI-ARIA 1.2 §[`aria-activedescendant`](https://www.w3.org/TR/wai-aria-1.2/#aria-activedescendant):

> Authors **MUST** ensure that one of the following two sets of conditions is met when
> setting the value of `aria-activedescendant` on an element with DOM focus:
>
> - The value of `aria-activedescendant` refers to an **owned element**. An owned element
>   is either a descendant of the element with DOM focus or a logical descendant as
>   indicated by the `aria-owns` attribute.
> - The element with DOM focus is a `combobox`, `textbox` or `searchbox` **with
>   `aria-controls`** referring to an element that supports `aria-activedescendant`, and
>   the value of `aria-activedescendant` refers to an owned element of the controlled
>   element.

Neither holds here:

- The element with DOM focus is a void `<input>`, so it can own nothing, and there is
  no `aria-owns` → first branch fails.
- There is no `aria-controls` anywhere in the component → second branch fails. (Its
  role is also the abstract `input` rather than `combobox`/`textbox`/`searchbox`.)

On top of that, the value hardcoded in JSX is the literal string `"/orders/"`
(line 58). That is a route. It happens to collide with a real element id, because
`ActionLinkItem` sets `id={href}`
([`useActionTriggers.tsx:23`](https://github.com/saleor/saleor-dashboard/blob/main/src/components/NavigatorSearch/useActionTriggers.tsx#L21-L41),
used with `orderListUrl()` at line 147) — so on mount the attribute announces the
"Go to orders" item regardless of what is actually highlighted, and announces a
dangling reference as soon as the query filters that item out.

`useCommandMenuInput.updateAriaActiveDescendant` does keep the attribute in sync
afterwards
([`useCommandMenuInput.ts:10-14`](https://github.com/saleor/saleor-dashboard/blob/main/src/components/NavigatorSearch/useCommandMenuInput.ts#L10-L14)),
so the stale value is a mount-time defect — but the missing `aria-controls` makes the
relationship invalid at *all* times, not just on mount.

### (c) `aria-expanded` on a role that does not support it

`aria-expanded` is supported on `application`, `button`, `checkbox`, `combobox`,
`gridcell`, `link`, `listbox`, `menuitem`, `row`, `rowheader`, `tab`, `treeitem`
([ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/#aria-expanded)). `textbox` and
`searchbox` are not on that list, and the abstract `input` maps to nothing. It is also
written as a bare attribute (always `true`) and never reflects popup state.

### (d) `role="option"` with no `listbox` container, and `aria-selected` on the wrong node

`useActionTriggers.tsx:21-41` renders:

```tsx
<Link href={href} data-href={href} id={href} className="command-menu-item">   // line 23
  <Box className="command-menu-item-content" ... role="option" tabIndex={-1}>  // line 32
```

- `role="option"` requires a `listbox` (or `group` within one) owner. The results
  container in `CommandContent.tsx` has no role at all, and `grep -rn 'role="listbox"'
  src/` returns nothing in the repository — so these options are orphaned.
- Keyboard navigation applies `aria-selected` to the **anchor**, not to the element
  carrying `role="option"`:
  [`useActionItems.ts`](https://github.com/saleor/saleor-dashboard/blob/main/src/components/NavigatorSearch/useActionItems.ts)
  collects `.command-menu-item` (line 20 — that is the `<Link>`) and sets
  `aria-selected` on it (line 32). `aria-selected` is only supported on `gridcell`,
  `option`, `row` and `tab` — never on `link`.
- Consequently `aria-activedescendant` also points at the anchor id rather than at the
  `option` element.

Net effect for a screen-reader user: the field is not announced as a combobox, the
popup is not announced as a list, item count/position are not conveyed, and the
highlighted item is announced inconsistently or not at all.

## Steps to reproduce the problem

1. Sign in to the Dashboard.
2. Press <kbd>Ctrl</kbd>/<kbd>Cmd</kbd> + <kbd>K</kbd> to open the Navigator.
3. Inspect `#navigator-search-input` in DevTools:
   `role="input"`, `aria-expanded="true"`, `aria-activedescendant="/orders/"`, no
   `aria-controls`.
4. Inspect the results list: items carry `role="option"` on a `<div>` nested inside an
   `<a class="command-menu-item">`, with no ancestor `role="listbox"`.
5. Press <kbd>Tab</kbd> / <kbd>↓</kbd> to move the highlight and re-inspect: the
   `aria-selected="true"` attribute lands on the `<a>`, while `role="option"` stays on
   its child `<div>`.
6. Run an axe-core audit (4.12.1 used here) on the open Navigator — `aria-roles`,
   `aria-allowed-attr` and `aria-required-parent` all report violations.
7. With VoiceOver / NVDA, arrow through the list — the active item is not reliably
   announced.

## What did you expect to happen?

The Navigator should implement the normative
[APG combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)
(list-with-popup variant), so assistive technology announces the field as a combobox,
the popup as a list, and each highlighted item as it changes.

### Proposed fix

`combobox` **requires** both `aria-controls` and `aria-expanded`, and supports
`aria-activedescendant` — so switching the role makes the existing attributes legal
instead of removing them.

```diff
   <Box
     id="navigator-search-input"
     as="input"
-    role="input"
+    role="combobox"
+    type="text"
     autoFocus
     onChange={handleSearch}
     value={inputValue}
     autoComplete="off"
+    aria-autocomplete="list"
+    aria-controls="navigator-search-listbox"
+    aria-expanded={isOpen}
+    aria-label={intl.formatMessage(messages.searchPlaceholder)}
     ...
-    aria-activedescendant="/orders/"
-    aria-expanded
+    aria-activedescendant={activeDescendantId}
   />
```

Then, in `CommandContent.tsx`, give the results container the matching role and id:

```diff
-      <Box __height="370px" style={{ scrollbarWidth: "thin" }} width="100%" overflowY="scroll" ...>
+      <Box
+        id="navigator-search-listbox"
+        role="listbox"
+        aria-label={/* e.g. "Search results" */}
+        __height="370px"
+        style={{ scrollbarWidth: "thin" }}
+        width="100%"
+        overflowY="scroll"
+        ...
+      >
```

And align the option/selection nodes in `useActionTriggers.tsx` + `useActionItems.ts`
so that a **single** element is the option:

- move `role="option"` (and the `id` used by `aria-activedescendant`) onto the element
  that `collectItems` selects, i.e. the `.command-menu-item` node itself, instead of
  splitting them across `<a>` and its child `<div>`;
- keep `aria-selected` on that same element (it is valid on `option`, invalid on `link`);
- prefer a stable synthetic id (`navigator-option-<n>`) over the raw route, so the
  reference does not depend on route shape.

Rendering the option as a `<div role="option">` rather than an `<a>` inside a listbox
is usually simpler — APG's list-with-popup pattern expects options, not links, and
navigation is already handled programmatically by `useActionItems.takeAction`.

`aria-activedescendant` initialisation should come from state rather than a literal, so
the mount-time value is either absent or the genuinely active option;
`useCommandMenuInput` already provides `clearActiveDescendant()` for the "nothing
active" case.

### Scope note

`grep -rn 'role="input"' src/` returns exactly one hit
(`NavigatorSearchInput.tsx:45`) — the rest of the Dashboard inherits the same invalid
role from `@saleor/macaw-ui`'s `Input` component, filed separately against
`saleor/macaw-ui`. Fixing only one of the two leaves the other in place.

### References

- [WAI-ARIA 1.2 — Abstract Roles](https://www.w3.org/TR/wai-aria-1.2/#abstract_roles)
- [WAI-ARIA 1.2 — `aria-activedescendant`](https://www.w3.org/TR/wai-aria-1.2/#aria-activedescendant)
- [WAI-ARIA 1.2 — `aria-expanded`](https://www.w3.org/TR/wai-aria-1.2/#aria-expanded)
- [WAI-ARIA 1.2 — `combobox`](https://www.w3.org/TR/wai-aria-1.2/#combobox) (required: `aria-controls`, `aria-expanded`)
- [ARIA APG — Combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)
- [axe-core rule descriptions](https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md)

## Environment

```
Browser and version: Chromium (Playwright), latest stable
OS and version: Linux
Dashboard version: 3.23.16 (self-hosted Docker build); code verified unchanged on main @ 3.23.22
Core version: Saleor 3.23
```

---

ebrahim2355:
## Scope of the change

Fixes #6808 — the Navigator (<kbd>Ctrl</kbd>/<kbd>Cmd</kbd> + <kbd>K</kbd>) shipped combobox-style
ARIA attributes on an abstract role, so none of them took effect. axe-core reported three
**critical** violations on the open palette; screen readers announced neither the field as a
combobox, nor the popup as a list, nor which result was highlighted.

There is no visual change — this is purely the accessibility tree.

### What was wrong

| | |
|---|---|
| `role="input"` | abstract; [MUST NOT be used in content](https://www.w3.org/TR/wai-aria-1.2/#abstract_roles), and it overrode the native input's `textbox` |
| `aria-expanded` | not supported on `textbox`/`searchbox`; written bare, so always `true` |
| `aria-activedescendant="/orders/"` | hardcoded route, and with no `aria-controls` neither normative branch of the attribute's requirements could hold |
| `role="option"` | sat on a **child** of the element that receives `aria-selected` (invalid on `link`), with no `listbox` owner anywhere |

### What this does

The field becomes a `combobox` whose `aria-controls` points at the results container, which
becomes a labelled `listbox`. `aria-expanded` follows the popup. `role="option"`, the `id` and
`aria-selected` now all sit on **one** node — the same node `useActionItems` collects — so
`aria-activedescendant` always resolves to an option owned by the controlled element.

Two option nodes had no `id` at all (the invite-staff item and "See all global search results"),
which made `updateAriaActiveDescendant` write `aria-activedescendant=""` whenever they were
highlighted. They now have ids, and the setter drops the attribute instead of writing an empty
IDREF.

### Decisions worth reviewing

**Action items stay anchors.** The issue suggested rendering options as `<div>` rather than `<a>`.
An explicit `role="option"` already overrides the implicit `link` role, so the pattern is valid
either way — keeping the anchor preserves middle-click / open-in-new-tab and the existing
`data-href` navigation.

**Ids stay route-based** rather than synthetic `navigator-option-<n>`. `restoreFocus(preferredId)`
re-finds the previously highlighted item by id after the list is filtered; index-based ids shift
under filtering, whereas route ids are stable (and are valid HTML5 ids).

**Resource rows are the one place a role swap wasn't enough.** A listbox may only own options, so
the results table and its row groups become presentational and each `<tr>` becomes the option. And
because an option's content is presentational, the cells can no longer be links — a link nested in
an option is dropped by assistive technology while still taking focus (axe: `nested-interactive`).
The row navigates on click instead, which is what <kbd>Enter</kbd> already did via
`takeAction()`.

The cost is that <kbd>Cmd</kbd>-clicking a *resource row* in the palette no longer opens a new tab
(action items, still anchors, are unaffected). I did not reimplement it with `window.open`, because
this codebase already documents that opening internal hrefs in a new tab breaks under a non-root
`APP_MOUNT_URI` (see the comment in `src/ripples/components/AllRipplesModal.tsx`). Happy to add it
back if you'd rather pay that complexity.

`ListSkeleton` is presentational too, so the listbox does not own a `table` while results load.

### Verification

axe-core 4.12.1 over the rendered popup (real components, both loading and loaded states):

- **before** — `aria-roles`, `aria-allowed-attr`, `aria-required-parent`, all *critical*
- **after** — no violations, with every rule enabled (only `color-contrast` off, which jsdom cannot compute)

Also added regression tests that pin the invariants rather than the markup: the field is a combobox
controlling the listbox; `role="option"`, `id` and the collected class live on a single node; rows
are options inside the Navigator and plain table rows everywhere else; and no option contains a
nested link.

`pnpm lint`, `pnpm check-types` (incl. `tsc-strict`), `pnpm knip` and the full Jest suite
(831 suites / 5622 tests) all pass.


lkostrowski (approved):
lgtm, thanks

PR #6859, opened 2026-08-23
Show payment app health on channels and add a channel widgets mount

mirekm:
Merchants can see tripped breakers and critical app problems on the channel page, and payment apps can render their own status or config without leaving the view.


lkostrowski (approved):
remember to update app-sdk

PR #6953, opened 2026-09-21
test(e2e): run against a local Saleor stack

lkostrowski:
## What this does

Adds `e2e/` — a Playwright suite that stands up its own Saleor in Docker, builds the
dashboard from the working tree, and restores the database before every test. The existing
Cloud-snapshot suite moves to `e2e-legacy/` unchanged and is frozen.

`pnpm e2e` is the whole interface. No credentials, no provisioning, nothing to configure.

## Why

The existing suite drives a provisioned Saleor Cloud instance and asserts against 780 lines
of hardcoded snapshot object ids (`e2e-legacy/data/e2eTestData.ts`). Two consequences:

- **A failure can't be reproduced.** Observing one means provisioning a Cloud instance
  first, so in practice a red run gets re-run rather than investigated.
- **A test can't be added** without extending the snapshot and the ids that describe it,
  and tests share one database, so each one has to clean up after itself or be written
  around what the others left behind.

The new suite removes both: the stack is disposable and local, and every test starts from a
known database.

## High-level changes

| Area | Change |
| --- | --- |
| New suite | `e2e/` — compose stack, seeding, actors, fixtures, one proof spec |
| Old suite | `playwright/` → `e2e-legacy/`, frozen; `pnpm e2e:legacy` still runs it |
| CI | `.github/workflows/e2e-local.yml`, dispatch-only, one job per target |
| Scripts | `pnpm e2e`, `e2e:main`, `e2e:all`, `e2e:fresh`, `e2e:down`; old ones renamed to `e2e:legacy*` |
| Docs | `e2e/README.md` (new), `docs/running-tests.md` and root `README.md` rewritten |
| Types/lint | `check-types:e2e` + `check-types:e2e-legacy`; eslint, knip, tsconfig excludes repointed |

No `src/` behaviour changes. The only edit under `src/` is a comment in
`src/auth/tokenStorage.ts` fixing the paths it points at. No UI change, so no screenshots.

### Two Saleor targets

The dashboard speaks to two API generations, and which one a build speaks to is decided by
`FF_USE_STAGING_SCHEMA` — which is **build-time**: it reaches the bundle through vite's
`define` (`vite.config.js`), and that decides which generated GraphQL tree survives
tree-shaking. There is no runtime switch, so a target is a Saleor image *and* a dashboard
build.

| Target | Image | Schema | API | Dashboard |
| --- | --- | --- | --- | --- |
| `stable` | `ghcr.io/saleor/saleor:3.23` | main | 8200 | 9200 |
| `main` | `ghcr.io/saleor/saleor:unstable-main` | staging | 8300 | 9300 |

Both tags float on purpose, so a Saleor patch release is picked up by the next run instead
of by a version-bump PR. Separate ports and compose project names mean a stack left running
from one target can never answer for the other. When 3.24 ships, `e2e/config.ts` is the one
file to edit.

### Seeds are built, not written

A scenario is a named database state, built once per Saleor image through `populatedb` and
the GraphQL API, then captured as `pg_dump --data-only`. Tests never run a build — they
restore the dump, which is a truncate and a `COPY` (~1.5-2.5s).

The base scenario is Saleor's own `populatedb`, chosen over a bootstrap of our own because
Saleor core maintains it, so it follows schema changes for free — and because it mints one
staff user per permission, which is what makes missing-permission flows testable.

**Dumps are cached under the image's content id, not its tag.** Both tags float, so a new
`3.23` patch has to invalidate them. The cached logins share that key, so a rebuilt seed can
never be paired with a token minted against the users it replaced.

### Isolation, and opting into parallel

The database is restored **before every test**, and those tests run one at a time. That is
the point of the suite: a failing test reproduces on its own rather than as a function of
the tests that ran before it. Specs need no cleanup discipline and no unique names.

Serial is the default because a restore empties every table for a second or two — a test
running alongside one would see a shop with no products in it. A block that writes nothing
opts out with `{ tag: parallel() }` and then runs concurrently. The tag takes a *scenario
name*, so a group and the database state it runs on are the same thing and cannot drift
apart. Playwright project `dependencies` sequence the phases; they are the only ordering
primitive that spans workers, which is also why a group's restore is a project rather than a
hook.

Signing in is the one thing that cannot be tagged: Saleor throttles `tokenCreate` per IP
unconditionally (`saleor/account/throttling.py`), so parallel UI logins fail with
`LOGIN_ATTEMPT_DELAYED`. Specs that arrive already signed in are unaffected — a stored
session refreshes its token, and nothing throttles that.

### Sessions

A session is one `localStorage` entry: the dashboard reads `_saleorRefreshToken` on start,
so there is no UI login to drive and no cookie to carry. Global setup signs every actor in
over the API once and caches the state, which also means a run cannot start if a seeded user
can't authenticate.

### Speed

A cold run pulls the image, migrates, seeds and builds the dashboard: ~2.5 min to the first
test. A warm run reaches it in ~13s — the image pull, the migration, the seed build and the
`vite build` are each skipped when their inputs are unchanged. Every phase announces itself
with its elapsed time (`[e2e 1m59s] > running populatedb`), so a slow run is visibly a slow
run rather than a hang.

### CI

`e2e-local.yml` is dispatch-only, with a job per target. A run is dominated by a `vite build`
per target, so putting it on every pull request before the suite is broad enough would mostly
buy queue time — that's a one-line change to the `on:` block when it's worth it.

## Trying it

```bash
pnpm e2e            # stable target
pnpm e2e:main       # Saleor's unreleased main
pnpm e2e --ui       # ...and anything else Playwright understands
pnpm e2e:fresh      # drop the database volume and caches first
pnpm e2e:down       # stop the stacks
```

Requires Docker. First run takes a few minutes.

## Deliberately not in this PR

`e2e/README.md` carries the full list with what would undo each one. The ones worth knowing
here:

- **One proof spec** (`tests/login.spec.ts`). It exercises the whole stack end to end, but
  porting real coverage from `e2e-legacy/` is follow-up work. The next port worth making is
  a single-permission spec — the part of the actor model most likely to be subtly wrong, and
  the first place a `populatedb` difference between `3.23` and `unstable-main` would show.
- **No parallel group has members yet**, since the one spec can't be tagged.
- **The serial phase is serial.** One database per worker would parallelise the resetting
  tests too; it also needs one Saleor API per database, because the bundle bakes `API_URL`
  at build time.
- **`nginx/replace-env-vars.sh` is not exercised** — the bundle is mounted read-only with
  its environment baked in, so the shipped image's runtime substitution has no coverage
  here. That wants a smoke test of the actual Docker image.
- **No changeset**, per `CLAUDE.md`: test infrastructure and CI are excluded.

## Review notes

`e2e-legacy/` is a pure `git mv` plus three path fixes (`auth.js`'s `.auth` directory, one
import in `basePage.ts`, and `testMatch`/`testIgnore` in the moved config). Reviewing it as
a rename is enough.


mirekm (dismissed):

.github/workflows/e2e-local.yml
```diff
@@ -0,0 +1,90 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions: {}
+
+jobs:
+  e2e:
+    runs-on: ubuntu-22.04
```
NyanKiyoshi: Outdated and deprecated

.github/workflows/e2e-local.yml
```diff
@@ -0,0 +1,90 @@
...
+      # at all - means run them both.
+      matrix:
+        target: ${{ fromJSON((inputs.target == 'stable' || inputs.target == 'main') &&
+          format('["{0}"]', inputs.target) || '["stable","main"]') }}
+
+    permissions:
+      contents: read
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
NyanKiyoshi: Outdated

.github/workflows/e2e-local.yml
```diff
@@ -0,0 +1,90 @@
...
+        with:
+          name: e2e-${{ matrix.target }}-report--attempt-${{ github.run_attempt }}
+          path: |
+            e2e/test-results
+            playwright-report
+          retention-days: 7
+
+      - name: Tear the stack down
+        if: always()
+        run: node --disable-warning=MODULE_TYPELESS_PACKAGE_JSON e2e/run.ts --down --target=${{ matrix.target }}
```
NyanKiyoshi: Use env vars instead of template interpolation (`${{ matrix.target }}`)

.github/workflows/e2e-local.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: Run E2E
+        env:
+          E2E_TARGET: ${{ matrix.target }}
+          GREP: ${{ inputs.grep }}
+        run: |
+          # `pnpm run generate` runs both schemas; the staging tree is what the `main`
+          # target's bundle is built from, so it cannot be skipped for either target.
+          pnpm run generate
+          node --disable-warning=MODULE_TYPELESS_PACKAGE_JSON e2e/run.ts \
+            --target="$E2E_TARGET" ${GREP:+--grep "$GREP"}
```
NyanKiyoshi: Missing quotes:
```suggestion
          args=()
          if [[ -n "$GREP" ]]; then
              args+=( --grep )
          fi
          node --disable-warning=MODULE_TYPELESS_PACKAGE_JSON e2e/run.ts \
            --target="$E2E_TARGET" "${args[@]}"
```

.github/workflows/e2e-local.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Upload report
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
```
NyanKiyoshi: Outdated

e2e/compose.yml
```diff
@@ -0,0 +1,97 @@
...
+# variable from `lib/docker.ts`, and the two get separate compose project names, so a
+# stable stack and a main stack never share a container, a volume or a port.
+
+services:
+  api:
+    # Both targets use floating tags; the suite re-resolves them itself before starting
+    # anything, because the image's content id is what keys the seed cache. See `pullImage`.
+    image: ${SALEOR_IMAGE}
+    ports:
+      - "${SALEOR_API_PORT}:8000"
```
NyanKiyoshi:
```suggestion
      - "127.0.0.1:${SALEOR_API_PORT}:8000"
```

NyanKiyoshi (changes requested):

.github/workflows/lint-workflows.yml
```diff
@@ -0,0 +1,68 @@
...
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          # Deleted files cannot be linted, so only added/copied/modified/renamed ones.
+          mapfile -t changed < <(
+            git diff --name-only --diff-filter=ACMR "$BASE_SHA...$HEAD_SHA" -- '.github/workflows/**'
+          )
+
+          if [[ ${#changed[@]} -eq 0 ]]; then
```
NyanKiyoshi:
```suggestion
          if [[ "${#changed[@]}" -eq 0 ]]; then

File: .github/workflows/lint-workflows.yml
Comment (NyanKiyoshi) [inline .github/workflows/lint-workflows.yml:1]: If you want to lint workflows, you can also add https://github.com/zizmorcore/zizmor-action, can be a separate workflow, e.g., `.github/workflows/zizmor.yaml`

File: package.json [hunk 8]
```diff
@@ -9,6 +9,7 @@
   "author": "Saleor Commerce",
   "main": "src/index.tsx",
   "scripts": {
+    "actionlint": "docker run --rm --volume \"$PWD\":/repo --workdir /repo rhysd/actionlint:1.7.12@sha256:b1934ee5f1c509618f2508e6eb47ee0d3520686341fec936f3b79331f9315667",
```
NyanKiyoshi: Will not work with some systems without `docker.io` (due to `rhysd` being an unknown package in Podman unlike things like `postgres` and `python`):

```suggestion
    "actionlint": "docker run --rm --volume \"$PWD\":/repo --workdir /repo docker.io/rhysd/actionlint:1.7.12@sha256:b1934ee5f1c509618f2508e6eb47ee0d3520686341fec936f3b79331f9315667",
```

NyanKiyoshi (dismissed):

NyanKiyoshi (approved):

PR #6954, opened 2026-09-23
Datetime display skill

mirekm:
I propose to use it for handling display of dates in Dashboard and App WIdgets between relative, absolute UTC, and localised time.

- The label names the event and follows `MerchantDate`: `Placed 4 minutes ago`, `Placed today at 2:29 PM`, then a calendar date. No seconds. A bare relative time only when a column or group header already says what it is.
- Hover shows two clocks, UTC and the local offset (`GMT+2`, `GMT+5:30`), each with that clock’s own date and time, including seconds.
- Date-only values stay on `Date`. Named events extend `MerchantDate`; a time that already has a header extends `DateTime`.


.claude/skills/saleor-datetime-display/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+
+## When this tooltip earns its space
+
+Use it for a timestamp a merchant may need to match exactly: order placed, created,
+timeline events.
+
+Skip it for date-only fields (`Date`). Skip it when a column or group header already
+names the calendar day and the cell has no time.
+
+## What the page shows
```
lkostrowski: Im not sure how this is a skill? If we have a component that take date string and displays it according the rules, the skill only should say to use that component, agent dont have to figure this out (which is slow and expensive)

mirekm: It's a skill as not every date looks now or will look the same in Dashboard and apps therefor I expect different components. Alternative would be to adjust every component showing date and this is not mutually exclusive (we should do it too).

